PR #5, opened 2019-08-22
Added mock infra and route externalization.

michaelkleinhenz:
This adds mocking commandline parameters and a best-practice approach on how to set test mode. It also externalizes the route service implementation and creates a first default "health" route. Also adds tests for services.


health/HealthService.go
```diff
@@ -0,0 +1,33 @@
+package health
```
alexeykazakov: Please use snake_case for filenames. It's a common practice if not the official convention.

health/HealthService.go
```diff
@@ -0,0 +1,33 @@
...
+
+// NewHealthService returns a new HealthService instance.
+func NewHealthService(isTestMode bool) *Service {
+	r := new(Service)
+	r.isTestMode = isTestMode
+	return r
+}
+
+// getHealthInfo returns the health info.
+func (s *Service) getHealthInfo() map[string]bool {
```
alexeykazakov: HTTP health check in k8s is based on the http response code. Not on the response body.
Any response between 200 and 400 will be considered as success. See https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#define-a-liveness-http-request

We can encode some meaningful content as an extra information (for debugging, tests, or whatever reason) but we still have to return a proper response code.

PR #6, opened 2019-08-22
Scaffolding for Registration Service

tinakurian:
This PR adds basic support for configuration, testing and kubernetes api. 

Relates to: https://jira.coreos.com/browse/CRT-219


pkg/registrationserver/routes.go
```diff
@@ -0,0 +1,66 @@
+//go:generate go run -tags=dev pkg/static/assets_generate.go
```
michaelkleinhenz: This needs to be
```
//go:generate go run -tags=dev ../../pkg/static/assets_generate.go
```

michaelkleinhenz: ..it is always contextualized from the directory the directive is found. If you move `main.go` into `pkg`, this should also be moved back to `main.go` (and the path again updated).

pkg/static/assets_generate.go
```diff
@@ -14,7 +14,7 @@ func main() {
 		PackageName:  "static",
 		BuildTags:    "!dev",
 		VariableName: "Assets",
-		Filename:     "static/generated_assets.go",
+		Filename:     "pkg/static/generated_assets.go",
```
michaelkleinhenz: This needs to be 
```
Filename:     "../static/generated_assets.go",
```
as this contextualized from the source position. There may be also some changes to this when moving `main.go` into `pkg`.

Makefile
```diff
@@ -52,16 +52,16 @@ build-prod:
```
michaelkleinhenz: This did not work for me unless I change the `generate` line to 
```
go generate pkg/registrationserver/routes.go
```
to explicitly run the file with the `generate` comment. All of my other comments on the paths may be related to this. Imo this is weird, I would expect `go generate` to pick up all `generate` comments in the source, but for me, it looks like it is only looking at `main.go` (you can add `-v` to the `go generate` command to see which files it is looking at). In general, the asset generation is not working correctly for me with your changes. I can make it work by updating the paths (see other comments), but I guess there may be a better way. And with `main.go` moving into `pkg`, this needs update as well I guess.

pkg/registrationserver/health_check_handler.go
```diff
@@ -0,0 +1,13 @@
+package registrationserver
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+// HealthCheckHandler returns a default heath check result.
```
michaelkleinhenz: This needs more detailing and a mock mode. But I can adapt my PR for this after this one is merged.

pkg/registrationserver/registrationserver.go
```diff
@@ -0,0 +1,126 @@
...
+	Commit = "0"
+	// BuildTime set by build script in ISO 8601 (UTC) format:
+	// YYYY-MM-DDThh:mm:ssTZD (see https://www.w3.org/TR/NOTE-datetime for
+	// details)
+	BuildTime = "0"
+	// StartTime in ISO 8601 (UTC) format
+	StartTime = time.Now().UTC().Format("2006-01-02T15:04:05Z")
+)
+
+// AppServer bundles configuration, logging, and HTTP server objects in a single
```
michaelkleinhenz: This should be `RegistrationServer`.

pkg/registrationserver/routes.go
```diff
@@ -0,0 +1,66 @@
...
+// serve the SPA in the given static directory.
+type spaHandler struct {
+	Assets http.FileSystem
+}
+
+// ServeHTTP inspects the URL path to locate a file within the static dir
+// on the SPA handler. If a file is found, it will be served. If not, the
+// file located at the index path on the SPA handler will be served. This
+// is suitable behavior for serving an SPA (single page application).
+func (h spaHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
michaelkleinhenz: I guess this one is never used...

main.go
```diff
@@ -1,77 +1,83 @@
...
-// HealthCheckHandler returns a default heath check result.
-func HealthCheckHandler(w http.ResponseWriter, r *http.Request) {
-	// default handler for system health
-	w.Header().Set("Content-Type", "application/json")
-	json.NewEncoder(w).Encode(map[string]bool{"alive": true})
-}
+	// listen concurrently to allow for graceful shutdown
+	go func() {
+		srv.Logger().Printf("Listening on %q...", srv.Config().GetHTTPAddress())
+		if err := srv.HTTPServer().ListenAndServe(); err != nil {
```
michaelkleinhenz: AFAIK, this has removed the SSL support...see `master`. I guess there should be a config directive each for certificate and key files and this line here should check whether to run in http or https mode. I also added an `insecure` config in `master`. The default would be https mode, with bail out when the cert and key files could not be found. If someone wants to run in http mode, that must be explicitly given. Makes it a bit more secure.

pkg/registrationserver/registrationserver.go
```diff
@@ -0,0 +1,126 @@
...
+import (
+	"log"
+	"net/http"
+	"os"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/codeready-toolchain/registration-service/pkg/configuration"
+	"github.com/gorilla/handlers"
```
michaelkleinhenz: This needs to be "handlers", not "Handlers". It produces weird issues, looks like they have some case sensitivity in there.

pkg/kubeclient/client_test.go
```diff
@@ -0,0 +1,13 @@
+package kubeclient
+
+import (
+	"github.com/stretchr/testify/assert"
+	"k8s.io/client-go/kubernetes/fake"
+	"testing"
+)
+
+func TestNewKubeClient(t *testing.T) {
```
michaelkleinhenz: I get this when running the tests:
```
# k8s.io/client-go/rest
../../../../pkg/mod/k8s.io/client-go@v11.0.0+incompatible/rest/request.go:598:31: not enough arguments in call to watch.NewStreamWatcher
        have (*versioned.Decoder)
        want (watch.Decoder, watch.Reporter)
FAIL    github.com/codeready-toolchain/registration-service/pkg/kubeclient [build failed]
```

michaelkleinhenz: ...but only after I did a `go get -u` and a `go mod tidy`. This refreshes the dependencies. Before that, the test was working for me.

pkg/kubeclient/client_test.go
```diff
@@ -0,0 +1,13 @@
+package kubeclient
+
+import (
+	"github.com/stretchr/testify/assert"
```
michaelkleinhenz: I needed to fetch this manually using `GO111MODULE=off go get -v k8s.io/client-go/kubernetes/fake`. I think this needs to be added to the dependency list in the `Makefile`.

pkg/configuration/configuration.go
```diff
@@ -0,0 +1,141 @@
+// Package configuration is in charge of the validation and extraction of all
+// the configuration details from a configuration file or environment variables.
+package configuration
+
+import (
+	"strings"
+	"time"
+
+	errs "github.com/pkg/errors"
+	"github.com/spf13/viper"
```
michaelkleinhenz: I also needed to fetch this manually using `GO111MODULE=off go get github.com/spf13/viper`. It *is* in go.mod though, so I don't know why this is not working. It worked from the beginning in the terminal as well. This is really a mess when the terminal behaves differently from the vscode language server.

pkg/health/health_check_handler.go
```diff
@@ -0,0 +1,46 @@
...
+func (srv *Service) HealthCheckHandler(w http.ResponseWriter, r *http.Request) {
+	// default handler for system health
+	w.Header().Set("Content-Type", "application/json")
+	healthInfo := srv.getHealthInfo()
+	if healthInfo["alive"].(bool) {
+		w.WriteHeader(http.StatusOK)
+	} else {
+		w.WriteHeader(http.StatusInternalServerError)
+	}
+	json.NewEncoder(w).Encode(healthInfo)
```
golangcibot: Error return value of `(*encoding/json.Encoder).Encode` is not checked (from `errcheck`)

michaelkleinhenz:
LGTM

pkg/configuration/configuration.go
```diff
@@ -0,0 +1,208 @@
...
+
+	varHTTPWriteTimeout = "http.write_timeout"
+	// DefaultHTTPWriteTimeout specifies the default timeout for HTTP writes
+	DefaultHTTPWriteTimeout = time.Second * 15
+
+	varHTTPReadTimeout = "http.read_timeout"
+	// DefaultHTTPReadTimeout specifies the default timeout for HTTP reads
+	DefaultHTTPReadTimeout = time.Second * 15
+
+	varHTTPInsecure = "http.insecure"
```
xcoulon: all these HTTPS/cert related variables are not needed (will run in HTTP inside the container, secured HTTPS connection will be handled at the cluster's edge router)

pkg/registrationserver/registrationserver.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/gorilla/mux"
+	errs "github.com/pkg/errors"
+)
+
+var (
+	// Commit current build commit set by build script
```
xcoulon: this could be moved in the `configuration` pkg

pkg/registrationserver/registrationserver.go
```diff
@@ -0,0 +1,126 @@
...
+
+	logger      *log.Logger
+	routesSetup sync.Once
+}
+
+// New creates a new RegistrationServer object with reasonable defaults.
+func New(configFilePath string) (*RegistrationServer, error) {
+	srv := &RegistrationServer{
+		router: mux.NewRouter(),
+		logger: log.New(os.Stderr, "", 0),
```
xcoulon: do we need to log to `Stderr`?

pkg/registrationserver/registrationserver.go
```diff
@@ -0,0 +1,126 @@
+package registrationserver
```
xcoulon: I would suggest to use `server` as the pkg name, and `server.go` as the "main" file in this pkg. The repo is `registration-service`, so I believe there's no need to duplicate the `registration` term

pkg/testutils/unset_env_var_and_restore.go
```diff
@@ -0,0 +1,28 @@
+package testutils
```
xcoulon: this `testutils` package may not need to be a child of `pkg` (AFAIU, `pkg` is the "root" for the public packages)

pkg/configuration/configuration.go
michaelkleinhenz: Yes, but the SSL support is only a few lines (without the configuration options, it's actually <5 lines), and as this is more or less a scaffolding, I would actually leave it in. Maybe we need it when we're possibly re-using this scaffolding for other services.

pkg/registrationserver/registrationserver.go
michaelkleinhenz: Agree, those should be moved to the configuration package (though not the configuration options).

pkg/testutils/unset_env_var_and_restore.go
michaelkleinhenz: Agree, I guess this was just missed when the packages got moved around.

pkg/registrationserver/registrationserver.go
michaelkleinhenz: Good question. What is the best practice for a k8s application? Logging to stdout?

pkg/configuration/configuration.go
michaelkleinhenz: I would do that in the non-scaffolding part to keep the base open for any use. But to avoid more discussion here, I'll remove it.

pkg/kubeclient/client_test.go
```diff
@@ -0,0 +1,13 @@
...
+	"k8s.io/client-go/kubernetes/fake"
+	"testing"
+)
+
+func TestNewKubeClient(t *testing.T) {
+	k := KubeClient{}
+	k.CoreClient = fake.NewSimpleClientset()
+	assert.NotNil(t, k.CoreClient, "Kubecore client shouldn't be nil")
```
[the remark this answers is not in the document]
michaelkleinhenz: Nothing, but this is only scaffolding. There is no real boundary, but I would expect to have the actual k8s code and tests in CRT-72 instead of the scaffolding code. The goal here would be to just add the correct libs to the code and do as much as to confirm they are there and can be used.

pkg/health/health_check_handler.go
```diff
@@ -0,0 +1,50 @@
...
+	return r
+}
+
+// getHealthInfo returns the health info.
+func (srv *Service) getHealthInfo() map[string]interface{} {
+	m := make(map[string]interface{})
+	// TODO: this need to get actual health info.
+	m["alive"] = !srv.config.IsTestingMode()
+	m["testingmode"] = srv.config.IsTestingMode()
+	m["version"] = srv.config.GetVersion()
```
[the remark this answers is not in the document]
michaelkleinhenz: Ok, let's do this in a seperate PR outside of the scaffolding.

pkg/kubeclient/client.go
```diff
@@ -0,0 +1,39 @@
...
+
+	return getOpenshiftAPIConfig(host, bearerToken)
+}
+
+func getOpenshiftAPIConfig(host string, bearerToken string) rest.Config {
+	return rest.Config{
+		Host:        host,
+		BearerToken: bearerToken,
+		TLSClientConfig: rest.TLSClientConfig{
+			Insecure: true,
```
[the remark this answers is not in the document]
michaelkleinhenz: Yes. But remember, this is only bare scaffolding. It is only there to confirm we added the correct libs to the project and they are usable. I wouldn't expect any useful code here. Same for tests (see below).

main.go
```diff
@@ -1,120 +1,101 @@
...
-}
-
-// HealthCheckHandler returns a default heath check result.
-func HealthCheckHandler(w http.ResponseWriter, r *http.Request) {
-	// default handler for system health
-	w.Header().Set("Content-Type", "application/json")
-	json.NewEncoder(w).Encode(map[string]bool{"alive": true})
-}
+// Version is the service version
+const Version string = "0.0.1"
```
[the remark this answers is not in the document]
michaelkleinhenz: Fixed in 3f8a31b.

pkg/configuration/configuration.go
```diff
@@ -0,0 +1,219 @@
...
+	"strings"
+	"time"
+
+	errs "github.com/pkg/errors"
+	"github.com/spf13/viper"
+)
+
+var (
+	// Commit current build commit set by build script
+	Commit = "0"
```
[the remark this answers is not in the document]
michaelkleinhenz: Fixed in 3f8a31b. Values are now set at build time.

pkg/configuration/configuration.go
```diff
@@ -0,0 +1,219 @@
...
+	// BuildTime set by build script in ISO 8601 (UTC) format:
+	// YYYY-MM-DDThh:mm:ssTZD (see https://www.w3.org/TR/NOTE-datetime for
+	// details)
+	BuildTime = "0"
```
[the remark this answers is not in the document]
michaelkleinhenz: Fixed in 3f8a31b. Values are now set at build time.

main.go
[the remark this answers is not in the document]
michaelkleinhenz: ..and in 8e0bf9d.

pkg/configuration/configuration.go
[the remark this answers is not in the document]
michaelkleinhenz: ..and in 8e0bf9d.

pkg/configuration/configuration.go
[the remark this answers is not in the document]
michaelkleinhenz: ..and in 8e0bf9d.

pkg/health/health_check_handler_test.go
```diff
@@ -0,0 +1,101 @@
...
+		// We create a ResponseRecorder (which satisfies http.ResponseWriter) to record the response.
+		rr := httptest.NewRecorder()
+		// Our handlers satisfy http.Handler, so we can call their ServeHTTP method
+		// directly and pass in our Request and ResponseRecorder.
+		handler.ServeHTTP(rr, req)
+		// Check the status code is what we expect.
+		assert.Equal(t, rr.Code, http.StatusInternalServerError, "handler returned wrong status code: got %v want %v", rr.Code, http.StatusInternalServerError)
+		// Check the response body is what we expect.
+		var data map[string]interface{}
+		if err := json.Unmarshal(rr.Body.Bytes(), &data); err != nil {
```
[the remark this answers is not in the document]
michaelkleinhenz: Fixed.

pkg/health/health_check_handler_test.go
```diff
@@ -0,0 +1,101 @@
...
+			t.Fatal(err)
+		}
+		val, ok := data["version"]
+		assert.True(t, ok, "no version key in health response")
+		valString, ok := val.(string)
+		assert.True(t, ok, "returned 'version' value is not of type 'string'")
+		assert.Equal(t, valString, configRegistry.GetVersion())
+	})
+
+	t.Run("health in testing mode", func(t *testing.T) {
```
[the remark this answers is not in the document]
michaelkleinhenz: Fixed, that test was removed as the version field was removed from config anyway. See 8e0bf9d.

pkg/server/server.go
```diff
@@ -0,0 +1,114 @@
...
+// Router returns the app server's HTTP router
+func (srv *RegistrationServer) Router() *mux.Router {
+	return srv.router
+}
+
+// GetRegisteredRoutes returns all registered routes formatted with their
+// methods, paths, queries and names. It is a good idea to print this
+// information on server start to give you an idea of what routes are
+// available in the system.
+func (srv *RegistrationServer) GetRegisteredRoutes() (string, error) {
```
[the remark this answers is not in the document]
michaelkleinhenz: Fixed in 5c2d9b2.

pkg/unset_env_var_and_restore.go
```diff
@@ -0,0 +1,28 @@
...
+// In a test you can use this to temporarily set an environment variable:
+//
+//    func TestFoo(t *testing.T) {
+//        restoreFunc := testutils.UnsetEnvVarAndRestore("foo")
+//        defer restoreFunc()
+//        os.Setenv(key, "bar")
+//
+//        // continue as if foo=bar
+//    }
+func UnsetEnvVarAndRestore(key string) func() {
```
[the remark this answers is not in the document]
michaelkleinhenz: Fixed in 199eca3.

pkg/health/health_check_handler_test.go
```diff
@@ -45,41 +43,41 @@ func TestHealthCheckHandler(t *testing.T) {
 		// Check the response body is what we expect.
 		var data map[string]interface{}
 		if err := json.Unmarshal(rr.Body.Bytes(), &data); err != nil {
-			t.Fatal(err)
+			require.NoError(t, err)
```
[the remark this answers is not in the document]
michaelkleinhenz: I should really start reading the lines before and after the line in question :-)

pkg/server/server_test.go
```diff
@@ -0,0 +1,25 @@
...
+	// specific config params do not matter for testing the routes setup.
+	srv, err := registrationserver.New("../../example-config.yml")
+	require.NoError(t, err)
+
+	// setting up the routes.
+	err = srv.SetupRoutes()
+	require.NoError(t, err)
+
+	// check if there is an error when retriving the routes,
+	// note that the actual routes are tested elsewhere.
```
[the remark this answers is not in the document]
michaelkleinhenz: No, the routes test is a bit more complicated and is in `routes_test.go`. `server.go` sets up the server framework, `routes.go` sets up the routes, so I think it makes sense to test the two areas in seperate test files.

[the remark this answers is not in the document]
michaelkleinhenz: or we can just remove `server_test.go` as there is nothing that is not also tested in `routes_test.go` as well. But this is a scaffolding. There might be stuff added to `server.go` that needs testing later.

MatousJobanek:
May I have one proposal? Could we please split this PR to more PRs:
1. README, Dockerfile, gitignore
2. Makefile + configuration
3. Kubeclient (it's not used anywhere for now)
4. The rest

it would improve the code review. Is it possible @tinakurian @michaelkleinhenz ?

michaelkleinhenz:
@sbryzak I'll add the documentation here: https://github.com/codeready-toolchain/registration-service/pull/8. Will be there in ~1h.

sbryzak:
@tinakurian @michaelkleinhenz Alexey pointed me to [this issue](https://jira.coreos.com/browse/CRT-217) for which I understand there was a team discussion during one of the meetings that I was unable to attend due to timezone.  I've done some further evaluation today and would like to make an alternate suggestion to Gorilla.  It seems that Gin (Github repo [here](https://github.com/gin-gonic/gin)) is much more popular and under a lot more active development.  My research also seems to indicate that it is far more performant than Gorilla also, and the API looks simple and powerful.  I'd like us to consider switching to Gin as it seems like a safer choice of library.

michaelkleinhenz:
@sbryzak yes, we intentionally chose Gorilla over Goa (those were the two we looked at) as for Gorilla, there was considerable amount of experience in our group. It was already used in openshift.io and we took some existing code from Konrad's repositories.

For Gin: let's consider this, but get this one finished first. We also intentionally choose a "lightweight" framework instead of Goa as this enables us to easily switch between similar frameworks. A move from Gorilla to something like Gin should be doable in a reasonable amout of time (though not until next Tuesday). With Goa, we would have a much deeper vendor lock.

PR #9, opened 2019-08-30
Fix coverage reports

michaelkleinhenz:
This fixes the coverage reports..

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -7,6 +7,5 @@ test: generate
 	@-rm -f $(COV_DIR)/coverage.txt
 	go test -count=1 -coverprofile=$(COV_DIR)/profile.out -covermode=atomic ./...
 ifeq (,$(wildcard $(COV_DIR)/profile.out))
-	cat $(COV_DIR)/profile.out >> $(COV_DIR)/coverage.txt
-	rm $(COV_DIR)/profile.out
+	mv $(COV_DIR)/profile.out $(COV_DIR)/coverage.txt
 endif 
--- a/make/test.mk
+++ b/make/test.mk
@@ -1,4 +1,12 @@
+COV_DIR = coverage
+
 .PHONY: test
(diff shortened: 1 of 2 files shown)


make/test.mk
```diff
@@ -1,4 +1,18 @@
...
 ## runs all tests with bundles assets
 test: generate
-	go test -count=1 ./...
+	# test the test package seperately, no coverage report
+	go test -count=1 ./test
+	# test the rest of the code, with coverage report	
+	@-mkdir -p $(COV_DIR)
+	@-rm -f $(COV_DIR)/coverage.txt
+	@-for d in $(GOPKGS) ; do \
+	  go test -coverprofile=profile.out -covermode=atomic -coverpkg=$$GOPKGS $$d ; \
```
alexeykazakov: You don't have to test each package individually when using coverpkg. Just run go test once with that flag.

make/test.mk
```diff
@@ -1,4 +1,18 @@
+COV_DIR = coverage
+GOPKGS = $(shell go list ./... | grep -v /test)
+
 .PHONY: test
 ## runs all tests with bundles assets
 test: generate
-	go test -count=1 ./...
+	# test the test package seperately, no coverage report
+	go test -count=1 ./test
```
alexeykazakov: Also, it's actually fine to include "test" package in the coverage report if it's utility for tests which also covered by tests. Which is the case.
So, let's just run "go test" once for everything. I just thought that "test" package is a utility for tests which doesn't have any tests (to test test utilities) at all.

michaelkleinhenz: At least now I know how Make expands bash commands :-)

michaelkleinhenz: Fixed 4b40afb.

make/test.mk
michaelkleinhenz: Fixed 4b40afb.

PR #10, opened 2019-09-04
Replace gorilla API with gin, introduce API versioning

sbryzak:
This PR replaces gorilla with gin, and introduces simple versioning for API endpoints.

A couple of notes about this PR:

- Versioning has been added, so now we have a "v1" set of endpoint handlers, e.g. `/api/v1/health`
- The `TestRoutes()` test in `routes_test.go` has been removed, as it doesn't really seem useful.  If all service endpoints are tested individually and all tests pass, then by logical extension the routes must be correct.
- gin supports multiple modes (for logging purposes I believe) - by default it runs in "debug" mode but we can set this either via an environment variable or we could introduce a configuration param.


cmd/main.go
```diff
@@ -44,7 +44,7 @@ func main() {
 		panic(err.Error())
 	}
 
-	routesToPrint, err := srv.GetRegisteredRoutes()
+	routesToPrint := srv.GetRegisteredRoutes()
 	if err != nil {
```
alexeykazakov: You can now remove this `if` block.

sbryzak: Done.

PR #13, opened 2019-09-05
Scaffolding for signup service

sbryzak:
Subtask of https://jira.coreos.com/browse/CRT-67

This PR adds empty scaffolding for both the `/signup` and `/signup_callback` endpoints.  

Signed-off-by: Shane Bryzak <sbryzak@gmail.com>

### DIFF
--- a/pkg/signup/signup_callback_handler.go
+++ b/pkg/signup/signup_callback_handler.go
@@ -8,22 +8,22 @@ import (
 	"net/http"
 )
 
-// SignupService implements the signup endpoint, which is invoked for new user registrations.
-type SignupService struct {
+// SignupCallbackService implements the signup callback endpoint.
+type SignupCallbackService struct {
 	config *configuration.Registry
 	logger *log.Logger
 }
 
-// NewSignupService returns a new SignupService instance.
-func NewSignupService(logger *log.Logger, config *configuration.Registry) *SignupService {
-	return &SignupService{
+// NewSignupCallbackService returns a new SignupCallbackService instance.
+func NewSignupCallbackService(logger *log.Logger, config *configuration.Registry) *SignupCallbackService {
+	return &SignupCallbackService{
 		logger: logger,
 		config: config,
 	}
 }
 
 // HandleRequest returns a default heath check result.
-func (srv *SignupService) HandleRequest(ctx *gin.Context) {
+func (srv *SignupCallbackService) HandleRequest(ctx *gin.Context) {
 	ctx.Writer.Header().Set("Content-Type", "application/json")
 
 	err := json.NewEncoder(ctx.Writer).Encode(nil)
--- a/pkg/signup/signup_callback_handler_test.go
+++ b/pkg/signup/signup_callback_handler_test.go
@@ -1,0 +1,1 @@
+package signup
--- a/pkg/signup/signup_callback_handler_test.go
+++ b/pkg/signup/signup_callback_handler_test.go
@@ -1,1 +1,45 @@
-package signup
+package signup_test
+
+import (
+	"github.com/codeready-toolchain/registration-service/pkg/configuration"
+	"github.com/codeready-toolchain/registration-service/pkg/signup"
+	"github.com/gin-gonic/gin"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"log"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"testing"
+)
+
+func TestSignupCallbackHandler(t *testing.T) {
+	// Create a request to pass to our handler. We don't have any query parameters for now, so we'll
+	// pass 'nil' as the third parameter.
+	req, err := http.NewRequest("GET", "/api/v1/signup_callback", nil)
+	require.NoError(t, err)
+
+	// create logger and registry.
+	logger := log.New(os.Stderr, "", 0)
+	configRegistry := configuration.CreateEmptyRegistry()
+
+	// set the config for testing mode, the handler may use this.
+	configRegistry.GetViperInstance().Set("testingmode", true)
+	assert.True(t, configRegistry.IsTestingMode(), "testing mode not set correctly to true")
+
+	// create handler instance.
+	signupCallbackService := signup.NewSignupCallbackService(logger, configRegistry)
+	handler := gin.HandlerFunc(signupCallbackService.HandleRequest)
+
+	t.Run("signup", func(t *testing.T) {
+		// We create a ResponseRecorder (which satisfies http.ResponseWriter) to record the response.
+		rr := httptest.NewRecorder()
+		ctx, _ := gin.CreateTestContext(rr)
+		ctx.Request = req
+
+		handler(ctx)
+
+		// Check the status code is what we expect.
+		require.Equal(t, http.StatusOK, rr.Code)
+	})
+}
--- a/pkg/signup/signup_handler.go
+++ b/pkg/signup/signup_handler.go
@@ -1,0 +1,34 @@
+package signup
+
+import (
+	"encoding/json"
+	"github.com/codeready-toolchain/registration-service/pkg/configuration"
+	"github.com/gin-gonic/gin"
+	"log"
+	"net/http"
+)
+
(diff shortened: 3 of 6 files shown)


pkg/signup/signup_callback_handler.go
```diff
@@ -0,0 +1,34 @@
...
+
+// NewSignupCallbackService returns a new SignupCallbackService instance.
+func NewSignupCallbackService(logger *log.Logger, config *configuration.Registry) *SignupCallbackService {
+	return &SignupCallbackService{
+		logger: logger,
+		config: config,
+	}
+}
+
+// HandleRequest returns a default heath check result.
```
alexeykazakov: Health check result? Copy-paste from health check endpoint/handler I guess.

pkg/signup/signup_callback_handler.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+// SignupCallbackService implements the signup callback endpoint.
+type SignupCallbackService struct {
+	config *configuration.Registry
+	logger *log.Logger
+}
+
+// NewSignupCallbackService returns a new SignupCallbackService instance.
+func NewSignupCallbackService(logger *log.Logger, config *configuration.Registry) *SignupCallbackService {
```
alexeykazakov: What this callback endpoint is about? Is it for checking the signup status? Why you call it "callback"? IMO it's a bit confusing name.

What if we use `POST /signup` for creating a signup/registering a new user and `GET /signup` for checking the signup status?

tinakurian: i agree with Alexey. Maybe we could call this `NewSignupService`. And below have the endpoints as `/signup` with the appropriate http verbs.

pkg/signup/signup_callback_handler_test.go
```diff
@@ -0,0 +1,45 @@
+package signup_test
+
+import (
+	"github.com/codeready-toolchain/registration-service/pkg/configuration"
```
tinakurian: can we organize these imports?

pkg/signup/signup_callback_handler.go
```diff
@@ -0,0 +1,34 @@
+package signup
+
+import (
+	"encoding/json"
+	"github.com/codeready-toolchain/registration-service/pkg/configuration"
```
tinakurian: can we organize these imports?

pkg/signup/signup_callback_handler_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/stretchr/testify/require"
+	"log"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"testing"
+)
+
+func TestSignupCallbackHandler(t *testing.T) {
+	// Create a request to pass to our handler. We don't have any query parameters for now, so we'll
```
tinakurian: can we be consistent with our comments? seems like some of them start with upper-case and others start with lower case..

pkg/signup/signup_handler.go
```diff
@@ -0,0 +1,34 @@
+package signup
+
+import (
+	"encoding/json"
+	"github.com/codeready-toolchain/registration-service/pkg/configuration"
```
tinakurian: can we organize these imports?

pkg/signup/signup_handler_test.go
```diff
@@ -0,0 +1,45 @@
+package signup_test
+
+import (
+	"github.com/codeready-toolchain/registration-service/pkg/configuration"
+	"github.com/codeready-toolchain/registration-service/pkg/signup"
```
tinakurian: as mentioned above, if we could organize these imports, that would be good.

pkg/signup/signup_handler_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"testing"
+)
+
+func TestSignupHandler(t *testing.T) {
+	// Create a request to pass to our handler. We don't have any query parameters for now, so we'll
+	// pass 'nil' as the third parameter.
+	req, err := http.NewRequest("GET", "/api/v1/signup", nil)
```
tinakurian: could we use http.MethodGet instead of "GET"?

pkg/signup/signup_handler.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+}
+
+// HandleRequest returns a default heath check result.
+func (srv *SignupService) HandleRequest(ctx *gin.Context) {
+	ctx.Writer.Header().Set("Content-Type", "application/json")
+
+	err := json.NewEncoder(ctx.Writer).Encode(nil)
+	if err != nil {
+		http.Error(ctx.Writer, err.Error(), 500)
```
tinakurian: can we use http.StatusInternalServerError instead of 500?

pkg/signup/signup_callback_handler_test.go
```diff
@@ -0,0 +1,45 @@
...
+	// pass 'nil' as the third parameter.
+	req, err := http.NewRequest("GET", "/api/v1/signup_callback", nil)
```
tinakurian: can we use http.MethodGet instead of "GET"?

pkg/signup/signup_callback_handler.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+}
+
+// HandleRequest returns a default heath check result.
+func (srv *SignupCallbackService) HandleRequest(ctx *gin.Context) {
+	ctx.Writer.Header().Set("Content-Type", "application/json")
+
+	err := json.NewEncoder(ctx.Writer).Encode(nil)
+	if err != nil {
+		http.Error(ctx.Writer, err.Error(), 500)
```
tinakurian: can we use http.StatusInternalServerError instead of 500?

pkg/signup/signup_callback_handler.go
tinakurian: agreed with @alexeykazakov we can remove the callback files and just add post/get and handle both scenarios withint he signup_handler.go

sbryzak: Michael and I discussed and agreed on these during our meeting on Friday.  The `/signup` endpoint is invoked as part of the landing page, it produces a redirect URL that sends the user to the RHD SSO to authenticate.  The `/signup-callback` endpoint is where the RHD login redirects the user to, it is this endpoint that will do the work of creating the `UserSignup` resource in the host cluster.  We will then have another endpoint, `check-status` which can be used to check the status of the user's provisioning request.

tinakurian: i think we should consider renaming the endpoints. isn't it possible that the user is just signing in and not necessarily signing up? also restful endpoints are meant to point to resources. so perhaps a better endpoint for /signup would be /users and a better endpoint for /signup-callback would be /usersignup since that is the resource we are creating..

alexeykazakov: Why we would need an endpoint for redirecting user to RHD in our backend at all?

Those `signup` and `check-status` endpoints were designed to be use to manage signups on the cluster side. This is why I suggested to actually renamed them to `POST /signup` and  `GET /signup`. It's all about SignUp resource management.

The redirect-callback concept is new to me and I'm not quite understand what it would give us.

Could you please elaborate on that?

cc: @michaelkleinhenz

tinakurian: after talking with @alexeykazakov we will need to remove the /signup-callback url. i think we should also rename the endpoint to /usersignup to avoid any confusion and make it clear that we are only handling the usersignup resource rather than the entire signup flow in the backend...

background: 
1. we want the registration app to be stateless, so we don't want to have to add back-end storage/database
2. because of [1.] we are planning to do the redirect in the ui so the flow is as follows: 
            2a. ui makes the call to rhd sso with a redirect url back to the ui
            2b. rhd sso will signup/login the user and provide back a code
            2c. ui makes another call to rhd sso via ajax with the code
            2d. rhd returns back to ui with the token
            2e. ui makes a call to /usersignup with token 
            2f. create usersignup resource and return with appropriate status code

2a-2d is handled by the javascript library.

PR #14, opened 2019-09-11
cleaning up registration app

tinakurian:
Relates to: https://jira.coreos.com/browse/CRT-240

cleaning up the repository by removing the callback endpoint, fixing comments, and using http consts where applicable


pkg/keycloak/keycloak_handler.go
```diff
@@ -0,0 +1,50 @@
+package keycloak
```
alexeykazakov: 1. There should not be any Keycloak specific code. We should not depend on the identity provider implementation unless we really have to.
Right now RHD supports OAuth2 and OpenID Connect specs. And it happens to run Keycloak. But we don't care.

2. Why would we want to have such endpoint to provide information about the Identity Provider in backend at all? Backend doesn't use this information. It's purely for UI. We could keep it in backend if we think we may need to use different clients which will consume such information (like different UIs/front-ends). So, was it your reasoning? Or something else?

michaelkleinhenz: Afaik we will need to have Keycloak-specific code in the project. While the JWT validation itself is generic, retrieving the public key is not imo.

pkg/health/health_check_handler.go
```diff
@@ -36,8 +36,8 @@ func (srv *Service) getHealthInfo() map[string]interface{} {
 }
 
 // HealthCheckHandler returns a default heath check result.
-func (srv *Service) HealthCheckHandler(ctx *gin.Context) {
-	// default handler for system health
+func (srv *HealthCheckService) GetHealthCheckHandler(ctx *gin.Context) {
```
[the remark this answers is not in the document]
tinakurian: Yes, i was thinking as what's done in Java. But furthermore, how would this look if we had to handle more than just a Get for a service? For instance, if in signup we have both a Get and Post how would the handle functions be named? I think there needs to be a way that is indicative of what http action the handler is handling...

pkg/health/health_check_handler.go
```diff
@@ -2,29 +2,29 @@ package health
...
 	"net/http"
 
 	"github.com/codeready-toolchain/registration-service/pkg/configuration"
+	"github.com/gin-gonic/gin"
 )
 
-// Service implements the service health endpoint.
-type Service struct {
+// HealthCheckService implements the service health endpoint.
+type HealthCheckService struct {
```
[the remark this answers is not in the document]
tinakurian: i was just copying what @sbryzak had introduced with his PR in the signup handler. I think the important part is consistency. I'm fine either way, but lets just agree to keep the code consistent. If we agreed to let in the signup handler with the struct type being "SignupService" when i think we should agree to make this the same... if that's not what we wanted then we should have made the change in that PR to be "Service" and "New" as was done in HealthCheck.

pkg/keycloak/keycloak_handler.go
michaelkleinhenz: Also, we need to supply the config to the JS Keycloak client. This can be done by:

1. burn the config static into the JS.
2. have it served as a static file from the backend.
3. have it generated from the backend config and served to the UI on an endpoint.

(1) is bad, it would mean to have configurable values compiled in or at least on two different locations (and not configurable using the config system, environment etc).

(2) is almost as bad as (1), it is just externalized from the JS, everything else from (1) applies as well.

(3) would be the way to have those config values unified with the rest of the config and managed and exposed in a sane way. BUT it requires to serve a Keycloak JSON from the backend.

I'd definitely say that a unified, managable config is more important than a generic backend. Especially because the UI *is* Keycloak-specific and an atomic part of the whole application.

michaelkleinhenz: If we want to go generic, we should dump that Keycloak client and use the standard OAuth2 flow (which was what we started with). But I admit the Keycloak JS client makes things much easier :-)

tinakurian: i think there are instances where static data that is purely used only for the ui are served via the back end. this can be seen in the static content hosting pattern for instance where style sheets, html pages, and even static drop down menus are populated via an endpoint. i think in our case, this json is very small and it can either sit in the ui or go in the back end. personally, i prefer to have it served..

alexeykazakov: > While the JWT validation itself is generic, retrieving the public key is not imo.

Retrieving the public keys is not Keycloak specific. It's JWK spec (see https://tools.ietf.org/html/rfc7517).
We have nothing KC specific in auth service and it works fine with RHD which runs Keycloak. But if RHD tomorrow switches to another implementation which still supports Oauth2, OpenID Connect, JWT, JWK, etc. specs then we should be good.

> Especially because the UI is Keycloak-specific and an atomic part of the whole application.

No. It's not. As I already mentioned multiple times we are going to use Keycloak JS Adapter not because it's KC specific. But because it actually works with services that implement Oauth2, OpenID Connect, JWT, JWK, etc. just fine. In addition to that the adapter has KC specific functionality too but we don't want/need to use those functionality. The generic part of KC Adapter is enough for us.
Many moons ago KC JS Adapter worked only with KC. But then we (OSIO-che team) modified it (contributed to upstream) so it can work with any auth service which implements the specs I mentioned above. For example we now use that JS adapter in che.openshift.io which talks to auth.openshift.io (our own go service which implements these specs. Not Keycloak!). We could use something else if we really wanted but why not using KC Adapter as a generic lib which works fine?

Regarding configuration for UI. UI will need configuration as any other our service. Auth is just one of piece of such configuration. It will also need to know URLs for reg-service, information about env (staging, prod, etc), etc. So, we should discuss what configuration we want to provide and how. But of course nothing should be hardcoded in JS/html.
But again it has nothing to do with KC specifics.

> i think in our case, this json is very small and it can either sit in the ui or go in the back end. personally, i prefer to have it served.. it just adds for more flexibility down the line. Perhaps one day these values will need to be extended, or perhaps one or more of these fields would be needed somewhere else.

This is a good point Tina. This is what I guessed about your motivation but wanted to confirm. The auth specific configuration would probably make sense to serve in reg-service since at least part of this config (like public keys URL) could be interesting to multiple clients (including the reg service itself). But we should not make it KC specific and we don't have to, really.

pkg/health/health_check_handler.go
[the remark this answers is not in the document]
tinakurian: i think it is also common practice even in golang to use Get, Create, Delete for handler function names.. see these examples: 

1. https://github.com/sarulabs/di-example/blob/master/app/handlers/cars.go
2. https://itnext.io/structuring-a-production-grade-rest-api-in-golang-c0229b3feedc
3. https://medium.com/@adigunhammedolalekan/build-and-deploy-a-secure-rest-api-with-go-postgresql-jwt-and-gorm-6fadf3da505b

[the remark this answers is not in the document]
alexeykazakov: Oh... this `Get` is for `GET` http method? Then yeah.. having it in the name could be good solution. Just not sure if we should add it in the beginning.. We can easily confuse it with getters/settes as I just did :) Maybe `HealthCheckGetHandler`?
Or add Get/Post/etc only when we need to have multiple methods like for `GetSignupHandler` and `PostSignupHandler`. But I'm fine with keeping it `GetHealtCheckHandler` too.

@sbryzak WDYT?

pkg/health/health_check_handler.go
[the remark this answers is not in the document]
alexeykazakov: I agree. Consistency is a must here. So, as I said if you think that using the long names would be better in other services for whatever reason (multiple services per package or something else) then let's use long names everywhere. It's minor anyway (consistency is not minor though).

pkg/health/health_check_handler.go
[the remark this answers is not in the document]
alexeykazakov: > i think it is also common practice even in golang to use Get, Create, Delete for handler function names.. see these examples:

Yep. It looks like it is. +1 to have it as `GetXXXHandler`, `PostXXXHandler`, etc.

pkg/health/health_check_handler.go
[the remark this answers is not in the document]
sbryzak: Within the signup package there are two services, `SignupService` and `SignupCallbackService`.  Generic struct and function names such as `Service` and `New` are not going to work here.

pkg/health/health_check_handler.go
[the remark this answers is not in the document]
sbryzak: I would avoid adding `Get` here also.  If for example we modified the endpoint to instead use `POST`, it would mean we would also need to rename this function.

pkg/server/routes.go
```diff
@@ -53,20 +53,16 @@ func (srv *RegistrationServer) SetupRoutes() error {
 
 		// /status is something you should always have in any of your services,
 		// please leave it as is.
-		healthService := health.New(srv.logger, srv.Config())
-		// TODO uncomment these once the services are available
-		//signupService := signup.NewSignupService(srv.logger, srv.Config())
-		//signupCallbackService := signup.NewSignupCallbackService(srv.logger, srv.Config())
```
sbryzak: Can we leave these lines here (commented out) until we are ready to implement them?  I am working on this endpoint this week.

pkg/signup/signup_callback_handler.go
```diff
@@ -1,34 +0,0 @@
...
-import (
-	"encoding/json"
-	"github.com/codeready-toolchain/registration-service/pkg/configuration"
-	"github.com/gin-gonic/gin"
-	"log"
-	"net/http"
-)
-
-// SignupCallbackService implements the signup callback endpoint.
-type SignupCallbackService struct {
```
sbryzak: Why was this file deleted?  This handler is intended to be the callback endpoint for which the SSO authentication will redirect the user to.

pkg/signup/signup_handler.go
```diff
@@ -22,13 +23,13 @@ func NewSignupService(logger *log.Logger, config *configuration.Registry) *Signu
 	}
 }
 
-// HandleRequest returns a default heath check result.
```
sbryzak: Thanks for fixing this for me. :)

pkg/signup/signup_handler.go
```diff
@@ -22,13 +23,13 @@ func NewSignupService(logger *log.Logger, config *configuration.Registry) *Signu
...
-func (srv *SignupService) HandleRequest(ctx *gin.Context) {
+// CreateSignupHandler returns signup info.
+func (srv *SignupService) CreateSignupHandler(ctx *gin.Context) {
```
sbryzak: I would leave the name of this function as it was.  It's not creating a signup handler here, it's handling the signup request (the handler was already *created* previously).

pkg/health/health_check_handler.go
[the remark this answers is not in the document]
sbryzak: Just to add to this, @tinakurian all of your examples are of REST services but the registration service isn't REST.  If we *were* developing a REST service, I would approach this differently.

pkg/keycloak/keycloak_handler.go
sbryzak: I need to add my +1 to not having any Keycloak-specific code here, at least in the Go code.  (This is speaking from experience working on fabric8-auth)

pkg/signup/signup_callback_handler.go
tinakurian: @sbryzak i had commented on your PR about the flow from discussing with @alexeykazakov . It can be found here: https://github.com/codeready-toolchain/registration-service/pull/13#discussion_r322421842 we don't need the SignupCallbackService..

pkg/signup/signup_handler.go
tinakurian: it's not about the handler, it's about the resource.. the resource is a POST, which would be to create... see design doc here: https://docs.google.com/document/d/19UzaXVqiOo9JqDYz3qUKguJwW-7zceA4UtTr358R_CA/edit#heading=h.olhyszn0h7id

pkg/health/health_check_handler.go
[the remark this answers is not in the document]
tinakurian: this is not quite right, see design doc here: https://docs.google.com/document/d/19UzaXVqiOo9JqDYz3qUKguJwW-7zceA4UtTr358R_CA/edit#heading=h.olhyszn0h7id and comment here: https://github.com/codeready-toolchain/registration-service/pull/13#discussion_r322421842

pkg/server/routes.go
tinakurian: we no longer need the signupCallbackService, see other comments for details

pkg/health/health_check_handler.go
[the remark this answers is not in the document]
tinakurian: @sbryzak i'm not sure i understand - this is definitely rest.

pkg/health/health_check_handler.go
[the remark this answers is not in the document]
sbryzak: So I just had a bluejeans call with @alexeykazakov and we are now on the same page in terms of design - Michael and I had previously discussed and agreed on which endpoints we would create and it seems we had all approached this from different angles. :)  Anyway, please disregard my previous comments about the `SignupCallback` handler, also after discussing with Alexey it seems after all that we will be going with a REST-style service, so +1 for the `Get` / `Post` function name prefixes also.

pkg/keycloak/keycloak_handler.go
tinakurian: it's removed, going to keep this PR to just cleanup

pkg/health/health_check_handler.go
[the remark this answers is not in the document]
tinakurian: just to be clear - going to keep the non-generic names and have also used get/post for handler names.

pkg/health/health_check_handler.go
```diff
@@ -2,29 +2,29 @@ package health
...
 }
 
-// New returns a new healthService instance.
-func New(logger *log.Logger, config *configuration.Registry) *Service {
-	r := new(Service)
-	r.logger = logger
-	r.config = config
-	return r
+// NewHealthCheckService returns a new HealthCheckService instance.
+func NewHealthCheckService(logger *log.Logger, config *configuration.Registry) *HealthCheckService {
```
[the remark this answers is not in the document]
tinakurian: resolved here: https://github.com/codeready-toolchain/registration-service/pull/14#discussion_r323537173

tinakurian:
thanks @sbryzak 🎉🍺

tinakurian:
thanks @alexeykazakov 🎉🍺

pkg/keycloak/keycloak_handler.go
michaelkleinhenz: @alexeykazakov ok, so the Keycloak client is a generic client. But to make it completely generic, we would have to provide the client.js with our service instead of pulling it "live" from the Keycloak instance. Providing it static is not recommended by the Keycloak doc for good reasons. Pulling it live on the other side would make the process Keycloak-specific. I guess no other implementation will provide a client JS with the same API on `/auth/js/keycloak.js`. Not nitpicking here, but I think this genericness discussion is somewhat over the top.

PR #17, opened 2019-09-12
AuthService Config Endpoint

michaelkleinhenz:
This adds the `/authconfig` endpoint to provide the configuration for the UI auth client.


example-config.yml
```diff
@@ -7,3 +7,9 @@ http.compress: false
 http.idle_timeout: "15s"
 http.write_timeout: "15s"
 http.read_timeout: "15s"
+auth_client.config.realm: "myRealm"
```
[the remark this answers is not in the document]
michaelkleinhenz: At least the realm is included in the generated config from the latest Keycloak as the OICD JSON. What we could do here is to put the complete config into an embedded JSON field in the configuration and just provide that JSON. Would remove the non-generic fields from the fixed code. Thoughts?

pkg/authconfig/authconfig_handler.go
```diff
@@ -0,0 +1,48 @@
...
+// New returns a new Service instance.
+func New(logger *log.Logger, config *configuration.Registry) *Service {
+	return &Service{
+		logger: logger,
+		config: config,
+	}
+}
+
+func (srv *Service) getAuthClientConfig() map[string]interface{} {
+	m := make(map[string]interface{})
```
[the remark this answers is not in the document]
michaelkleinhenz: I'll wait fixing this until https://github.com/codeready-toolchain/registration-service/pull/17#pullrequestreview-287551927 is decided.

example-config.yml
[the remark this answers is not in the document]
michaelkleinhenz: Bonus question. It looks like the publi key can be retrieved from `http://keycloak.server/auth/realms/<realm>`. The resulting structure looks to me like Keycloak specific. How should we deal with that?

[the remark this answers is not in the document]
michaelkleinhenz: Here's what my local Keycloak returns:
```
{"realm":"demoRealm","public_key":"MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAi04yxaQb7e1+tfcDoXe8K2DZ+rJ2yVVjBoT9Tw0jOout5w84x2/r5t/4aCBQjo9IO7UVWTtvE0cOk1WtykXvso7iYh9ry9jsZtrJNS0QXykcOJZJLVxyh1uatrbpM5heKYNz5fs5hp+3Qh5XkyCkLigIkOoLMXO1tLkNvjiEdR1zslqEOXaqWsp6HlUcu1JOuEv1LsxFuCnKc9ZvZDm0mQQJiOAl1QRvSU3pgX3IjuoefY6+6NQAYm1MQjOzWSnNkQTTEIWgIRu8QVgxko50pR3fTC7LWj6AQCv5GkW1r5zIv9OSzjoiN8A/UHASWh0Z6oy0eLeY775EhVfrg/KjYwIDAQAB","token-service":"http://localhost:8080/auth/realms/demoRealm/protocol/openid-connect","account-service":"http://localhost:8080/auth/realms/demoRealm/account","tokens-not-before":0}
```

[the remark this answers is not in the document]
michaelkleinhenz: There is also `http://localhost:8080/auth/realms/<realm>/protocol/openid-connect/certs` to retrieve the keys. This one looks generic. What is the recommended way of retrieving those?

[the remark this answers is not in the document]
alexeykazakov: We should be using `http://keycloak.server/auth/realms/<realm>/protocol/openid-connect/certs` one. It's a standard JWK thing. And this endpoint (among the others) is actually defined in the standard OIDC config endpoint - `<keycloakURL>/auth/realms/{realm}/.well-known/openid-configuration`

`http://keycloak.server/auth/realms/<realm>` is undocumented, KC specific and should not be used (as we were told by the KC guys).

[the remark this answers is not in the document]
michaelkleinhenz: That also does not look like what we need. The `fabric8-commons` code retrieves the keys from an endpoint that returns an array of elements of this format:
```
type JSONWebKey struct {
	Key          interface{}
	Certificates []*x509.Certificate
	KeyID        string
	Algorithm    string
	Use          string
}
```
The above endpoint returns a different format.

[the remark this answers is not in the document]
michaelkleinhenz: Yeah, it's already using the JWK url. The actual retrieval and parsing is in #21.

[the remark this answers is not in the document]
michaelkleinhenz: Fixed d7ee97e. As agreed last week, replaced the discrete config values with a json-based approach. The code also checks if the config json is valid json before sending it out.

pkg/authconfig/authconfig_handler.go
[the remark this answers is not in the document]
michaelkleinhenz: This does not apply anymore as we don't have the discrete config anymore. See d7ee97e.

pkg/authconfig/authconfig_handler.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// AuthconfigHandler returns signup info.
+func (srv *Service) AuthconfigHandler(ctx *gin.Context) {
+	ctx.Writer.Header().Set("Content-Type", "application/json")
+	ctx.Writer.WriteHeader(http.StatusOK)
+	authClientConfig := srv.getAuthClientConfig()
+	err := json.NewEncoder(ctx.Writer).Encode(authClientConfig)
+	if err != nil {
+		http.Error(ctx.Writer, err.Error(), http.StatusInternalServerError)
```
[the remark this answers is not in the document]
michaelkleinhenz: Fixed cb0210e.

michaelkleinhenz:
@xcoulon we're treating the auth client config as a blackbox, just giving it out to the UI. The goal is that the backend is independent from the UI-used client for the auth service. We store the config as an embedded JSON and just sending it out. See `example-config.yaml`. This was the result of a conversation we had late last week.

pkg/authconfig/authconfig_handler.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	return true
+}
+
+// AuthconfigHandler returns signup info.
+func (srv *Service) AuthconfigHandler(ctx *gin.Context) {
+	ctx.Writer.Header().Set("Content-Type", "application/json")
+	ctx.Writer.WriteHeader(http.StatusOK)
+	configStr := srv.config.GetAuthClientConfigAuthJSON()
+ 	if !srv.isJSON(configStr) {
```
alexeykazakov: Do we really care if it's JSON or not? We are returning here the raw config which is used only by front-end and which we just host in the backend as is, right? So, shouldn't we just return this raw string in the body here?
If we want to do some kind of validation of the config value then this should be done in the configuration package when we are loading the configuration. But not in the `/authconfig` handler every time this endpoint is called.

There should be some JIRA for improving our configuration loading. We need to add config validation when initializing our configuration and fail our health check if there is something wrong with the configuration, so we don't start a service in prod with invalid configuration. If we don't have such JIRA we should create it.

pkg/authconfig/authconfig_handler_test.go
```diff
@@ -0,0 +1,124 @@
...
+	// pass 'nil' as the third parameter.
+	req, err := http.NewRequest(http.MethodGet, "/api/v1/authconfig", nil)
+	require.NoError(t, err)
+
+	// Create logger and registry.
+	logger := log.New(os.Stderr, "", 0)
+	configRegistry := configuration.CreateEmptyRegistry()
+
+	// Set the config for testing mode, the handler may use this.
+	configRegistry.GetViperInstance().Set("testingmode", true)
```
alexeykazakov: I think it would be useful if we set the testingmode globally for our tests instead of setting it explicitly in every test which requires that mode.

There are multiple ways to achieve that. We could create a test suite (see https://godoc.org/github.com/stretchr/testify/suite) which we can re-use in all relevant tests and which will do some initialization for every test in the suite.

For example in `test` package:
```
type UnitTestSuite struct {
    suite.Suite
    Config *configuration.Registry
}

func (s *UnitTestSuite) SetupSuite() {
    s.Config = ... // create a config and set it up
}

func (s *UnitTestSuite) TearDownSuite() {
	s.Config = nil // Summon the GC!
}
```

Then we can use suite our tests:

```
type TestAuthClientConfigSuite struct {
	test.UnitTestSuite
}

func TestRunAuthClientConfigSuite(t *testing.T) {
	suite.Run(t, &TestAuthClientConfigSuite{&test.UnitTestSuite{}})
}

func (s *TestAuthClientConfigSuite) TestHandler() {
	// You can test the handler here. The configuration is available in s.Config
}
```

pkg/authconfig/authconfig_handler.go
michaelkleinhenz: I need to check, but I intentionally put the validation into the handler as I'd expect config changes be recognized on change, not on restart. But on the other hand, it's a good point that the config should not have any req's attached. I'll remove that check,

michaelkleinhenz: Created a story for the config check: https://jira.coreos.com/browse/CRT-255

michaelkleinhenz: Fixed 9259857. Also renamed the config key from ..json to ..raw.

pkg/authconfig/authconfig_handler_test.go
michaelkleinhenz: Makes sense, let's do this in a seperate PR. I created one with the base TestSuite and the AuthClientTest as an example. We can update all existing tests in that PR: https://github.com/codeready-toolchain/registration-service/pull/24

michaelkleinhenz:
@xcoulon the config is treated as a blackbox, it's format and content is solely depending on the UI, so we don't care about the actual content.

pkg/authconfig/authconfig_handler.go
michaelkleinhenz: Also added a `content-type` config value in bc604a2.

alexeykazakov: > I need to check, but I intentionally put the validation into the handler as I'd expect config changes be recognized on change, not on restart.

Unfortunately we can't do that.
We will use env vars with mapping to a ConfigMap in production to pass the configuration to viper/configuration. So, a pod/service restart is required to pick something new from a ConfigMap -> env var -> viper.
Also I remember viper had some race condition issues if we try to modify the underlying map of already-in-use viper instance. But even if it's not thread safe we still can't use it because of how ConfgiMap - env var mapping works in kube/openshift.

pkg/authconfig/authconfig_handler.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// AuthconfigHandler returns signup info.
+func (srv *Service) AuthconfigHandler(ctx *gin.Context) {
+	ctx.Writer.Header().Set("Content-Type", srv.config.GetAuthClientConfigAuthContentType())
+	ctx.Writer.WriteHeader(http.StatusOK)
+	_, err := ctx.Writer.WriteString(srv.config.GetAuthClientConfigAuthRaw())
+	if err != nil {
+		srv.logger.Println("error writing response body", err.Error())
+		http.Error(ctx.Writer, err.Error(), http.StatusInternalServerError)
```
alexeykazakov: As I mentioned before I think we should return payloads in JSON format instead of raw strings (with some exceptions like authconfig payload). Including error responses. So, we will need to introduce `Error` struct which we will marshal to JSON and return in the body. Something like:
```
type Error struct {
    Status string
    Code string
    Title string
    Details string
    ...
}
```
And then construct it when needed: `NewError(http.StatusInternalServerError, "oopsie woopsie... some human readable title", err)`

But let's do it in a separate PR (would need a JIRA for that :) )

pkg/authconfig/authconfig_handler_test.go
```diff
@@ -0,0 +1,110 @@
...
+		require.Equal(t, configRegistry.GetAuthClientConfigAuthContentType(), rr.Header().Get("Content-Type"))
+
+		// Check the response body is what we expect.
+		// get config values from endpoint response
+		var data map[string]interface{}
+		err = json.Unmarshal(rr.Body.Bytes(), &data)
+		require.NoError(t, err)
+
+		// get the configured values
+		var config map[string]interface{}
```
alexeykazakov: Ideally the configuration value tests should be in configuration tests. Not in the handler tests (because handlers doesn't care. It's just a raw string for the handler) and you already tested the strings in https://github.com/codeready-toolchain/registration-service/pull/17/files#diff-557f6e3c317a9e7443fff482bdcf1fc1R49

But I'm OK to keep it here :)

test/unittestsuite.go
```diff
@@ -0,0 +1,34 @@
+package testutils
```
alexeykazakov: I thought you said you want this Suite in a separate PR? :)

pkg/controller/authconfig.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+// AuthConfig implements the auth config endpoint, which is invoked to 
+// retrieve the auth config for the ui.
+type AuthConfig struct {
+	config *configuration.Registry
+	logger *log.Logger
+}
+
+// NewAuthConfig returns a new Service instance.
```
alexeykazakov:
```suggestion
// NewAuthConfig returns a new AuthConfig instance.
```

pkg/controller/authconfig.go
```diff
@@ -0,0 +1,36 @@
...
+func NewAuthConfig(logger *log.Logger, config *configuration.Registry) *AuthConfig {
+	return &AuthConfig{
+		logger: logger,
+		config: config,
+	}
+}
+
+// AuthconfigHandler returns signup info.
```
alexeykazakov:
```suggestion
// GetHandler returns raw auth config content for UI.
```

pkg/controller/authconfig.go
tinakurian: Fixed here: 15c63cd12305a66b053f4c1ddd92bbe52f72427f

pkg/controller/authconfig.go
tinakurian: good catch @alexeykazakov fixed here: 15c63cd12305a66b053f4c1ddd92bbe52f72427f

pkg/authconfig/authconfig_handler.go
tinakurian: created a jira for this here: https://jira.coreos.com/browse/CRT-256 :-)

test/unittestsuite.go
michaelkleinhenz: Yes. And there is already one. Could be that there is some mixup here. Will check.

pkg/authconfig/authconfig_handler.go
xcoulon: @alexeykazakov @michaelkleinhenz Actually, it should work now: Viper can now detect changes in Config Maps and update the config accordingly. I fixed it in https://github.com/spf13/viper/pull/415. It's worth the try to verify, though ;)

test/unittestsuite.go
michaelkleinhenz: Ok, somehow that testsuite file ended up in this PR, all other changes for the suites are where they should be. Removed it in 97c2f18

PR #19, opened 2019-09-13
Dynamic Auth Library Loading

michaelkleinhenz:
This adds code to load the JS Auth Client dynamically from a configurable URL. This needs to be updated when https://github.com/codeready-toolchain/registration-service/pull/17 is merged.

Open question: this assumes that the key `library_url` is part of the auth config retrieved from the backend (see https://github.com/codeready-toolchain/registration-service/pull/17). This then possibly needs the JS to delete that key from the config before giving it to the JS client as his configuration. Another possible solution for dynamically loading the Auth Client Library would be to add `<script src='/somelocal.js'>` to the HTML and then redirect that query from the backend to the final URL.

Any thoughts on which variant would be better?


michaelkleinhenz:
My opinion: I think both ways are valid, I would prefer the current one as it is already there.

michaelkleinhenz:
As per meeting last week, we will leave it as it is right now, no redirect.

pkg/assets/index.html
```diff
@@ -1,3 +1,46 @@
...
+    script.onreadystatechange = loadFunction;
+    document.getElementsByTagName("head")[0].appendChild(script);
+  };
+
+  // load   
+  loadAuthLibrary(authLibraryURL, function() {
+    // success
+    alert("Library load success!")
+    // do something with the library
+    fooBar()
```
tinakurian: fooBar() can probably be removed :-)

pkg/configuration/configuration.go
```diff
@@ -66,6 +66,10 @@ const (
 	// DefaultTestingMode specifies whether the services should run in testing mode.
 	DefaultTestingMode = false
 
+	varAuthClientLibraryURL = "auth_client.library_url"
+	// DefaultAuthClientLibraryURL is the default auth library location.
+	DefaultAuthClientLibraryURL = "https://keycloak.service/auth/js/keycloak.js"
```
xcoulon: I don't understand why you use such a default value. It's not valid, is it?

xcoulon: (same question for the `DefaultAuthClientPublicKeysURL` below)

pkg/assets/index.html
michaelkleinhenz: That JS needs to be reworked entirely. This is the next task after this one here is merged.

pkg/configuration/configuration.go
michaelkleinhenz: Yeah, there is no valid url for that and the default will not work. This is sort of a mandatory config. The key url was updated in the other PR to be default "", so that by default no keys are loaded.

PR #20, opened 2019-09-13
Configuration for Public Key URL

michaelkleinhenz:
This adds a new config option for the public key retrieval.

### DIFF
--- a/example-config.yml
+++ b/example-config.yml
@@ -7,3 +7,4 @@ http.compress: false
 http.idle_timeout: "15s"
 http.write_timeout: "15s"
 http.read_timeout: "15s"
+auth_client.public_keys_url: "http://localhost:8080/auth/realms/demoRealm/protocol/openid-connect/certs"
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -65,6 +65,10 @@ const (
 	varTestingMode = "testingmode"
 	// DefaultTestingMode specifies whether the services should run in testing mode.
 	DefaultTestingMode = false
+
+	varAuthClientPublicKeysURL = "auth_client.public_keys_url"
+	// DefaultAuthClientPublicKeysURL is the default log level used in your service.
+	DefaultAuthClientPublicKeysURL = "http://keycloak.service/auth/realms/myRealm/protocol/openid-connect/certs"
 )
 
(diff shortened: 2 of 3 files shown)


example-config.yml
```diff
@@ -7,3 +7,4 @@ http.compress: false
 http.idle_timeout: "15s"
 http.write_timeout: "15s"
 http.read_timeout: "15s"
+auth_client.public_keys_url: "http://localhost:8080/auth/realms/demoRealm/protocol/openid-connect/certs"
```
xcoulon: side question: what's the benefit of such an `example-config.yml` file, since we're going to use env vars when deploying on OpenShift, and there's a default value configured in https://github.com/codeready-toolchain/registration-service/pull/20/files#diff-ec24f891497e9e435f8f9956e23cb379R71 ? 

Also, what config file will be provided in here https://github.com/codeready-toolchain/registration-service/blob/09ec1dcf76c340c2c75e33e8898a537254c04aca/pkg/server/server.go#L28 when the application will be deployed on staging and production ?
I wonder if the `server.NewServer(configfile string)` should be converted to something like `server.NewServer(config configuration. Registry)` instead, to be more flexible.
(Also, I'm responsible for the `configuration.Registry` type name back in the f8 days, but I wonder if `configuration.Configuration` wouldn't be more appropriate - naming is hard! 😂)

michaelkleinhenz: First question: the config file is an alternative way of specifying the config params. That whole system was taken from Konrad's code on openshift.io and we (or at least I am) using the config file for the local dev cycle. Alas, the config file is an option for the configuration, so we should imo provide an example.

Second: the configFilePath is empty when there is no file given and the configuration subsystem handles that (as well as the precedence between config file and env).

And I am not saying we shouldn't do some improvements here, but let's di this in a seperate PR as this one is dealing with the key retrieval.

alexeykazakov: @michaelkleinhenz we actually used to have config files in fabric8-services too. But it turned out to be an option we never used :) It just complicated our code and we didn't have a single use case when such file could be useful.
They were not used in tests. Nor in dev env. Nor even as examples because in production the configuration comes from env vars via OpenShift Templates & Secretes & ConfigMaps. Not from config files.

So +1 to getting rid of config files but in a separate PR :)

PR #21, opened 2019-09-13
Public Key Retrieval, Management and JWT Validation

michaelkleinhenz:
This adds the retrieval of public keys for token verification from the auth server and the JWT token validation.


pkg/signup/keymanager.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		return nil, err
+	}
+	res, err := httpClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	// cleanup and close after being done
+	defer func() {
+		ioutil.ReadAll(res.Body)
```
golangcibot: Error return value of `ioutil.ReadAll` is not checked (from `errcheck`)

go.mod
```diff
@@ -3,19 +3,27 @@ module github.com/codeready-toolchain/registration-service
 go 1.12
 
 require (
+	github.com/dgrijalva/jwt-go v3.2.0+incompatible
+	github.com/dnaeon/go-vcr v1.0.1 // indirect
 	github.com/evanphx/json-patch v4.5.0+incompatible // indirect
+	github.com/fabric8-services/fabric8-auth-client v0.0.0-20190313133658-59d93fb7492b
+	github.com/fabric8-services/fabric8-common v0.0.0-20190529115613-c08218d5adf2
```
alexeykazakov: You are copy-pasted the fetching key thing from https://github.com/fabric8-services/fabric8-common/blob/b10e057d860d730661b2440dc1326c7d82606acc/auth/jwk/fetch_keys.go (which is fine), so, I guess we don't need any fabric8-* stuff here? Just wonder if it's some leftovers from you local experiments.

michaelkleinhenz: Yeah, might be. Will fix that.

go.mod
```diff
@@ -3,19 +3,27 @@ module github.com/codeready-toolchain/registration-service
...
 	github.com/gin-contrib/gzip v0.0.1
 	github.com/gin-contrib/sse v0.1.0 // indirect
 	github.com/gin-gonic/gin v1.4.0
+	github.com/goadesign/goa v2.0.5+incompatible
```
xcoulon: I thought we didn't use goa?

pkg/server/routes.go
```diff
@@ -54,8 +54,12 @@ func (srv *RegistrationServer) SetupRoutes() error {
 		// /status is something you should always have in any of your services,
 		// please leave it as is.
 		healthService := health.NewHealthCheckService(srv.logger, srv.Config())
-		signupService := signup.NewSignupService(srv.logger, srv.Config())
-
+		signupService, err := signup.NewSignupService(srv.logger, srv.Config())
```
xcoulon: don't you need to bind the service to an URL template, too?

pkg/signup/keymanager.go
```diff
@@ -0,0 +1,172 @@
...
+
+// JSONKeys the remote keys encoded in a json document
+type JSONKeys struct {
+	Keys []interface{} `json:"keys"`
+}
+
+// KeyManager manages the public keys for token validation.
+type KeyManager struct {
+	config *configuration.Registry
+	logger *log.Logger
```
xcoulon: why do you need a pointer to the logger?

pkg/signup/keymanager.go
```diff
@@ -0,0 +1,172 @@
+package signup
+
+import (
+	"bytes"
+	"crypto/rsa"
+	"encoding/json"
+	"errors"
+	"io"
+	"io/ioutil"
+	"log"
```
xcoulon: shouldn't you use `github.com/sirupsen/logrus` instead?

pkg/signup/keymanager.go
```diff
@@ -0,0 +1,172 @@
...
+		config: config,
+		keyMap: make(map[string]*rsa.PublicKey),
+	}
+	// fetch raw keys
+	var keys []*PublicKey
+	var err error
+	if !isTesting {
+		keys, err = km.fetchKeys(keysEndpointURL)	
+	} else {
+		keysRaw := config.GetViperInstance().Get("testkeys")
```
xcoulon: what about adding a method to get these keys from the config, instead of having to use the underlying `viper` lib?

pkg/signup/keymanager_test.go
```diff
@@ -0,0 +1,90 @@
...
+	jwt1 := "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImtleTEifQ.eyJqdGkiOiIwMjgyYjI5Yy01MTczLTQyZDgtODE0NS1iNDVmYTFlMzUzOGMiLCJleHAiOjAsIm5iZiI6MCwiaWF0IjoxNTE3MDE1OTUyLCJpc3MiOiJ0ZXN0IiwiYXVkIjoiY29kZXJlYWR5LXJlZ2lzdHJhdGlvbi1zZXJ2aWNlIiwic3ViIjoiMjM5ODQzOTgtODU1YS00MmQ2LWE3ZmUtOTM2YmI0ZTkyYTBkIiwidHlwIjoiQmVhcmVyIiwic2Vzc2lvbl9zdGF0ZSI6ImVhZGMwNjZjLTEyMzQtNGE1Ni05ZjM1LWNlNzA3YjU3YTRlMCIsImFjciI6IjAiLCJhbGxvd2VkLW9yaWdpbnMiOlsiKiJdLCJhcHByb3ZlZCI6dHJ1ZSwiZW1haWxfdmVyaWZpZWQiOnRydWUsIm5hbWUiOiJUZXN0MSBVc2VyMSIsImNvbXBhbnkiOiJUZXN0IENvbXBhbnkgMSIsInByZWZlcnJlZF91c2VybmFtZSI6InRlc3R1c2VyMSIsImdpdmVuX25hbWUiOiJUZXN0MSIsImZhbWlseV9uYW1lIjoiVXNlcjEiLCJlbWFpbCI6InRlc3R1c2VyMUB0ZXN0LnQifQ.XoLc1_ESvotJwwfK-zsyu4wySeFalGuHWB1cHRVBPKmztOPgQUGOb4zplhyKAuPPf0x3WGV50ESNW9t-YZoD-DJceMJY_AOzt0LBAoGoeovsVHbrHNGgMDEJsgjQd_beMQsqVGeJrReN9hnqZ8iPz1itMzzTUskG1TQylcr_ez4"
+
+	// Set the config for testing mode, the handler may use this.
+	configRegistry.GetViperInstance().Set("testingmode", true)
```
xcoulon: same here: what about having proper methods instead of having to access the underlying "viper" layer?

pkg/signup/signup_handler.go
```diff
@@ -5,22 +5,31 @@ import (
 	"log"
 	"net/http"
 
-	"github.com/codeready-toolchain/registration-service/pkg/configuration"
 	"github.com/gin-gonic/gin"
+
+	"github.com/codeready-toolchain/registration-service/pkg/configuration"
```
xcoulon: this import should be in the "second" section, ie, above the `github.com/gin-gonic/gin`, to follow our way of sorting/organizing imports

pkg/signup/tokenparser.go
```diff
@@ -0,0 +1,92 @@
+package signup
+
+import (
+	"errors"
+	"log"
+
+	jwt "github.com/dgrijalva/jwt-go"
+
+	"github.com/codeready-toolchain/registration-service/pkg/configuration"
```
xcoulon: same here (organizing imports)

pkg/signup/tokenparser.go
```diff
@@ -0,0 +1,92 @@
...
+
+// Permissions represents a "permissions" claim in the AuthorizationPayload
+type Permissions struct {
+	ResourceSetName *string  `json:"resource_set_name"`
+	ResourceSetID   *string  `json:"resource_set_id"`
+	Scopes          []string `json:"scopes"`
+	Expiry          int64    `json:"exp"`
+}
+
+// TokenParser represents a token.
```
xcoulon: comment looks wrong

pkg/signup/tokenparser_test.go
```diff
@@ -0,0 +1,68 @@
...
+	]}`
+	// test JWT, signed with above keys
+	jwt0 := "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImtleTAifQ.eyJqdGkiOiIwMjgyYjI5Yy01MTczLTQyZDgtODE0NS1iNDVmYTFlMzUzOGIiLCJleHAiOjAsIm5iZiI6MCwiaWF0IjoxNTE3MDE1OTUyLCJpc3MiOiJ0ZXN0IiwiYXVkIjoiY29kZXJlYWR5LXJlZ2lzdHJhdGlvbi1zZXJ2aWNlIiwic3ViIjoiMjM5ODQzOTgtODU1YS00MmQ2LWE3ZmUtOTM2YmI0ZTkyYTBjIiwidHlwIjoiQmVhcmVyIiwic2Vzc2lvbl9zdGF0ZSI6ImVhZGMwNjZjLTEyMzQtNGE1Ni05ZjM1LWNlNzA3YjU3YTRlOSIsImFjciI6IjAiLCJhbGxvd2VkLW9yaWdpbnMiOlsiKiJdLCJhcHByb3ZlZCI6dHJ1ZSwiZW1haWxfdmVyaWZpZWQiOnRydWUsIm5hbWUiOiJUZXN0MCBVc2VyMCIsImNvbXBhbnkiOiJUZXN0IENvbXBhbnkgMCIsInByZWZlcnJlZF91c2VybmFtZSI6InRlc3R1c2VyMCIsImdpdmVuX25hbWUiOiJUZXN0MCIsImZhbWlseV9uYW1lIjoiVXNlcjAiLCJlbWFpbCI6InRlc3R1c2VyMEB0ZXN0LnQifQ.a2JbOJXEB7IAK4JUEcW886VNJoQeHJ_yQiN5dgoPECpKw7PdQAajVBEnSRj6irDsNCjBznDnthSg_rfRv10_cN5K--FoB3eynylMm2Zwdj1JiVpiYd-lK0OfdLUfXlWalq-GLAoiI0UzDBjQE_Izzr58KaeYzjP7Gb9et99zVLDpnmLCoa9vCrXL7G-Dir4l2CqVtorp653Y19EY-A4TeJEitrqg4Uc8SXGskgpXnDZogX0IzFnwvDXawxeUUUoVaSugaxeyAh-ZyFMOFMtHXOM5M-7_3_kMMeVlKrCrbMRN6as5-jQ2vyrZND6fnVuZhrGgoIZNYfN5qaJHgpMikQ"
```
xcoulon: you could probably use constants here and share them with between the `tokenparser_test.go` and `keymanager_test.go`

go.mod
michaelkleinhenz: Even then, this should be updated on a build. Wondering why this does not work...

go.mod
michaelkleinhenz: We don't. I am looking at why this is not updating on build.

pkg/server/routes.go
michaelkleinhenz: Yes, that is done a few lines down..

pkg/signup/keymanager.go
michaelkleinhenz: What is the alternative? We're always handing around pointers to the logger there as the logger is created with some params in the server struct..

pkg/signup/keymanager.go
michaelkleinhenz: This is what was used in the initial scaffolding. I did not change something there..

pkg/signup/keymanager.go
michaelkleinhenz: Not in this case. This code is only used in the test run where there is no full config instance but an "injected" one. Also, we don't want config values that are only used for testing to be mandatory in a production config, so the "main" config should not be used to distribute the test keys.

pkg/signup/keymanager_test.go
michaelkleinhenz: See above, this would spoil the production code with config settings that are mandatory and only used in the test case.

pkg/signup/signup_handler.go
michaelkleinhenz: mmh..I always had it vice versa, the most generic on top, then the more specific and the local, most specific stuff on the bottom..

pkg/signup/tokenparser_test.go
michaelkleinhenz: I intentionally did not do that as there may be more and/or different tokens coming in in later PRs.

pkg/signup/keymanager.go
michaelkleinhenz: Fixed.

go.mod
michaelkleinhenz: Fixed 31a3d4a.

go.mod
michaelkleinhenz: Fixed 31a3d4a.

pkg/signup/signup_handler.go
michaelkleinhenz: Fixed 5f83a27.

pkg/signup/tokenparser.go
michaelkleinhenz: Fixed 5f83a27.

pkg/signup/tokenparser.go
michaelkleinhenz: Fixed 5f83a27.

pkg/signup/keymanager.go
xcoulon: what about replacing the keys in the tests, instead of using a different set of keys?

pkg/signup/keymanager.go
xcoulon: ok, but that's the logger impl we've used on the fabric8 services, and I believe we could (should?) keep using it here.

pkg/signup/keymanager.go
xcoulon: ok, I guess it makes it "shared" amongst services, which may be better if it manages file descriptors..

pkg/server/routes.go
xcoulon: ok, sorry, I missed that

pkg/signup/tokenparser_test.go
xcoulon: ok

pkg/signup/keymanager.go
michaelkleinhenz: Let's do this in a seperate PR then. This would touch more than just the key management code.

pkg/signup/keymanager.go
michaelkleinhenz: Not possible as those are retrieved from an http endpoint and this is the whitebox test. There is another issue here, there are intentionally no setter for config values. When we want to set a value from the test, we need to access the underlying infra.

pkg/signup/signup_handler.go
```diff
@@ -5,22 +5,31 @@ import (
...
 // NewSignupService returns a new SignupService instance.
-func NewSignupService(logger *log.Logger, config *configuration.Registry) *SignupService {
-	return &SignupService{
+func NewSignupService(logger *log.Logger, config *configuration.Registry) (*SignupService, error) {
+	service := &SignupService{
 		logger: logger,
 		config: config,
 	}
+	// create new KeyManager
+	keyManager, err := NewKeyManager(logger, config)
```
alexeykazakov: Instead of creating a new manager (which means we will fetch keys every time we create a service) for individual services we should use the only manager for all the services. We can initialize it in main and then re-use in all the services.
We would want to use the only instance of the manager for multiple reasons. Performance (don't fetch keys multiple times when not needed). Code-reusage. Also, our manager will need to evolve in the future. We will need to have some internal cache for the keys which we will update from time to time (keys could rotate in RHD). And having just one instance will be preferable for that.

Or we could keep introduce a singleton manager similar to what we use in fabric8-* services - https://github.com/fabric8-services/fabric8-common/blob/2cc2344d55776abe18178fc1c2efa8abc9996565/auth/token.go#L58
Then you don't have to pass the manager around. You just initialize it in main and then access everywhere via `auth.DefaultManager(config)`.

pkg/signup/tokenparser_test.go
alexeykazakov: While I think it's OK to hardcode some tokens in our tests for testing invalid token signatures it would be pretty hard to maintain the tests if we use hardcoded tokens which are supposed to have valid signatures.
Instead we should generate tokens in our tests programmaticaly. Take a look at https://github.com/fabric8-services/fabric8-common/blob/master/test/auth/auth.go for example. There are a bunch of utils which helps to generate tokens for tests using pre-defined test keys. So, in your tests you can always focus on the token content. 

And if you need to change some token claim in the test you don't need to manually re-encode the token. It also helps a lot with code readability. And if you need to change the test key you don't have to re-encode all the tokens, etc.

test/tokengenerator.go
```diff
@@ -0,0 +1,141 @@
...
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		keySet := &WebKeySet{}
+		for kid, key := range tg.keyMap {
+			thisKey, err := jwk.New(&key.PublicKey)
+			if err != nil {
+				fmt.Printf("fatal error adding keys to key service: %s", err.Error())
+				os.Exit(-1)
+			}
+			thisKey.Set(jwk.KeyIDKey, kid)
```
golangcibot: Error return value of `thisKey.Set` is not checked (from `errcheck`)

pkg/signup/tokenparser_test.go
michaelkleinhenz: Done 5f83a27. If the goal there was to make things easier and light, we failed :-)

pkg/signup/signup_handler.go
michaelkleinhenz: I implemented this in f32824a, but I am not sure if that retrieval makes sense. The `func DefaultManager(config ManagerConfiguration)` uses `sync.Once`. That means repeated calls to it can be done, but have no effect. If that would be used to also *retrieve* the instance, then the caller always has to give a `config`, even if that is not used and has no effect on updated config values. I guess the rest of the code should retrieve the instance by directly accessing the `Manager` field, but still, any updated config and call to `DefaultManager()` would result in a silent ignore. That looks like a bogus design to me.

I updated this in 9560d40 to a different impl, that has an additional method to retrieve the KeyManager from global (and that one reminded me that golang has no polymophic method signatures, doh).

pkg/signup/keymanager_test.go
```diff
@@ -208,20 +196,26 @@ func TestKeyManagerKeyParsing(t *testing.T) {
...
 		}
 		for _, tt := range statictests {
 			t.Run(tt.name, func(t *testing.T) {
 				_, err = jwt.Parse(tt.jwt, func(token *jwt.Token) (interface{}, error) {
-					return keyManager.Key(tt.kidToUse)
+					kid := token.Header["kid"]
+					require.NotEqual(t, tt.kid, kid)
+					return keyManager.Key(tt.kid)
 				})
 				require.Error(t, err)
```
alexeykazakov: Worth checking the error message too to make sure you got the expected token validation error here.

pkg/signup/tokenparser_test.go
```diff
@@ -55,23 +89,23 @@ func TestTokenParser(t *testing.T) {
...
 				require.Equal(t, tt.username, claims.Username)
 				require.Equal(t, tt.email, claims.Email)
 			})
 		}
 	})
 
 	t.Run("parse invalid token", func(t *testing.T) {
-		_, err := tokenParser.FromString(jwt2)
+		_, err := tokenParser.FromString(jwt_invalid)
 		require.Error(t, err)
```
alexeykazakov: Check the error message here too.

test/tokengenerator.go
```diff
@@ -0,0 +1,141 @@
...
+	token := jwt.New(jwt.SigningMethodRS256)
+	token.Claims.(jwt.MapClaims)["uuid"] = identity.ID
+	token.Claims.(jwt.MapClaims)["preferred_username"] = identity.Username
+	token.Claims.(jwt.MapClaims)["sub"] = identity.ID
+	token.Claims.(jwt.MapClaims)["jti"] = uuid.NewV4().String()
+	token.Claims.(jwt.MapClaims)["session_state"] = uuid.NewV4().String()
+	token.Claims.(jwt.MapClaims)["iat"] = time.Now().Unix()
+	token.Claims.(jwt.MapClaims)["exp"] = time.Now().Unix() + 60*60*24*30
+	token.Claims.(jwt.MapClaims)["nbf"] = 0
+	token.Claims.(jwt.MapClaims)["iss"] = "fabric8-auth"
```
alexeykazakov: Copy-paste.

test/tokengenerator.go
```diff
@@ -0,0 +1,141 @@
...
+	token.Claims.(jwt.MapClaims)["typ"] = "Bearer"
+	token.Claims.(jwt.MapClaims)["approved"] = true
```
alexeykazakov: We won't have this "approved" claim in our prod tokens. It's fabric8-auth specific thing. It doesn't harm though.

test/tokengenerator.go
```diff
@@ -0,0 +1,141 @@
...
+				fmt.Printf("fatal error adding keys to key service: %s", err.Error())
+				os.Exit(-1)
+			}
+			thisKey.Set(jwk.KeyIDKey, kid)	
+			keySet.Keys = append(keySet.Keys, thisKey)
+		}
+		jsonKeyData, err := json.Marshal(keySet)
+		if err != nil {
+			fmt.Printf("fatal error creating key service: %s", err.Error())
+			os.Exit(-1)
```
alexeykazakov: Panic. Don't exit. (@xcoulon wink wink)

test/tokengenerator.go
```diff
@@ -0,0 +1,141 @@
...
+func (tg *TokenGenerator) GetKeyService() string {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		keySet := &WebKeySet{}
+		for kid, key := range tg.keyMap {
+			thisKey, err := jwk.New(&key.PublicKey)
+			if err != nil {
+				fmt.Printf("fatal error adding keys to key service: %s", err.Error())
+				os.Exit(-1)
```
alexeykazakov: Panic. Don't exit.812180

pkg/server/routes.go
```diff
@@ -54,8 +55,12 @@ func (srv *RegistrationServer) SetupRoutes() error {
...
+		if err != nil {
+			// error creating signup service, bail out.
+			log.Fatal(err)
```
alexeykazakov: Shouldn't we return the error here?

pkg/signup/tokenparser_test.go
alexeykazakov: Looks better now. Thanks.

> If the goal there was to make things easier and light, we failed 

I will ask you again if it's easier to use when you will be adding new tests for future code or modifying the existing ones or debugging them ;)
No any manual token generation and parsing needed anymore.

michaelkleinhenz:
> Hm... I don't see the key manager anymore at all :)

Moved to auth package..

pkg/signup/keymanager_test.go
michaelkleinhenz: Fixed 1786a4d.

pkg/signup/tokenparser_test.go
michaelkleinhenz: Fixed 99f4773.

test/tokengenerator.go
michaelkleinhenz: Fixed 49b23c7.

test/tokengenerator.go
michaelkleinhenz: Fixed with the other `panic()`s.

test/tokengenerator.go
michaelkleinhenz: Fixed.

pkg/server/routes.go
michaelkleinhenz: This part was restructured by Tina anyway..

michaelkleinhenz:
Closing this as it was split up in #26. #27 and #28.

PR #23, opened 2019-09-17
Add signup_client utility class

sbryzak:
Fixes:

- https://jira.coreos.com/browse/CRT-225
- https://jira.coreos.com/browse/CRT-245

This pull request introduces a utility class that handles the creation of the UserSignup resource. 
 It converts the user's username to a valid resource name (DNS-1123 compliant), and uses the Kubernetes client library to interact with the host cluster.


go.mod
```diff
@@ -30,10 +30,11 @@ require (
 	golang.org/x/tools v0.0.0-20190909030654-5b82db07426d // indirect
 	google.golang.org/appengine v1.6.1 // indirect
 	gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15 // indirect
-	gopkg.in/inf.v0 v0.9.1 // indirect
 	k8s.io/api v0.0.0-20190820101039-d651a1528133 // indirect
 	k8s.io/apimachinery v0.0.0-20190823012420-8ca64af22337 // indirect
-	k8s.io/client-go v0.0.0-20190620085101-78d2af792bab
+	k8s.io/client-go v0.0.0-20190819141724-e14f31a72a77
```
alexeykazakov: The new client set API in client-go looks very nice. But let’s make sure it’s available in OpenShift 4.2.
OpenShift 4.2 will be based on Kubernetes 1.14. See the corresponding client-go version - https://github.com/kubernetes/client-go/tree/kubernetes-1.14.6 (sha: 7e43eff).

We will need to pin all our kube dependencies to 1.14 version for now. It may break our minishift compatibility though.

tinakurian:
> there is no `class` in Go ;)

Reading through go doc i think sort of insists that the concept of what a class is still exists within go. It just doesn't quite look the _exact_ same as in other languages. See here: http://www.golangpatterns.info/object-oriented/classes and even go doc uses Structs and Classes _somewhat_ interchangeably https://golang.org/doc/faq#Is_Go_an_object-oriented_language. I think this makes sense when we are talking about OO languages - a class is really just a blueprint.

xcoulon:
> > there is no `class` in Go ;)
> 
> @xcoulon I've read some really interesting blogs and go doc articles that help make this more clear. Reading through go doc i think sort of insists that the concept of what a class is still exists within go. It just doesn't quite look the _exact_ same as in other languages. See here: http://www.golangpatterns.info/object-oriented/classes and even go doc uses Structs and Classes _somewhat_ interchangeably https://golang.org/doc/faq#Is_Go_an_object-oriented_language. I think this makes sense when we are talking about OO languages - a class is really just a blueprint.

My apologies, I did not mean to be pedantic here. I thought that correct term in Go was `type`/`struct` instead of `classes`, that's all ;)

tinakurian:
> > > there is no `class` in Go ;)
> > 
> > 
> > @xcoulon I've read some really interesting blogs and go doc articles that help make this more clear. Reading through go doc i think sort of insists that the concept of what a class is still exists within go. It just doesn't quite look the _exact_ same as in other languages. See here: http://www.golangpatterns.info/object-oriented/classes and even go doc uses Structs and Classes _somewhat_ interchangeably https://golang.org/doc/faq#Is_Go_an_object-oriented_language. I think this makes sense when we are talking about OO languages - a class is really just a blueprint.
> 
> My apologies, I did not mean to be pedantic here. I thought that correct term in Go was `type`/`struct` instead of `classes`, that's all ;)

hmmm? noooo don't apologize! i was just sharing something interesting about the class idea. just seems weird to me the interchanging use. wasn't trying to say you were being pedantic!!

xcoulon:
> > > > there is no `class` in Go ;)
> > > 
> > > 
> > > @xcoulon I've read some really interesting blogs and go doc articles that help make this more clear. Reading through go doc i think sort of insists that the concept of what a class is still exists within go. It just doesn't quite look the _exact_ same as in other languages. See here: http://www.golangpatterns.info/object-oriented/classes and even go doc uses Structs and Classes _somewhat_ interchangeably https://golang.org/doc/faq#Is_Go_an_object-oriented_language. I think this makes sense when we are talking about OO languages - a class is really just a blueprint.
> > 
> > 
> > My apologies, I did not mean to be pedantic here. I thought that correct term in Go was `type`/`struct` instead of `classes`, that's all ;)
> 
> hmmm? noooo don't apologize! i was just sharing something interesting about the class idea. just seems weird to me the interchanging use. wasn't trying to say you were being pedantic!!

ok ;)

pkg/signup/signup_client.go
```diff
@@ -0,0 +1,111 @@
...
+	//v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	USER_SIGNUP_RESOURCE_NAME = "usersignups"
+)
+
+func NewUserSignupClient(cfg *rest.Config) (*UserSignupV1Alpha1Client, error) {
+	scheme := runtime.NewScheme()
+	crtapi.SchemeBuilder.AddToScheme(scheme)
```
golangcibot: Error return value of `crtapi.SchemeBuilder.AddToScheme` is not checked (from `errcheck`)

test/fake/signup_client.go
```diff
@@ -0,0 +1,124 @@
...
+
+type FakeUserSignupClient struct {
+	Tracker   testing.ObjectTracker
+	Scheme    *runtime.Scheme
+	namespace string
+}
+
+func NewFakeUserSignupClient(namespace string, initObjs ...runtime.Object) *FakeUserSignupClient {
+	clientScheme := runtime.NewScheme()
+	crtapi.SchemeBuilder.AddToScheme(clientScheme)
```
golangcibot: Error return value of `crtapi.SchemeBuilder.AddToScheme` is not checked (from `errcheck`)

pkg/signup/signup_client.go
```diff
@@ -0,0 +1,99 @@
...
+	ClientSet *kubernetes.Clientset
+}
+
+func NewSignupClient() (*signupClient, error) {
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	clientset, err := kubernetes.NewForConfig(config)
```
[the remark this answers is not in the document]
tinakurian: @sbryzak i think it just moved a little: https://github.com/codeready-toolchain/registration-service/pull/23/files#diff-9033005564475e0bb99f8773d4829417R31

i think this should be the job the client - unless you think otherwise? i was thinking the client could encapsulate all this specific code for getting a client. but if you want, we can remove the client and do it differently.

sbryzak:
I can still improve coverage slightly, however there are some new bits that can only be tested in e2e tests.

alexeykazakov:
We definitely can't test everything in Unit tests but I believe we can improve the coverage for sure :)

pkg/signup/signup_service.go
```diff
@@ -0,0 +1,93 @@
...
+	created, err := c.Client.Create(userSignup)
+	if err != nil {
+		return nil, err
+	}
+
+	return created, nil
+}
+
+func (c *SignupServiceImpl) transformAndValidateUserName(username string) (string, error) {
+	replaced := strings.ReplaceAll(strings.ReplaceAll(username, "@", "-at-"), ".", "-")
```
alexeykazakov: We have to add something which will prevent creating multiple usersignups for the same user before we integrate this with the Signup controller. Otherwise it will break our signup/login flow right away.
If you prefer to address it in a separate PR then we need to do the following:

- Add two more subtasks to the signup story:
  1) Prevent creating multiple usersignups
  2) Integrate the client with the signup endpoint/controller

`i.` can't be done before `ii.`

pkg/signup/signup_client.go
```diff
@@ -0,0 +1,114 @@
...
+	crtapi "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/client-go/rest"
+	// See note below
+	//v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	USER_SIGNUP_RESOURCE_NAME = "usersignups"
```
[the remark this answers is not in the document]
sbryzak: Good change, implemented.

go.mod
```diff
@@ -31,11 +32,24 @@ require (
...
-	k8s.io/client-go v0.0.0-20190620085101-78d2af792bab
-	k8s.io/kube-openapi v0.0.0-20190816220812-743ec37842bf // indirect
+	k8s.io/apimachinery v0.0.0-20190607205628-5fbcd19f360b
+	k8s.io/client-go v11.0.0+incompatible
 	k8s.io/utils v0.0.0-20190809000727-6c36bc71fc4a // indirect
+	sigs.k8s.io/controller-runtime v0.1.12
+	sigs.k8s.io/yaml v1.1.0 // indirect
+)
+
+replace (
```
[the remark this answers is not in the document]
sbryzak: Done.

pkg/signup/signup_client.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &UserSignupV1Alpha1Client{
+		restClient: client,
+	}, nil
+}
+
+type UserSignupV1Alpha1Client struct {
```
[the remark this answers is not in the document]
sbryzak: I followed the pattern as recommended by [this article](https://medium.com/velotio-perspectives/extending-kubernetes-apis-with-custom-resource-definitions-crds-139c99ed3477).  I believe it benefits us with some future-proofing as the actual client implementation is tied to the CRD version.

I'm not sure we can easily introduce a generic client.. this one is specific to `UserSignup`, from the function parameters to its return types.  How do you think a generic client would look?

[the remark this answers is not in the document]
alexeykazakov: While the article you mention is about something else I think I got the idea behind the versioning in the client names. It could help when we need to support multiple API versions during operator API updates.

So, I'm OK to keep it as is.

About the generic client... Yeah.. it's probably too tricky/not worth doing. But at least maybe we could create some helper function to create a rest client for the scheme and GV and reuse it in the MUR and UserSignup clients: 

```
func RestClient(cfg *rest.Config, gv schema.GroupVersion, scheme *runtime.Scheme) (rest.Interface, error) {
	config := *cfg
	config.GroupVersion = &gv
	config.APIPath = "/apis"
	config.ContentType = runtime.ContentTypeJSON
	config.NegotiatedSerializer = serializer.DirectCodecFactory{CodecFactory: serializer.NewCodecFactory(scheme)}
	config.UserAgent = rest.DefaultKubernetesUserAgent()

	client, err := rest.RESTClientFor(&config)
	if err != nil {
		return nil, err
	}

	return client, nil
}
```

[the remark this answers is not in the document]
alexeykazakov: Or with some generic client:

```
import (
	crtapi "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"

	"k8s.io/apimachinery/pkg/runtime"
	"k8s.io/apimachinery/pkg/runtime/serializer"
	"k8s.io/client-go/rest"
)

func NewV1Alpha1Client(cfg *rest.Config, namespace string, registerObjects ...runtime.Object) (*V1Alpha1Client, error) {
	scheme := runtime.NewScheme()
	err := crtapi.SchemeBuilder.AddToScheme(scheme)
	if err != nil {
		return nil, err
	}
	crtapi.SchemeBuilder.Register(registerObjects...)

	config := *cfg
	config.GroupVersion = &crtapi.SchemeGroupVersion
	config.APIPath = "/apis"
	config.ContentType = runtime.ContentTypeJSON
	config.NegotiatedSerializer = serializer.DirectCodecFactory{CodecFactory: serializer.NewCodecFactory(scheme)}
	config.UserAgent = rest.DefaultKubernetesUserAgent()

	client, err := rest.RESTClientFor(&config)
	if err != nil {
		return nil, err
	}

	return &V1Alpha1Client{
		restClient: client,
		ns:         namespace,
	}, nil
}

type V1Alpha1Client struct {
	restClient rest.Interface
	ns         string
}
```

Then UserSignup client impl:

```
const (
	userSignupResourcePlural = "usersignups"
)

func NewUserSignupV1Alpha1Client(cfg *rest.Config, namespace string) (*UserSignupV1Alpha1Client, error) {
	client, err := NewV1Alpha1Client(cfg, namespace, &crtapi.UserSignup{}, &crtapi.UserSignupList{})
	if err != nil {
		return nil, err
	}

	return &UserSignupV1Alpha1Client{client}, nil
}

type UserSignupV1Alpha1Client struct {
	*V1Alpha1Client
}

func (c *UserSignupV1Alpha1Client) Get(name string) (*crtapi.UserSignup, error) {
	result := &crtapi.UserSignup{}
	err := c.restClient.Get().
		Namespace(c.ns).
		Resource(userSignupResourcePlural).
		Name(name).
		Do().
		Into(result)
	return result, err
}

func (c *UserSignupV1Alpha1Client) Create(obj *crtapi.UserSignup) (*crtapi.UserSignup, error) {
	result := &crtapi.UserSignup{}
	err := c.V1Alpha1Client.restClient.Post().
		Namespace(c.ns).
		Resource(userSignupResourcePlural).
		Body(obj).
		Do().
		Into(result)
	return result, err
}
```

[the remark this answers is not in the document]
alexeykazakov: And usage would be:

```
client, err := signup.NewUserSignupV1Alpha1Client(config, namespace)`
userSignUp, err := client.Get(name)
```
Or even `signup.NewV1Alpha1Client(config, namespace)` instead of `signup.NewUserSignupV1Alpha1Client(config, namespace)`.

WDYT?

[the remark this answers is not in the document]
sbryzak: Resolving this as we discussed offline

pkg/signup/signup_client.go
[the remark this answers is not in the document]
sbryzak: @tinakurian I discussed the client with Alexey today and we decided on a slight refactor to bring it more inline with the kubernetes API.

pkg/signup/signup_service.go
sbryzak: Two new issues created in JIRA:

- https://jira.coreos.com/browse/CRT-271

- https://jira.coreos.com/browse/CRT-272

alexeykazakov: Both CRT-271 and CRT-272 are about the same problem, aren't they?

So, it's all about:
> Prevent creating multiple usersignups

But what about a JIRA for:
> Integrate the client with the signup endpoint/controller

test/fake/signup_client.go
```diff
@@ -0,0 +1,128 @@
+package fake
+
+import (
+	"encoding/json"
```
tinakurian: can we organize these imports? 

i installed https://godoc.org/golang.org/x/tools/cmd/goimports - it works okay, does basic organization. also removes unused imports which is kind of nice...

pkg/signup/signup_service_test.go
```diff
@@ -0,0 +1,107 @@
+package signup_test
+
+import (
+	"context"
```
tinakurian: can we organize these imports?

i installed https://godoc.org/golang.org/x/tools/cmd/goimports - it works okay, does basic organization. also removes unused imports which is kind of nice...

pkg/signup/signup_service_test.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func (s *TestSignupServiceSuite) TestNewSignupService() {
+	// Simply test creation of the service, which should fail as the kubernetes env variables are not set
+	_, err := signup.NewSignupService(configuration.CreateEmptyRegistry())
+	require.Error(s.T(), err)
+}
+
+func (s *TestSignupServiceSuite) TestCreateUserSignup() {
+	svc, fake := newSignupServiceWithFakeClient()
```
tinakurian: we should probably try to use given/when/then to keep this consistent with the other repositories. we didn't do this in our other tests :-( guess i'll open a PR to add that tmrw if we decide we should add it here..

pkg/kubeclient/client.go
```diff
@@ -1,39 +1,67 @@
...
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/client-go/rest"
 )
 
-type KubeClient struct {
-	CoreClient kubernetes.Interface
-}
+func NewCRTV1Alpha1Client(cfg *rest.Config, namespace string) (*CRTV1Alpha1Client, error) {
```
tinakurian: could we add some documentation to our public apis?

pkg/kubeclient/mur.go
```diff
@@ -0,0 +1,28 @@
...
+
+type masterUserRecordClient struct {
+	crtClient
+}
+
+type MasterUserRecordInterface interface {
+	Get(name string) (*crtapi.MasterUserRecord, error)
+}
+
+func (c *masterUserRecordClient) Get(name string) (*crtapi.MasterUserRecord, error) {
```
tinakurian: could we add some documentation to our public apis?

pkg/kubeclient/signup.go
```diff
@@ -0,0 +1,40 @@
...
+type userSignupClient struct {
+	crtClient
+}
+
+type UserSignupInterface interface {
+	Get(name string) (*crtapi.UserSignup, error)
+	Create(obj *crtapi.UserSignup) (*crtapi.UserSignup, error)
+}
+
+func (c *userSignupClient) Get(name string) (*crtapi.UserSignup, error) {
```
tinakurian: could we add some documentation to our public apis?

pkg/kubeclient/signup.go
```diff
@@ -0,0 +1,40 @@
...
+	err := c.client.Get().
+		Namespace(c.ns).
+		Resource(userSignupResourcePlural).
+		Name(name).
+		Do().
+		Into(result)
+	return result, err
+}
+
+func (c *userSignupClient) Create(obj *crtapi.UserSignup) (*crtapi.UserSignup, error) {
```
tinakurian: could we add some documentation to our public apis?

pkg/signup/signup_service.go
```diff
@@ -0,0 +1,103 @@
...
+type SignupService interface {
+	CreateUserSignup(ctx context.Context, username, userID string) (*crtapi.UserSignup, error)
+}
+
+type SignupServiceImpl struct {
+	Namespace string
+	Client    *kubeclient.CRTV1Alpha1Client
+}
+
+func NewSignupService(cfg SignupServiceConfiguration) (SignupService, error) {
```
tinakurian: could we add some documentation to our public apis?

pkg/signup/signup_service.go
```diff
@@ -0,0 +1,103 @@
...
+		return nil, err
+	}
+
+	return &SignupServiceImpl{
+		Namespace: cfg.GetNamespace(),
+		Client:    client,
+	}, nil
+}
+
+func (c *SignupServiceImpl) CreateUserSignup(ctx context.Context, username, userID string) (*crtapi.UserSignup, error) {
```
tinakurian: could we add some documentation to our public apis?

pkg/kubeclient/client.go
```diff
@@ -1,39 +1,67 @@
...
-func getKubeConfig(parameters []string) rest.Config {
-	host := parameters[0] + "://" + parameters[1] + ":" + parameters[2]
-	bearerToken := parameters[3]
+type CRTV1Alpha1Client struct {
+	RestClient rest.Interface
+	NS         string
+}
 
-	return getOpenshiftAPIConfig(host, bearerToken)
+func (c *CRTV1Alpha1Client) UserSignups() UserSignupInterface {
```
tinakurian: could we add some documentation to our public apis?

pkg/kubeclient/client.go
```diff
@@ -1,39 +1,67 @@
...
+	}
 }
 
-func getOpenshiftAPIConfig(host string, bearerToken string) rest.Config {
-	return rest.Config{
-		Host:        host,
-		BearerToken: bearerToken,
-		TLSClientConfig: rest.TLSClientConfig{
-			Insecure: false,
+func (c *CRTV1Alpha1Client) MasterUserRecords() MasterUserRecordInterface {
```
tinakurian: could we add some documentation to our public apis?

pkg/signup/signup_service.go
sbryzak: Wouldn't that already be covered by https://jira.coreos.com/browse/CRT-246 ?

alexeykazakov: > Wouldn't that already be covered by https://jira.coreos.com/browse/CRT-246 ?

I don't know :) There is no any details in that JIRA. This PR doesn't include the `POST /signup` controller. This PR introduces the service which is supposed to be used by `POST /signup`. If this integration between `POST /signup` and the UserSignup service from this PR is supposed to be covered in CRT-246 then fine.

pkg/kubeclient/client.go
sbryzak: Done

pkg/signup/signup_service_test.go
sbryzak: I'm not a fan of this convention, it doesn't always fit in when a test is modelling complex workflows, so I'd prefer not to make it universally enforcable.

pkg/kubeclient/signup.go
sbryzak: Done

pkg/signup/signup_service.go
sbryzak: Done

pkg/signup/signup_service.go
sbryzak: Done

pkg/kubeclient/client.go
sbryzak: Done

pkg/kubeclient/client.go
sbryzak: Done

pkg/kubeclient/signup.go
sbryzak: Done

pkg/kubeclient/mur.go
sbryzak: Done

pkg/signup/signup_service_test.go
sbryzak: Neat tool, thanks for pointing it out :)  I've installed it into my GoLand environment so it now happens automatically when I save each file.

test/fake/signup_client.go
sbryzak: Done

pkg/signup/signup_service.go
sbryzak: The business logic for the signup controller's `PostHandler()` function will be handled in part by CRT-245.  There is also the matter of extracting the required claims from the user's access token, which I thought there was supposed to be a subtask for but I can't see it now. I'll follow up on this with Michael.

michaelkleinhenz: Extracting the claims we need is done. But we do need some glue between the existing reg service code and Shane's stuff. This also includes the async management and user "wait" page.

pkg/signup/signup_service.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+type SignupServiceImpl struct {
+	Namespace   string
+	UserSignups kubeclient.UserSignupInterface
+}
+
+// NewSignupService creates a service object for performing user signup-related activities
+func NewSignupService(cfg SignupServiceConfiguration) (SignupService, error) {
+	k8sConfig, err := rest.InClusterConfig()
```
alexeykazakov: If we need to introduce more services which will talk to kube api server we will need to move this configuration initialization to main.go to avoid initializing it multiple times.
But it's OK to keep it here for now. Just something to keep in mind :)

pkg/signup/signup_service.go
```diff
@@ -0,0 +1,105 @@
...
+
+type SignupServiceConfiguration interface {
+	GetNamespace() string
+}
+
+type SignupService interface {
+	CreateUserSignup(ctx context.Context, username, userID string) (*crtapi.UserSignup, error)
+}
+
+type SignupServiceImpl struct {
```
alexeykazakov: I know that previously (in fabric8) we followed that convention: 
Interfaces: `type SomeName interface {}`
Implementations: `type SomeNameImpl struct {}`

But looking at kubernetes I noticed that they follow a different convention:
Interfaces: `type SomeNameInterface interface {}`
Implementations: `type SomeName struct {}`

We followed that kube convention in our kube client too (`UserSignup` and `UserSignupInterface` for example).

I would stick with the kube one.

test/fake/signup_client.go
```diff
@@ -0,0 +1,112 @@
...
+	Scheme    *runtime.Scheme
+	namespace string
+}
+
+func NewFakeUserSignupClient(namespace string, initObjs ...runtime.Object) *FakeUserSignupClient {
+	clientScheme := runtime.NewScheme()
+	err := crtapi.SchemeBuilder.AddToScheme(clientScheme)
+	if err != nil {
+		log.Error(err, "Error adding to scheme")
+		os.Exit(1)
```
alexeykazakov: Don't exit. Let's panic (at most) if we can't do anything better.

test/fake/signup_client.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	crtapi.SchemeBuilder.Register(&crtapi.UserSignup{}, &crtapi.UserSignupList{})
+
+	tracker := testing.NewObjectTracker(clientScheme, scheme.Codecs.UniversalDecoder())
+	for _, obj := range initObjs {
+		err := tracker.Add(obj)
+		if err != nil {
+			log.Error(err, "failed to add object to fake user signup client", "object", obj)
+			os.Exit(1)
```
alexeykazakov: Same here.

pkg/signup/signup_service.go
sbryzak: @michaelkleinhenz let's include this glue code as part of CRT-246.

pkg/kubeclient/client.go
```diff
@@ -1,39 +1,70 @@
...
 
-	return kc
+	return &CRTV1Alpha1Client{
+		RestClient: client,
+		NS:         namespace,
+	}, nil
+}
+
+func getRegisterObject() []runtime.Object {
+	return []runtime.Object{&crtapi.UserSignup{}, &crtapi.UserSignupList{}}
```
alexeykazakov:
```suggestion
	return []runtime.Object{&crtapi.UserSignup{}, &crtapi.UserSignupList{}, &crtapi.MasterUserRecord{}, &crtapi.MasterUserRecordList{}}
```

test/fake/masteruserrecord_client.go
```diff
@@ -0,0 +1,111 @@
...
+		os.Exit(1)
+	}
+	crtapi.SchemeBuilder.Register(&crtapi.MasterUserRecord{}, &crtapi.MasterUserRecordList{})
+
+	tracker := testing.NewObjectTracker(clientScheme, scheme.Codecs.UniversalDecoder())
+	for _, obj := range initObjs {
+		err := tracker.Add(obj)
+		if err != nil {
+			log.Error(err, "failed to add object to fake user signup client", "object", obj)
+			panic("could not add object to tracker")
```
alexeykazakov: Let's maybe add the original error message too?
```suggestion
			panic("could not add object to tracker: " + err.Error())
```

test/fake/signup_client.go
```diff
@@ -0,0 +1,111 @@
...
+	Scheme    *runtime.Scheme
+	namespace string
+}
+
+func NewFakeUserSignupClient(namespace string, initObjs ...runtime.Object) *FakeUserSignupClient {
+	clientScheme := runtime.NewScheme()
+	err := crtapi.SchemeBuilder.AddToScheme(clientScheme)
+	if err != nil {
+		log.Error(err, "Error adding to scheme")
+		os.Exit(1)
```
alexeykazakov:
```suggestion
		panic("could not add to scheme: " + err.Error())
```

pkg/signup/signup_service_test.go
```diff
@@ -99,6 +100,32 @@ func (s *TestSignupServiceSuite) TestUserSignupInvalidName() {
...
+	fake.MockGet = func(string) (*v1alpha1.UserSignup, error) {
+		return nil, errors.New("an error occurred")
+	}
+
+	userID, err := uuid.NewV4()
+	require.NoError(s.T(), err)
+
+	_, err = svc.CreateUserSignup(context.Background(), "hank.smith@redhat.com", userID.String())
+	require.Error(s.T(), err)
```
alexeykazakov: Let’s check the error message too so we know that we got the expected error and not something else.

pkg/signup/signup_service_test.go
```diff
@@ -99,6 +100,32 @@ func (s *TestSignupServiceSuite) TestUserSignupInvalidName() {
...
+	svc, fake := newSignupServiceWithFakeClient()
+	fake.MockCreate = func(*v1alpha1.UserSignup) (*v1alpha1.UserSignup, error) {
+		return nil, errors.New("an error occurred")
+	}
+
+	userID, err := uuid.NewV4()
+	require.NoError(s.T(), err)
+
+	_, err = svc.CreateUserSignup(context.Background(), "jack.smith@redhat.com", userID.String())
+	require.Error(s.T(), err)
```
alexeykazakov: Here too.

pkg/signup/signup_service_test.go
```diff
@@ -0,0 +1,160 @@
...
+	userID, err := uuid.NewV4()
+	require.NoError(s.T(), err)
+
+	_, err = svc.CreateUserSignup(context.Background(), "john#gmail.com", userID.String())
+	require.Error(s.T(), err)
+}
+
+func (s *TestSignupServiceSuite) TestUserSignupNameExists() {
+	svc, fake := newSignupServiceWithFakeClient()
+	fake.Tracker.Add(&v1alpha1.UserSignup{
```
golangcibot: Error return value of `fake.Tracker.Add` is not checked (from `errcheck`)

test/fake/masteruserrecord_client.go
sbryzak: Done

test/fake/signup_client.go
sbryzak: Done

pkg/signup/signup_service_test.go
sbryzak: Compared the error returned from the service with the one returned by the mock get function.

test/fake/signup_client.go
sbryzak: Fixed

test/fake/signup_client.go
sbryzak: Fixed

pkg/signup/signup_service_test.go
sbryzak: Error now checked.

pkg/signup/signup_service.go
sbryzak: I know that I agreed it was better to be consistent with naming, but having thought about this some more I don't think it's necessarily better to conform to the Kubernetes naming standard for things that are actually outside of the Kubernetes framework.  So for the actual kube clients that we have created, I agree that they should conform to Kubernetes naming standards, however  `SignupService` is not a Kubernetes object, it's a service within our registration web app; while it talks to Kubernetes, it itself is designed as a standalone, "traditional" web application and as I mentioned previously I have more of an academic objection to the Kubernetes naming convention which I would rather not propagate further than we need to.

alexeykazakov: OKey :)

sbryzak:
I was able to add a basic test to cover the kubeclient constructor, however I after spending a couple of hours on it I didn't have any success with the fake `RESTClient` provided by `client-go`.  I may revisit this at the same time that we look at the e2e tests, if it still seems like these tests have value.

PR #24, opened 2019-09-18
Global Testmode and TestSuite

michaelkleinhenz:
This adds a global TestSuite that can be the base for all unit tests. The TestSuite sets some param like the `testmode`.


alexeykazakov:
I guess we can close it in favor of https://github.com/codeready-toolchain/registration-service/pull/29

PR #25, opened 2019-09-18
Fix naming

tinakurian:
This restructures healthCheck and signup to sit under the ```controller``` package. also renames files, structs, funcs and fixes comments to align with change. No functional changes here.

### DIFF
--- a/pkg/controller/health_check.go
+++ b/pkg/controller/health_check.go
@@ -1,4 +1,4 @@
-package health
+package controller
 
 import (
 	"encoding/json"
--- a/pkg/controller/health_check.go
+++ b/pkg/controller/health_check.go
@@ -9,22 +9,22 @@ import (
 	"github.com/gin-gonic/gin"
 )
 
-// HealthCheckService implements the service health endpoint.
-type HealthCheckService struct {
+// HealthCheck implements the health endpoint.
+type HealthCheck struct {
 	config *configuration.Registry
 	logger *log.Logger
 }
 
-// NewHealthCheckService returns a new HealthCheckService instance.
-func NewHealthCheckService(logger *log.Logger, config *configuration.Registry) *HealthCheckService {
-	return &HealthCheckService{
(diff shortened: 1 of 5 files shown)


pkg/controller/health_check.go
```diff
@@ -35,8 +35,8 @@ func (srv *HealthCheckService) getHealthInfo() map[string]interface{} {
 	return m
 }
 
-// HealthCheckHandler returns a default heath check result.
-func (srv *HealthCheckService) GetHealthCheckHandler(ctx *gin.Context) {
+// GetHealthCheckHandler returns a default heath check result.
+func (srv *HealthCheck) GetHealthCheckHandler(ctx *gin.Context) {
```
xcoulon: you could probably rename this func to just `Handler`, so when you register it in the server (see https://github.com/codeready-toolchain/registration-service/pull/25/files#diff-0fc56aea4678c02bd6ecd92471b1e16cR34), you'll have something like this:
```
handler := gin.HandlerFunc(healthCheck.Handler)
```

tinakurian: hmm i'm not sure that would work - if we have more than one handler func this name would need to be changed later

alexeykazakov: `healthCheck.Get` would work I think but what about `GetHandler()` instead?
So we would have something like:
```
healthCheckCtrl := controller.NewHealthCheck(srv.logger, srv.Config())
signupCtrl := controller.NewSignup(srv.logger, srv.Config())

v1.GET("/health", healthCheckCtrl.GetHandler)
v1.GET("/signup", signupCtrl.GetHandler)
v1.POST("/signup", signupCtrl.PostHandler)
```
But just 'Get()', `Post()` would work too I think...

tinakurian: fixed here: b45b29c341cf3a376237e278fc281b4a5d0dfdc2

PR #26, opened 2019-09-19
Test Token Generator

michaelkleinhenz:
This adds a tokengenerator for testing JWT and JWK.


test/tokengenerator_test.go
```diff
@@ -0,0 +1,147 @@
...
+	"testing"
+
+	"github.com/dgrijalva/jwt-go"
+	uuid "github.com/satori/go.uuid"
+	"github.com/stretchr/testify/require"
+	jose "gopkg.in/square/go-jose.v2"
+)
+
+func TestTokenGeneratorKeys(t *testing.T) {
+
```
tinakurian: for the tests, maybe we should is if/when/then comments since this is what we use in the other repositories?

michaelkleinhenz: Let's do this in #24 next sprint.

test/tokengenerator.go
```diff
@@ -0,0 +1,141 @@
...
+
+// NewTokenGenerator creates a new TokenGenerator.
+func NewTokenGenerator() *TokenGenerator {
+	tg := &TokenGenerator{}
+	tg.keyMap = make(map[string]*rsa.PrivateKey)
+	return tg
+}
+
+// CreateKey creates and stores a new key with the given kid.
+func (tg *TokenGenerator) CreateKey(kid string) (*rsa.PrivateKey, error) {
```
xcoulon: you could rename it as `AddPrivateKey`, so we know it not only generates, but also stores it

test/tokengenerator.go
```diff
@@ -0,0 +1,141 @@
...
+// NewIdentity returns a new, random identity
+func NewIdentity() *Identity {
+	return &Identity{
+		ID:       uuid.NewV4(),
+		Username: "testuser-" + uuid.NewV4().String(),
+	}
+}
+
+// TokenGenerator represents the test token and key manager.
+type TokenGenerator struct {
```
xcoulon: What about naming it `TokenManager` since it not only generates, but also stores keys?

test/tokengenerator.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// GetKeyService creates a http key service and return the URL
+func (tg *TokenGenerator) GetKeyService() string {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		keySet := &WebKeySet{}
+		for kid, key := range tg.keyMap {
+			thisKey, err := jwk.New(&key.PublicKey)
```
tinakurian: maybe newKey instead of thisKey?

test/tokengenerator.go
```diff
@@ -0,0 +1,141 @@
...
+			if err != nil {
+				panic(fmt.Sprintf("fatal error adding keys to key service: %s", err.Error()))
```
xcoulon: we should avoid `panic` even in tests. What about return an error instead?

test/tokengenerator_test.go
```diff
@@ -0,0 +1,147 @@
...
+	t.Run("create keys", func(t *testing.T) {
+		tokengenerator := NewTokenGenerator()
+		kid0 := uuid.NewV4().String()
+		key0, err := tokengenerator.CreateKey(kid0)
+		require.NoError(t, err)
+		kid1 := uuid.NewV4().String()
+		key1, err := tokengenerator.CreateKey(kid1)
+		require.NoError(t, err)
```
xcoulon: you should also verify that `key0` and `key1` are not `nil`

test/tokengenerator.go
```diff
@@ -0,0 +1,141 @@
...
+	tokenStr, err := token.SignedString(key)
+	if err != nil {
+		panic(errors.WithStack(err))
+	}
+	return tokenStr, nil
+}
+
+// GetKeyService creates a http key service and return the URL
+func (tg *TokenGenerator) GetKeyService() string {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
xcoulon: I'm not fully familiar with the `httptest` pkg, so my questions may be silly, but:
1. do you need to start the server, or is it implied?
2. how do you stop it once you don't need it anymore?

test/tokengenerator_test.go
```diff
@@ -0,0 +1,147 @@
...
+		// check key equality by comparing the modulus
+		require.NotEqual(t, key0.N, key1.N)
+	})
+
+	t.Run("get key", func(t *testing.T) {
+		tokengenerator := NewTokenGenerator()
+		kid0 := uuid.NewV4().String()
+		key0, err := tokengenerator.CreateKey(kid0)
+		require.NoError(t, err)
```
xcoulon: you should also verify that key0Retrieved and key1Retrieved are not nil

test/tokengenerator.go
tinakurian: @xcoulon looks like it starts on it's own, but we would need to call Close once we are done: https://golang.org/pkg/net/http/httptest/#NewServer

test/tokengenerator.go
alexeykazakov: Yeah, right. Panic is better than Exit() and a 500 error response is better than panic. It's not that critical as Exit() though :)

test/tokengenerator.go
tinakurian: fixed here: 56af8a9f5f97c17b87ede58a884271c3691ea3b1

test/tokengenerator.go
tinakurian: fixed here: 99f9f5a29e9fc882d73d514867a92a72773f5289

test/tokengenerator_test.go
tinakurian: fixed here: 2daa68a3e64dc9b8dbb591341c4a70bffdf0482c

test/tokengenerator_test.go
tinakurian: fixed here: 3af5601b2df6e621740160430d5a76981b66115c

test/tokengenerator.go
tinakurian: fixed here: c834aa3b62cefdbd437b17261a3d64c0f36870dc

test/tokengenerator.go
michaelkleinhenz: In theory, yes, but I had some issues with the defer and the server will be stopped anyway when the test is over. So I guess this is ok.

michaelkleinhenz: Ah, and starting is implied here. `NewServer()` starts it.

test/tokengenerator.go
tinakurian: fixed here: afa63c1cb951b6f2df5316d1b066ae6fa01bfc67

test/tokengenerator.go
alexeykazakov: Done in https://github.com/codeready-toolchain/registration-service/pull/26/commits/4911c512aeec45a41624cc11c28c4007e49af550

PR #27, opened 2019-09-19
Key and Token Manager

michaelkleinhenz:
This adds the Key and Token management with 3 components and associated tests:

- `testutils.TokenManager` is a test utility that can create JWK keys and JWT tokens. It also provides a fake http key service mocking a keycloak key service.
- `auth.KeyManager` fetches, parses, stores and provides the JWK keys pulled from a keycloak key service. This is used by other components for getting public keys for token validation. This is currently mainly used by `auth.TokenParser` to validate signed JWT tokens.
- `auth.TokenParser` parses JWK tokens, validates them and provides the claims to calling components. This is the actual component that will be used by the gin middleware in #28 to verify an incoming token.


pkg/auth/defaultmanager.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"errors"
+	"log"
+
+	"github.com/codeready-toolchain/registration-service/pkg/configuration"
+)
+
+var defaultKeyManager *KeyManager
+
+// DefaultKeyManagerWithConfig creates the default manager if it has not created yet.
```
xcoulon:
```suggestion
// DefaultKeyManagerWithConfig creates the default manager if it was not created yet.
```

pkg/auth/defaultmanager.go
```diff
@@ -0,0 +1,30 @@
...
+// This function must be called in main to make sure the default manager is created during service startup.
+// It will try to create the default manager only once even if called multiple times.
```
xcoulon: for that, you could use something like `sync.Once.Do()`, as in https://github.com/fabric8-services/fabric8-common/blob/2cc2344d55776abe18178fc1c2efa8abc9996565/auth/token.go#L58-L63
(See also https://golang.org/pkg/sync/#Once.Do)

pkg/auth/keymanager_test.go
```diff
@@ -0,0 +1,225 @@
...
+	logger := log.New(os.Stderr, "", 0)
+	configRegistry := configuration.CreateEmptyRegistry()
+
+	// Set the config for testing mode, the handler may use this.
+	configRegistry.GetViperInstance().Set("testingmode", true)
+	assert.True(t, configRegistry.IsTestingMode(), "testing mode not set correctly to true")
+
+	t.Run("missing logger", func(t *testing.T) {
+		_, err := auth.NewKeyManager(nil, configRegistry)
+		require.Error(t, err)
```
xcoulon: you could also verify the error message

pkg/auth/keymanager_test.go
```diff
@@ -0,0 +1,225 @@
...
+	})
+
+	t.Run("missing config", func(t *testing.T) {
+		_, err := auth.NewKeyManager(logger, nil)
+		require.Error(t, err)
```
xcoulon: you could also verify the error message

pkg/auth/keymanager_test.go
```diff
@@ -0,0 +1,225 @@
...
+	})
+
+	t.Run("missing logger and config", func(t *testing.T) {
+		_, err := auth.NewKeyManager(nil, nil)
+		require.Error(t, err)
```
xcoulon: you could also verify the error message

pkg/auth/tokenparser.go
```diff
@@ -0,0 +1,92 @@
...
+	EmailVerified bool           `json:"email_verified"`
+	Company       string         `json:"company"`
+	SessionState  string         `json:"session_state"`
+	Approved      bool           `json:"approved"`
+	Permissions   *[]Permissions `json:"permissions"`
+	jwt.StandardClaims
+}
+
+// Permissions represents a "permissions" claim in the AuthorizationPayload
+type Permissions struct {
```
xcoulon: Do we need permissions here? This looks like some legacy code from `fabric8-auth`

pkg/auth/tokenparser_test.go
```diff
@@ -0,0 +1,111 @@
...
+	jwt1, err := tokengenerator.GenerateSignedToken(*identity1, kid1, testutils.WithEmailClaim(email1))
+	require.NoError(t, err)
+
+	// create invalid test token (wrong set of claims, no email), signed with key1
+	username_invalid := uuid.NewV4().String()
+	identity_invalid := &testutils.Identity{
+		ID:       uuid.NewV4(),
+		Username: username_invalid,
+	}
+	jwt_invalid, err := tokengenerator.GenerateSignedToken(*identity_invalid, kid1)
```
xcoulon: you could move this part to the only subtest that uses it, to simplify the understanding of all this setup

pkg/auth/tokenparser_test.go
```diff
@@ -0,0 +1,111 @@
...
+	// create test keys
+	tokengenerator := testutils.NewTokenManager()
+	kid0 := uuid.NewV4().String()
+	_, err := tokengenerator.AddPrivateKey(kid0)
+	require.NoError(t, err)
+	kid1 := uuid.NewV4().String()
+	_, err = tokengenerator.AddPrivateKey(kid1)
+	require.NoError(t, err)
+
+	// create two test tokens, both valid
```
xcoulon: you could move this part to the only subtest that uses it, to simplify the understanding of all this setup

pkg/auth/defaultmanager.go
michaelkleinhenz: This is all a bit bad. I removed the `sync.Do`, because the impl from the legacy fabric8 code could silently fail there when called more than once. Found a better solution. Fixed 8954a70.

pkg/auth/defaultmanager.go
michaelkleinhenz: See above.

pkg/auth/keymanager_test.go
michaelkleinhenz: Fixed 7a0296f.

pkg/auth/keymanager_test.go
michaelkleinhenz: Fixed 7a0296f.

pkg/auth/keymanager_test.go
michaelkleinhenz: Fixed 7a0296f.

pkg/auth/tokenparser.go
michaelkleinhenz: Ah, ok, thought that was something RHD specific.. Fixed 1d3f3aa.

pkg/auth/tokenparser_test.go
michaelkleinhenz: Done d7a998c.

pkg/auth/tokenparser_test.go
michaelkleinhenz: Done d7a998c.

pkg/auth/defaultmanager.go
xcoulon: could you elaborate on the failure you mentioned, please? AFAIK, `sync.Once` is what we need here

xcoulon:
also, coverage drops by 4%. Is there any block of code that is not tested in this PR (besides the usual error checks)?

pkg/auth/defaultmanager.go
michaelkleinhenz: `sync.Once` is at least not enough as a repeated call to `DefaultKeyManagerWithConfig()` with an *updated* config (or logger) would just silently ignore the new config and return the existing instance. I'd say that is bad design as the caller don't get any indication that his call (with a different config) is returning an instance with the "old" config. If we don't want to have this called more than once, then let's drop an error when it is called more than once (or make it recognize updated arguments, which is not an option here).

Just returning something that does ignore the arguments will probably lead to hard to track down issues where a caller expects the config to be recognized if the call does return without an error.

And if we just check the instance existence, we don't need the `sync.Once` here.

xcoulon: ok, I see what you mean with the config update use-case, now. But should we use a lock to avoid concurrent init, at least?

michaelkleinhenz: Ok, Tina and I found a better alternative that has the issue solved while being thread safe. See 59c23be.

pkg/auth/defaultmanager.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+var mu sync.Mutex
+
+var defaultKeyManager *KeyManager
+
+// InitializeDefaultKeyManager creates the default manager if it has not created yet.
+// This function must be called in main to make sure the default manager is created during service startup.
+// It will try to create the default manager only once even if called multiple times.
+func InitializeDefaultKeyManager(logger *log.Logger, config *configuration.Registry) (*KeyManager, error) {
```
alexeykazakov: I think it's a good practice to create specific configuration interfaces and pass them instead of using configuration struct pointers.
Then it's clear what you actually need in the function. You don't need the entire configuration. You need only one or a few things from it.
See https://github.com/fabric8-services/fabric8-common/blob/master/auth/token.go#L58 and https://github.com/fabric8-services/fabric8-common/blob/c08218d5adf274cd1950db3636e825dc180921cc/auth/token.go#L70:6 as an example.

It also gives more flexibility in tests.

pkg/auth/tokenparser.go
```diff
@@ -0,0 +1,83 @@
...
+type TokenClaims struct {
+	Name          string         `json:"name"`
+	Username      string         `json:"preferred_username"`
+	GivenName     string         `json:"given_name"`
+	FamilyName    string         `json:"family_name"`
+	Email         string         `json:"email"`
+	EmailVerified bool           `json:"email_verified"`
+	Company       string         `json:"company"`
+	SessionState  string         `json:"session_state"`
+	Approved      bool           `json:"approved"`
```
alexeykazakov: It's OSIO legacy staff. Let's remove `Approved` claim.

pkg/auth/tokenparser.go
```diff
@@ -0,0 +1,83 @@
...
+// TokenClaims represents access token claims
+type TokenClaims struct {
+	Name          string         `json:"name"`
+	Username      string         `json:"preferred_username"`
+	GivenName     string         `json:"given_name"`
+	FamilyName    string         `json:"family_name"`
+	Email         string         `json:"email"`
+	EmailVerified bool           `json:"email_verified"`
+	Company       string         `json:"company"`
+	SessionState  string         `json:"session_state"`
```
alexeykazakov: We don't need SessionState. Also it's more implementation detail. It can be changed in RHD anytime.

pkg/auth/tokenparser.go
```diff
@@ -0,0 +1,83 @@
...
+
+// TokenParser represents a parser for JWT tokens.
+type TokenParser struct {
+	config     *configuration.Registry
+	logger     *log.Logger
+	keyManager *KeyManager
+}
+
+// NewTokenParser creates a new TokenParser.
+func NewTokenParser(logger *log.Logger, config *configuration.Registry, keyManager *KeyManager) (*TokenParser, error) {
```
alexeykazakov: Same here with configuration interface.

pkg/auth/tokenparser.go
```diff
@@ -0,0 +1,83 @@
...
+	if logger == nil {
```
alexeykazakov: IMO having the default TokenParser singleton would make more sense than having the default KeyManager singleton.
You don't need to access KeyManager directly. You need to access the token parser which is using the keys from the manager behind the sense. KeyManager is used only by the token manager. It's rather an implementation detail of the parser. KeyManager alone is useless.

pkg/auth/tokenparser.go
```diff
@@ -0,0 +1,83 @@
...
+		return nil, err
+	}
+	if claims, ok := token.Claims.(*TokenClaims); ok && token.Valid {
+		// we need username and email, so check if those are contained in the claims
+		if claims.Username == "" {
+			return nil, errors.New("token does not comply to expected claims: username missing")
+		}
+		if claims.Email == "" {
+			return nil, errors.New("token does not comply to expected claims: email missing")
+		}
```
alexeykazakov: We also need `sub` claim. It's the user ID.

pkg/auth/tokenparser_test.go
```diff
@@ -0,0 +1,111 @@
...
+	require.NoError(t, err)
+	kid1 := uuid.NewV4().String()
+	_, err = tokengenerator.AddPrivateKey(kid1)
+	require.NoError(t, err)
+
+	// startup public key service
+	keysEndpointURL := tokengenerator.NewKeyServer().URL
+
+	// set the config for testing mode, the handler may use this.
+	configRegistry.GetViperInstance().Set("testingmode", true)
```
alexeykazakov: Use UnitTestSuite.

pkg/auth/tokenparser_test.go
```diff
@@ -0,0 +1,111 @@
...
+	keyManager, err := auth.NewKeyManager(logger, configRegistry)
+	require.NoError(t, err)
+
+	// create TokenParser instance.
+	tokenParser, err := auth.NewTokenParser(logger, configRegistry, keyManager)
+	require.NoError(t, err)
+
+	t.Run("invalid arguments to new", func(t *testing.T) {
+		_, err := auth.NewTokenParser(nil, configRegistry, keyManager)
+		require.Error(t, err)
```
alexeykazakov: Check the error message.

pkg/auth/tokenparser_test.go
```diff
@@ -0,0 +1,111 @@
...
+		_, err = auth.NewTokenParser(logger, nil, keyManager)
+		require.Error(t, err)
```
alexeykazakov: Check the error message.

pkg/auth/tokenparser_test.go
```diff
@@ -0,0 +1,111 @@
...
+		_, err = auth.NewTokenParser(logger, configRegistry, nil)
+		require.Error(t, err)
```
alexeykazakov: Check the error message.

@xcoulon I wonder if we can add a linter for that? :)

pkg/auth/tokenparser_test.go
```diff
@@ -0,0 +1,111 @@
...
+			t.Run(tt.name, func(t *testing.T) {
+				claims, err := tokenParser.FromString(tt.jwt)
+				require.NoError(t, err)
+				require.Equal(t, tt.username, claims.Username)
+				require.Equal(t, tt.email, claims.Email)
+			})
+		}
+	})
+
+	t.Run("parse invalid token", func(t *testing.T) {
```
alexeykazakov: We should cover all the cases of invalid tokens:

1. No KID header in the token.
2. Token signed by unknown key.
3. Token signed by known key but the signature is invalid.
4. Signature is good but token expired.
5. Missing claims (individual tests for each missing claim)

pkg/auth/defaultmanager.go
```diff
@@ -0,0 +1,38 @@
...
+		if err != nil {
+			return nil, err
+		}
+		return defaultKeyManager, nil
+	}
+	return nil, errors.New("default KeyManager can be created only once")
+}
+
+// DefaultKeyManager returns the existing KeyManager instance.
+func DefaultKeyManager() (*KeyManager, error) {
```
alexeykazakov: As mentioned in Slack I still think that's Do.Once is better but if you still want to keep it as it is then let's make sure that we have 100% test coverage for both `InitializeDefaultKeyManager()` and `DefaultKeyManager()` functions.

pkg/auth/keymanager_test.go
```diff
@@ -0,0 +1,228 @@
...
+		require.NoError(t, err)
+
+		// Set the config for testing mode, the handler may use this.
+		configRegistry.GetViperInstance().Set("auth_client.public_keys_url", ts.URL)
+		assert.Equal(t, configRegistry.GetAuthClientPublicKeysURL(), ts.URL, "key url not set correctly for testing")
+
+		// Create KeyManager instance.
+		_, err = auth.NewKeyManager(logger, configRegistry)
+		// this needs to fail with an error
+		require.Error(t, err)
```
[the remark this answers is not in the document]
tinakurian: fixed here: ca066720bf18117dcf3bdc5c8a6d7d884df5b233

pkg/auth/keymanager_test.go
```diff
@@ -0,0 +1,228 @@
...
+	t.Run("parse keys, server not reachable", func(t *testing.T) {
+		// Set the config for testing mode, the handler may use this.
+		anURL := "http://www.google.com/"
+		configRegistry.GetViperInstance().Set("auth_client.public_keys_url", anURL)
+		assert.Equal(t, configRegistry.GetAuthClientPublicKeysURL(), anURL, "key url not set correctly for testing")
+
+		// Create KeyManager instance.
+		_, err := auth.NewKeyManager(logger, configRegistry)
+		// this needs to fail with an error
+		require.Error(t, err)
```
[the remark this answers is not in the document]
tinakurian: fixed here: ca066720bf18117dcf3bdc5c8a6d7d884df5b233

pkg/auth/tokenparser_test.go
tinakurian: fixed here: ca066720bf18117dcf3bdc5c8a6d7d884df5b233

pkg/auth/tokenparser_test.go
tinakurian: fixed here: ca066720bf18117dcf3bdc5c8a6d7d884df5b233

pkg/auth/tokenparser_test.go
tinakurian: fixed here: ca066720bf18117dcf3bdc5c8a6d7d884df5b233

pkg/auth/tokenparser.go
tinakurian: fixed here: 8fff4ce36eadcb70979eebcc6b6d5609c94ecbad

pkg/auth/keymanager_test.go
```diff
@@ -0,0 +1,228 @@
...
+	t.Run("parse keys, invalid url", func(t *testing.T) {
+		// Set the config for testing mode, the handler may use this.
+		notAnURL := "not an url"
+		configRegistry.GetViperInstance().Set("auth_client.public_keys_url", notAnURL)
+		assert.Equal(t, configRegistry.GetAuthClientPublicKeysURL(), notAnURL, "key url not set correctly for testing")
+
+		// Create KeyManager instance.
+		_, err := auth.NewKeyManager(logger, configRegistry)
+		// this needs to fail with an error
+		require.Error(t, err)
```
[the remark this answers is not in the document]
tinakurian: fixed here: ca066720bf18117dcf3bdc5c8a6d7d884df5b233

pkg/auth/tokenparser.go
tinakurian: fixed here: 24c8518f9df59bbb108003c9d1ad6da9adb0e4cd

pkg/auth/tokenparser_test.go
tinakurian: fixed here: 47c9cd9bc424126fc019026fda6b1a4afeeca586

pkg/auth/keymanager_test.go
```diff
@@ -0,0 +1,228 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestKeyManager(t *testing.T) {
+	// Create logger and registry.
+	logger := log.New(os.Stderr, "", 0)
+	configRegistry := configuration.CreateEmptyRegistry()
+
+	// Set the config for testing mode, the handler may use this.
+	configRegistry.GetViperInstance().Set("testingmode", true)
```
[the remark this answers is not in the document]
tinakurian: fixed here: 47c9cd9bc424126fc019026fda6b1a4afeeca586

pkg/auth/tokenparser_test.go
michaelkleinhenz: Fixed 96b6979. For (3): the signature implies a key, so I added a test where the signature is garbage (in addition to a test that has a valid signature from an unknown/non-kid-matching key).

pkg/auth/defaultmanager.go
michaelkleinhenz: Fixed c364d0a.

test/tokenmanager_test.go
```diff
@@ -38,6 +38,21 @@ func (s *TestTokenManagerSuite) TestTokenManagerKeys() {
...
+		tokenManager := NewTokenManager()
+		kid0 := uuid.NewV4().String()
+		key0, err := tokenManager.AddPrivateKey(kid0)
+		require.NoError(s.T(), err)
+		require.NotNil(s.T(), key0)
+		key0Retrieved, err := tokenManager.Key(kid0)
+		require.NoError(s.T(), err)
+		require.NotNil(s.T(), key0Retrieved)
+		tokenManager.RemovePrivateKey(kid0)
+		key0Retrieved, err = tokenManager.Key(kid0)
```
golangcibot: ineffectual assignment to `key0Retrieved` (from `ineffassign`)

pkg/auth/defaultmanager.go
xcoulon: Agreed, it's particularly useful for testing, indeed

pkg/auth/defaultmanager.go
```diff
@@ -0,0 +1,38 @@
...
+	if defaultKeyManager == nil {
+		var err error
+		defaultKeyManager, err = NewKeyManager(logger, config)
+		if err != nil {
+			return nil, err
+		}
+		return defaultKeyManager, nil
+	}
+	return nil, errors.New("default KeyManager can be created only once")
+}
```
xcoulon: this looks good (as discussed)

pkg/auth/defaultmanager_test.go
```diff
@@ -0,0 +1,51 @@
...
+		require.Equal(s.T(), "no default KeyManager created, call `InitializeDefaultKeyManager()` first", err.Error())
+	})
+
+	s.Run("first creation", func() {
+		_, err := auth.InitializeDefaultKeyManager(logger, s.Config)
+		require.NoError(s.T(), err)
+	})
+
+	s.Run("second redundant creation", func() {
+		_, err := auth.InitializeDefaultKeyManager(logger, s.Config)
```
xcoulon: you should also try to call this `InitializeDefaultKeyManager` with multiple go routines at the same time. See here: https://github.com/fabric8-services/fabric8-auth/blob/b38ee9bf66841ed18af4b01af5dd0ae7f0197b47/worker/worker_blackbox_test.go#L39-L70 In this code, we "prepare" a handful of go routines, but keep them "on hold" and suddenly release them all at once so they call all call the same function in parallel. All but one should fail in your case.

pkg/auth/defaultmanager.go
michaelkleinhenz: Ah, yes, GitHub had this hidden. Fixed in c748887.

pkg/auth/tokenparser.go
michaelkleinhenz: Fixed in f6e1393. Removed the config, TokeParser don't need a config.

pkg/auth/tokenparser.go
michaelkleinhenz: But TokenParser can freely instantiated, it does not hold state. This is the difference between it and KeyManager. I would put the singleton as deep as possible to lock as little as possible as a singleton..

pkg/auth/tokenparser.go
michaelkleinhenz: Fixed 5b7e2b7.

pkg/auth/defaultmanager_test.go
michaelkleinhenz: Ok, done 5876bb9.

pkg/auth/defaultmanager_test.go
```diff
@@ -1,22 +1,26 @@
...
 import (
+	"fmt"
 	"log"
 	"os"
+	"sync"
 	"testing"
+	"time"
 
-	"github.com/codeready-toolchain/registration-service/pkg/auth"
+//	"github.com/codeready-toolchain/registration-service/pkg/auth"
```
xcoulon: please remove this line

pkg/auth/defaultmanager_test.go
```diff
@@ -28,24 +32,70 @@ func (s *TestDefaultManagerSuite) TestKeyManagerDefaultKeyManager() {
...
+					KmErr:   err,
+				}
+				holder[i] = thisHolder
+			}(i)
+		}
+		latch.Done()
+		// wait a few cycles before checking the results
+		time.Sleep(time.Millisecond * 1000)
+		// check if only one entry has a KeyManager and the two others have errs
+		fmt.Println(holder)
```
xcoulon: please remove this line

pkg/auth/defaultmanager_test.go
```diff
@@ -28,24 +32,70 @@ func (s *TestDefaultManagerSuite) TestKeyManagerDefaultKeyManager() {
...
+				km, err := InitializeDefaultKeyManager(logger, s.Config)
+				thisHolder := &kmErrHolder{
+					KeyMngr: km,
+					KmErr:   err,
+				}
+				holder[i] = thisHolder
+			}(i)
+		}
+		latch.Done()
+		// wait a few cycles before checking the results
```
xcoulon: this comment does not apply here

pkg/auth/defaultmanager_test.go
```diff
@@ -28,24 +32,70 @@ func (s *TestDefaultManagerSuite) TestKeyManagerDefaultKeyManager() {
...
+		time.Sleep(time.Millisecond * 1000)
```
xcoulon: you can probably reduce this duration

pkg/auth/defaultmanager_test.go
```diff
@@ -28,24 +32,70 @@ func (s *TestDefaultManagerSuite) TestKeyManagerDefaultKeyManager() {
...
+		// check if only one entry has a KeyManager and the two others have errs
+		fmt.Println(holder)
+		fails := 0
+		success := 0
+		for i := 0; i < 3; i++ {
+			thisEntry := holder[i]
+			fmt.Println(thisEntry.KmErr)
```
xcoulon: please remove this line, or use `t.Logf(...)` so it'll be printed if the test fails

pkg/auth/defaultmanager_test.go
michaelkleinhenz: fixed 5876bb9.

pkg/auth/defaultmanager_test.go
michaelkleinhenz: fixed 5876bb9.

pkg/auth/defaultmanager_test.go
michaelkleinhenz: fixed 5876bb9.

pkg/auth/defaultmanager_test.go
michaelkleinhenz: fixed 5876bb9.

pkg/auth/defaultmanager_test.go
michaelkleinhenz: Fixed 8ede144.

pkg/auth/tokenparser.go
michaelkleinhenz: ...but I added that in 9fe2cac to get this done. We have <48 hours to complete the story.

alexeykazakov: The token parser holds the state indirectly (it holds KeyManager). It's all about usage IMO.
KeyManager is supposed to be used by the token parser too. So, it's rather a token parser implementation detail. And if we needed to use the parser in many places then having a default instance you can access everywhere (`auth.DefatultTokenParser()`) would help. But since we don't plan to use it only in the auth MW then I think it's not that important.

> but I added that in 9fe2cac to get this done

I actually didn't mean to have two defaults: KeyManager and TokenParser. I was talking about just one default TokenParser and no any default key manager.
But as I said it's not that important. I'm fine with keeping it as it is now or with reverting 9fe2cac.

pkg/auth/defaultmanager_test.go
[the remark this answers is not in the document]
michaelkleinhenz: Already fixed. Xavier also noticed that leftover..

pkg/auth/defaultmanager.go
```diff
@@ -36,3 +34,8 @@ func DefaultKeyManager() (*KeyManager, error) {
 	}
 	return defaultKeyManager, nil
 }
+
+// not exported, only used for test, removes singleton.
+func resetDefaultKeyManager() {
```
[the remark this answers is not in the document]
michaelkleinhenz: good point..fixed in c386bfa.

pkg/auth/tokenparser_test.go
```diff
@@ -0,0 +1,277 @@
...
+		// manipulate expiry
+		tDiff := -60 * time.Second
+		jwt0.Claims.(jwt.MapClaims)["exp"] = time.Now().UTC().Add(tDiff).Unix()
+		// serialize
+		jwt0string, err := tokengenerator.SignToken(jwt0, kid0)
+		require.NoError(t, err)
+		// validate token
+		_, err = tokenParser.FromString(jwt0string)
+		require.Error(t, err)
+		require.EqualError(t, err, "token is expired by 1m0s")
```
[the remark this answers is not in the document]
michaelkleinhenz: Yeah, indeed. I'll only check the prefix now. 10f07ed.

PR #28, opened 2019-09-19
JWT Auth Middleware

michaelkleinhenz:
This adds the auth middleware.


michaelkleinhenz:
Before someone asks: middleware is in a seperate package to avoid issues with the singletons (to be specific, the reset of the singletons when testing as seperate package tests are running in seperate contexts).

pkg/middleware/middleware_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	// create server
+	srv, err := server.New("")
+	require.NoError(t, err)	
+
+	// set the key service url in the config
+	os.Setenv(configuration.EnvPrefix+"_"+"AUTH_CLIENT_PUBLIC_KEYS_URL", keysEndpointURL)
+	assert.Equal(t, keysEndpointURL, srv.Config().GetAuthClientPublicKeysURL(), "key url not set correctly")
+	os.Setenv(configuration.EnvPrefix+"_"+"TESTINGMODE", "true")
+	assert.True(t, srv.Config().IsTestingMode(), "testing mode not set correctly")
```
alexeykazakov: I see that this auth setup is being duplicated in different tests. We should really just have one or two lines of code to generate a token with setup the test token generator with the default key manager in place.
Something like `tokengenerator.GenerateSignedToken()` which should use the default keys under the hood which are already available in the default token parser in the test mode.
Otherwise we will have to pass the whole auth setup over all our tests which requires authentication.

pkg/middleware/middleware_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	// Check that Engine() returns the router object.
+	require.NotNil(t, srv.Engine())
+
+	// do some requests
+	var authtests = []struct {
+		name    		string
+		urlPath 		string
+		method  		string
+		tokenHeader	string
```
alexeykazakov: Looks like a formatting issue. Do you have `go fmt` setup in your IDE?

pkg/server/routes.go
```diff
@@ -44,23 +48,42 @@ func (h StaticHandler) ServeHTTP(ctx *gin.Context) {
...
+		publicV1.GET("/authconfig", authConfigCtrl.GetHandler)
+
+		// private routes
+		privateV1 := srv.router.Group("/api/v1")
+		privateV1.Use(authMiddleware.HandlerFunc())
+		privateV1.POST("/signup", signupCtrl.PostHandler)
+
+		// if we are in testing mode, we also add a private health route for testing
+		if srv.Config().IsTestingMode() {
+			privateV1.GET("/health_private", healthCheckCtrl.GetHandler)
```
alexeykazakov: Can we call it `/auth_test` or something like that?

pkg/middleware/middleware.go
```diff
@@ -0,0 +1,124 @@
...
+			return
+		}
+		// validate time claims
+		if token.Valid() != nil {
+			m.respondWithError(c, http.StatusForbidden, "token has invalid time claims")
+			c.Abort()
+			return
+		}
+		// check if we have the needed claims for username and email
+		if token.Username == "" {
```
alexeykazakov: We do need to check Username, email (and also sub which is missing here) but it's already checked by the token parser - https://github.com/codeready-toolchain/registration-service/blob/4edbec2c9dbed84a985618392f2f7c5a45a36fa4/pkg/auth/tokenparser.go#L63-L75
So, all these claim checks are redundant here, aren't they?

pkg/middleware/middleware.go
```diff
@@ -0,0 +1,124 @@
...
+		}
+		if token.Email == "" {
+			m.respondWithError(c, http.StatusForbidden, "token does not have email set")
+			c.Abort()
+			return
+		}
+		// all checks done, add username and email to the context
+		c.Set(UsernameKey, token.Username)
+		c.Set(EmailKey, token.Email)
+		// for convenience, add the claims to the context
```
alexeykazakov: Probably also worth adding `sub` claim (userID) to the context explicitly. We will need to create a `UserSignup` resource.

pkg/server/routes.go
```diff
@@ -44,23 +48,42 @@ func (h StaticHandler) ServeHTTP(ctx *gin.Context) {
...
+// SetupRoutes registers handlers for various URL paths.
 func (srv *RegistrationServer) SetupRoutes() error {
 	var err error
 	srv.routesSetup.Do(func() {
+		// initialize default managers
+		keyManager, err := auth.InitializeDefaultKeyManager(srv.Logger(), srv.Config())
+		if err != nil {
+			err = errs.Wrapf(err, "failed to init default key manager: %s", err.Error())
+		}
+		_, err = auth.InitializeDefaultTokenParser(srv.logger, keyManager)
```
alexeykazakov: The default token parser could initialize the default key manager on its own, right? :)

pkg/server/routes.go
```diff
@@ -44,23 +48,42 @@ func (h StaticHandler) ServeHTTP(ctx *gin.Context) {
...
-// SetupRoutes registers handlers for various URL paths. You can call this
-// function more than once but only the first call will have an effect.
+// SetupRoutes registers handlers for various URL paths.
 func (srv *RegistrationServer) SetupRoutes() error {
 	var err error
 	srv.routesSetup.Do(func() {
+		// initialize default managers
+		keyManager, err := auth.InitializeDefaultKeyManager(srv.Logger(), srv.Config())
+		if err != nil {
+			err = errs.Wrapf(err, "failed to init default key manager: %s", err.Error())
```
alexeykazakov: Every error here is fatal. We should immediately abort and fail/exit. So, if something is wrong during setting up the routs the service should fail to initialize and the k8s deployment will fail too.
Just adding `return` here should be enough I think. Otherwise `err` can be overridden later (in the line 60 for example) and become `nil` again.

pkg/server/routes.go
```diff
@@ -44,23 +48,42 @@ func (h StaticHandler) ServeHTTP(ctx *gin.Context) {
...
+		}
+		_, err = auth.InitializeDefaultTokenParser(srv.logger, keyManager)
+		if err != nil {
+			err = errs.Wrapf(err, "failed to init default token parser: %s", err.Error())
```
alexeykazakov: Same here. It's fatal.

pkg/server/routes.go
```diff
@@ -44,23 +48,42 @@ func (h StaticHandler) ServeHTTP(ctx *gin.Context) {
...
 		signupCtrl := controller.NewSignup(srv.logger, srv.Config())
 
-		v1 := srv.router.Group("/api/v1")
-		{
-			v1.GET("/health", healthCheckCtrl.GetHandler)
-			v1.GET("/authconfig", authConfigCtrl.GetHandler)
-			v1.POST("/signup", signupCtrl.PostHandler)
+		// create the auth middleware
+		var authMiddleware *middleware.JWTMiddleware
+		authMiddleware, err = middleware.NewAuthMiddleware(srv.logger, srv.config)
```
alexeykazakov: Check err and return if needed.

pkg/server/routes.go
```diff
@@ -44,23 +48,42 @@ func (h StaticHandler) ServeHTTP(ctx *gin.Context) {
...
+
+		// public routes
+		publicV1 := srv.router.Group("/api/v1")
+		publicV1.GET("/health", healthCheckCtrl.GetHandler)
+		publicV1.GET("/authconfig", authConfigCtrl.GetHandler)
+
+		// private routes
```
alexeykazakov: Public vs. private is confusing here. It's not about API privacy. We should call them `secured` and `unsecured` routes/endpoints.

pkg/controller/health_check.go
```diff
@@ -27,7 +27,7 @@ func NewHealthCheck(logger *log.Logger, config *configuration.Registry) *HealthC
 func (srv *HealthCheck) getHealthInfo() map[string]interface{} {
 	m := make(map[string]interface{})
 	// TODO: this need to get actual health info.
-	m["alive"] = !srv.config.IsTestingMode()
+	m["alive"] = true
```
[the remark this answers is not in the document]
michaelkleinhenz: It's to make testing easier as there are tests that require testmode to be off, so I made this static and predictable no matter what mode is set. And there is still an additional result value to indicate we're running in test mode. Also, I'd expect that this is not the "final" health check but this is extended when we move to an actual runtime with some actual system health checks.

pkg/controller/signup.go
```diff
@@ -27,6 +28,8 @@ func NewSignup(logger *log.Logger, config *configuration.Registry) *Signup {
 func (srv *Signup) PostHandler(ctx *gin.Context) {
 	ctx.Writer.Header().Set("Content-Type", "application/json")
 
+	// the KeyManager can be accessed here: auth.DefaultKeyManager()
```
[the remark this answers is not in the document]
michaelkleinhenz: Leftover from when there was only one single manager for keys and tokens. This will be replaced in the next PR anyway.

pkg/controller/health_check.go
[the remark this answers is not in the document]
alexeykazakov: But this is "production" endpoint. This code is supposed to be executed in production. It should not make the health check always returning hardcoded OK just because it makes testing easier. Also, how would you test the usecase when the health check fails if running in test mode? :)

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/codeready-toolchain/registration-service
 go 1.12
 
 require (
+	github.com/appleboy/gin-jwt/v2 v2.6.2
```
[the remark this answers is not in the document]
michaelkleinhenz: Fixed 684a996.

pkg/middleware/middleware.go
```diff
@@ -0,0 +1,124 @@
...
+	EmailKey = "email"
+	// JWTClaimsKey is the context key for the claims struct
+	JWTClaimsKey = "jwtClaims"
+)
+
+// JWTMiddleware is the JWT token validation middleware
+type JWTMiddleware struct {
+	config      *configuration.Registry
+	logger      *log.Logger
+	keyManager  *auth.KeyManager
```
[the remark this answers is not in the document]
michaelkleinhenz: Fixed 8c5ddc2.

pkg/middleware/middleware.go
```diff
@@ -0,0 +1,124 @@
...
+	tokenParser *auth.TokenParser
+}
+
+// NewAuthMiddleware returns a new middleware for JWT authentication
+func NewAuthMiddleware(logger *log.Logger, config *configuration.Registry) (*JWTMiddleware, error) {
+	if logger == nil || config == nil {
+		return nil, errors.New("missing parameters for NewAuthMiddleware")
+	}
+	// wire up the key and token management
+	keyManagerInstance, err := auth.DefaultKeyManager()
```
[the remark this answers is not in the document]
michaelkleinhenz: Fixed 8c5ddc2.

pkg/middleware/middleware.go
```diff
@@ -0,0 +1,124 @@
...
+				return s[1], nil
+			}
+			// we're failing fast here, if there is an Authorization header, it is used or it fails
+			return "", errors.New("found bearer token header, but no token:" + headerToken)
+		}
+		// see above, failing fast
+		return "", errors.New("found unknown authorization header:" + headerToken)
+	}
+	// next, try GET param "token" (will return "" if n/a)
+	paramToken := c.Query("token")
```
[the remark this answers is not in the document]
michaelkleinhenz: I looked through how other middlewares do that and almost all of them support the following methods in order:
1. Auth Header
2. Urlencoded param
3. Client-side cookie
I do think (3) is a bit esoteric, so I sticked to (1) and (2). So for compliance to how other auth middlewares do that, I'll keep that.

pkg/middleware/middleware.go
```diff
@@ -0,0 +1,124 @@
...
+	if paramToken != "" {
+		return paramToken, nil
+	}
+	return "", errors.New("no token found")
+}
+
+func (m *JWTMiddleware) respondWithError(c *gin.Context, code int, message interface{}) {
+	c.AbortWithStatusJSON(code, gin.H{"error": message})
```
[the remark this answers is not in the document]
michaelkleinhenz: Created an issue: https://jira.coreos.com/browse/CRT-266

pkg/middleware/middleware.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// HandlerFunc returns the HanderFunc.
+func (m *JWTMiddleware) HandlerFunc() gin.HandlerFunc {
+	return func(c *gin.Context) {
+		// check if we have a token
+		tokenStr, err := m.extractToken(c)
+		if err != nil {
+			m.respondWithError(c, http.StatusForbidden, err.Error())
```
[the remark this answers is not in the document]
michaelkleinhenz: Fixed c1df2b9.

pkg/middleware/middleware.go
```diff
@@ -0,0 +1,124 @@
...
+			c.Abort()
+			return
+		}
+		// next, check the token
+		token, err := m.tokenParser.FromString(tokenStr)
+		if err != nil {
+			m.respondWithError(c, http.StatusForbidden, err.Error())
```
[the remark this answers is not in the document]
michaelkleinhenz: Fixed c1df2b9.

pkg/middleware/middleware.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// HandlerFunc returns the HanderFunc.
+func (m *JWTMiddleware) HandlerFunc() gin.HandlerFunc {
+	return func(c *gin.Context) {
+		// check if we have a token
+		tokenStr, err := m.extractToken(c)
+		if err != nil {
+			m.respondWithError(c, http.StatusForbidden, err.Error())
+			c.Abort()
```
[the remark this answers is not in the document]
michaelkleinhenz: Yeah, this is doubled. Fixed b9918e1.

pkg/middleware/middleware.go
```diff
@@ -0,0 +1,124 @@
...
+			return
+		}
+		// next, check the token
+		token, err := m.tokenParser.FromString(tokenStr)
+		if err != nil {
+			m.respondWithError(c, http.StatusForbidden, err.Error())
+			c.Abort()
```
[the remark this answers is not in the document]
michaelkleinhenz: Yeah, this is doubled. Fixed b9918e1.

pkg/middleware/middleware.go
```diff
@@ -0,0 +1,124 @@
...
+			return
+		}
+		// validate time claims
+		if token.Valid() != nil {
+			m.respondWithError(c, http.StatusForbidden, "token has invalid time claims")
```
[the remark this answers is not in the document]
michaelkleinhenz: Fixed c1df2b9.

pkg/middleware/middleware.go
```diff
@@ -0,0 +1,124 @@
...
+			c.Abort()
```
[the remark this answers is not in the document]
michaelkleinhenz: Fixed c1df2b9.

pkg/middleware/middleware.go
```diff
@@ -0,0 +1,124 @@
...
+			return
+		}
+		// check if we have the needed claims for username and email
+		if token.Username == "" {
+			m.respondWithError(c, http.StatusForbidden, "token does not have preferred_username set")
```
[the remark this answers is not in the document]
michaelkleinhenz: Fixed c1df2b9.

pkg/middleware/middleware.go
```diff
@@ -0,0 +1,124 @@
...
+			c.Abort()
```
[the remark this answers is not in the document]
michaelkleinhenz: Fixed c1df2b9.

pkg/middleware/middleware.go
```diff
@@ -0,0 +1,124 @@
...
+			return
+		}
+		if token.Email == "" {
+			m.respondWithError(c, http.StatusForbidden, "token does not have email set")
```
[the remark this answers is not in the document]
michaelkleinhenz: Fixed c1df2b9.

pkg/middleware/middleware.go
```diff
@@ -0,0 +1,124 @@
...
+			c.Abort()
```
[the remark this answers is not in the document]
michaelkleinhenz: Fixed c1df2b9.

pkg/middleware/middleware_test.go
michaelkleinhenz: Fixed c85005c.

pkg/server/routes.go
michaelkleinhenz: Fixed f749e9c.

pkg/middleware/middleware.go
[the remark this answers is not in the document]
alexeykazakov: Urlencoded params are visible in browser UI, they often leaked to logs, etc.
-1 to supporting them until we have some real use case which will force us to use them.

pkg/server/routes.go
michaelkleinhenz: Fixed 1da1617.

pkg/middleware/middleware_test.go
michaelkleinhenz: Yeah, but `tokengenerator.GenerateSignedToken()` is already doing that. The only difference is that I need to give the key explicitly to the instance of `TokenManager` and I think this is good because it makes the test explicitly state which key to use. I guess this might lead to hard to read test code if we use defaults here as the tests are frequently explicitly manipulate the available keys for the `TokenManager`.

Plus, this test is actually different as it is actually fetching the keys. Other tests directly use a ready prepped `TokenParser`. For the middleware, this is different here as it is more or less a blackbox test.

pkg/middleware/middleware.go
michaelkleinhenz: I usually like to have all components check their "owning" scope independently so there is no dependency on checks of vital states outside of the component. But yeah, removed in ceca752, added comment.

pkg/middleware/middleware.go
michaelkleinhenz: Fixed a803421.

pkg/server/routes.go
michaelkleinhenz: No, because the TokenParser might use a different KeyManager. Not yet, sure, but maybe later. TokenParser is completely independent and stateless. I wouldn't pull this into a state unneccessarily. Also, this makes testing easier.

pkg/server/routes.go
michaelkleinhenz: Fixed d4c84a5.

pkg/server/routes.go
michaelkleinhenz: Fixed d4c84a5.

pkg/server/routes.go
michaelkleinhenz: Fixed 476fb1b.

pkg/middleware/middleware.go
alexeykazakov: Then add a test for that vital state ;) So if the dependency behavior unexpectedly changes the test will catch it.
In other words it makes sense to have some redundancy in this tests. But not in implementation.

pkg/server/routes.go
alexeykazakov: > No, because the TokenParser might use a different KeyManager.

A token parser yes. But not the default one.

alexeykazakov: Can you provide a single use case when the default parser would need to use a different key manager?

Also, it’s already stateful since it holds a key manager. I don’t see how using the default manager in the default parser changes it.

alexeykazakov:
Can you add more test for different errors which are missed currently (see the codecov report)?

pkg/server/routes.go
michaelkleinhenz: Ok ok, I give up :-)
I restructured to have the KeyManager stuff not exported. So only TokenParser is used from the outside. See 6f342e0.

pkg/middleware/middleware.go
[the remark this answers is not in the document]
michaelkleinhenz: Fixed befa7a1.

pkg/middleware/middleware.go
michaelkleinhenz: But this is runtime, not test scope. Anyways, it is checked in tokenparser only now.

michaelkleinhenz:
> Can you add more test for different errors which are missed currently (see the codecov report)?

Added more tests in cfbfcef.

pkg/auth/defaultmanager.go
```diff
@@ -6,58 +6,68 @@ import (
...
 		if err != nil {
 			return nil, err
 		}
-		return defaultKeyManager, nil
+		return defaultKeyManagerHolder, nil
 	}
 	return nil, errors.New("default KeyManager can be created only once")
 }
 
 // DefaultKeyManager returns the existing KeyManager instance.
```
alexeykazakov:
```suggestion
// defaultKeyManager returns the existing KeyManager instance.
```

michaelkleinhenz: Fixed in #30

PR #30, opened 2019-09-20
JavaScript Wiring for Keycloak

michaelkleinhenz:
This fits everything together and wires up in the client UI.


pkg/controller/authconfig_test.go
```diff
@@ -57,61 +57,81 @@ func (s *TestAuthConfigSuite) TestAuthClientConfigHandler() {
 
 		// Check the response body is what we expect.
 		// get config values from endpoint response
-		var data map[string]interface{}
-		err = json.Unmarshal(rr.Body.Bytes(), &data)
+		var dataEnvelope map[string]interface{}
+		err = json.Unmarshal(rr.Body.Bytes(), &dataEnvelope)
```
alexeykazakov: Unmarshal it to a `configResponse` object instead of using a map. It will test that we encode that struct.

alexeykazakov: Fixed in https://github.com/codeready-toolchain/registration-service/pull/30/commits/8d5131c64d419d622cabcaa301530e037da92bc1

alexeykazakov: Also simplified the `assert.Equals()` error message. No need to explicitly add expected and actual values in the message args. It's already added by `assert.Equals()` to the output.

PR #33, opened 2019-09-24
check-go-format

tinakurian:
adding check-go-format

### DIFF
--- a/make/format.mk
+++ b/make/format.mk
@@ -2,3 +2,12 @@
 ## format go code
 format:
 	gofmt -s -l -w $(shell find  . -name '*.go' | grep -vEf .gofmt_exclude)
+
+check-go-format:
+	@gofmt -s -l $(shell find  . -name '*.go' | grep -vEf .gofmt_exclude) 2>&1 \
+		| tee /tmp/gofmt-errors \
+		| read \
+	&& echo "ERROR: These files differ from gofmt's style (run 'make format-go-code' to fix this):" \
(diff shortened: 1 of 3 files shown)


make/format.mk
```diff
@@ -2,3 +2,12 @@
 ## format go code
 format:
 	gofmt -s -l -w $(shell find  . -name '*.go' | grep -vEf .gofmt_exclude)
+
+check-go-format:
+	@gofmt -s -l $(shell find  . -name '*.go' | grep -vEf .gofmt_exclude) 2>&1 \
+		| tee /tmp/gofmt-errors \
+		| read \
+	&& echo "ERROR: These files differ from gofmt's style (run 'make format-go-code' to fix this):" \
```
alexeykazakov: Let's add `make format-go-code` too.

alexeykazakov: Or we can just call it `make format` IMO.

tinakurian: @alexeykazakov  :-) i think it's already here - couple lines above: https://github.com/codeready-toolchain/registration-service/pull/33/files#diff-3a7beb51a6188f6f4ab9a40bb5581d42R3 

but i can fix the message to say run 'make format' instead! good catch.

tinakurian: fixed here: 46f1911556ed12fd81ce558a947e8e8a27c0c050

alexeykazakov: Oh. Right. I missed that :)

PR #34, opened 2019-09-24
Use struct to construct json payload instead of map

tinakurian:
makes the change to use structs instead of maps when constructing json payload. 

fixes: https://jira.coreos.com/browse/CRT-256

### DIFF
--- a/pkg/controller/health_check.go
+++ b/pkg/controller/health_check.go
@@ -6,13 +6,19 @@ import (
 	"net/http"
 
 	"github.com/codeready-toolchain/registration-service/pkg/configuration"
+	"github.com/codeready-toolchain/registration-service/pkg/errors"
 	"github.com/gin-gonic/gin"
 )
 
 // HealthCheck implements the health endpoint.
 type HealthCheck struct {
-	config *configuration.Registry
-	logger *log.Logger
+	config      *configuration.Registry
+	logger      *log.Logger
+	Alive       bool
+	TestingMode bool
+	Revision    string
+	BuildTime   string
+	StartTime   string
 }
 
 // HealthCheck returns a new HealthCheck instance.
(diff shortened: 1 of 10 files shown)


pkg/errors/error_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+	s.Run("check json error payload", func() {
+		details := "testing payload"
+		errMsg := "testing new error"
+		code := http.StatusInternalServerError
+
+		err.EncodeError(ctx, errors.New(errMsg), code, details)
+
+		res := err.Error{}
+		json.Unmarshal(rr.Body.Bytes(), &res)
```
golangcibot: Error return value of `json.Unmarshal` is not checked (from `errcheck`)

pkg/controller/health_check.go
```diff
@@ -6,13 +6,19 @@ import (
...
 	"github.com/gin-gonic/gin"
 )
 
 // HealthCheck implements the health endpoint.
 type HealthCheck struct {
-	config *configuration.Registry
-	logger *log.Logger
+	config      *configuration.Registry
+	logger      *log.Logger
+	Alive       bool
```
alexeykazakov: We need to separate the HealthCheck controller from the Health Payload. These are two different objects.

tinakurian: Yeah agreed, fixed here: be101406e7a9d5eba80a3ba6bd2e55510c51d8e8

pkg/middleware/middleware_test.go
```diff
@@ -96,35 +98,66 @@ func (s *TestAuthMiddlewareSuite) TestAuthMiddlewareService() {
...
-	}
+	s.Run("health check requests", func() {
+		health := &controller.Health{}
+		resp := httptest.NewRecorder()
+		req, err := http.NewRequest(http.MethodGet, "/api/v1/health", nil)
+		require.NoError(s.T(), err)
+
+		srv.Engine().ServeHTTP(resp, req)
+		// Check the status code is what we expect.
+		assert.Equal(s.T(), http.StatusOK, resp.Code, "request returned wrong status code: got %v want %v", resp.Code, http.StatusOK)
```
alexeykazakov: No need to pass the expected and actual values as the message args. `asser.Equal()` will include it into the test failure message for you.
```suggestion
		assert.Equal(s.T(), http.StatusOK, resp.Code, "request returned wrong status code")
```

pkg/middleware/middleware_test.go
```diff
@@ -96,35 +98,66 @@ func (s *TestAuthMiddlewareSuite) TestAuthMiddlewareService() {
...
+	})
+
+	s.Run("static request", func() {
+		resp := httptest.NewRecorder()
+		req, err := http.NewRequest(http.MethodGet, "/favicon.ico", nil)
+		require.NoError(s.T(), err)
+
+		srv.Engine().ServeHTTP(resp, req)
+		// Check the status code is what we expect.
+		assert.Equal(s.T(), http.StatusOK, resp.Code, "request returned wrong status code: got %v want %v", resp.Code, http.StatusOK)
```
alexeykazakov:
```suggestion
		assert.Equal(s.T(), http.StatusOK, resp.Code, "request returned wrong status code")
```

pkg/middleware/middleware_test.go
```diff
@@ -96,35 +98,66 @@ func (s *TestAuthMiddlewareSuite) TestAuthMiddlewareService() {
...
+			s.Run(tt.name, func() {
+				resp := httptest.NewRecorder()
+				req, err := http.NewRequest(tt.method, tt.urlPath, nil)
+				require.NoError(s.T(), err)
+				if tt.tokenHeader != "" {
+					req.Header.Set("Authorization", tt.tokenHeader)
+				}
+				srv.Engine().ServeHTTP(resp, req)
+				// Check the status code is what we expect.
+				assert.Equal(s.T(), tt.status, resp.Code, "request returned wrong status code: got %v want %v", resp.Code, tt.status)
```
alexeykazakov:
```suggestion
				assert.Equal(s.T(), tt.status, resp.Code, "request returned wrong status code")
```

PR #36, opened 2019-09-26
Make targets to deploy registration service

alexeykazakov:
To deploy Registration Service on local minishift:

- start your minishift then run:

```
eval $(minishift docker-env)
make deploy-dev
make update-etc-hosts
```

The service will be available at http://minishift.local:31000

This PR also:
- [x] adds `openshift-ci` docker images as preparation for openshift-ci 
- [x] adds `deployment/deployment_dev_cluster.yaml` as preparation for deploying to dev cluster

See CRT-250


dev.mk
```diff
@@ -0,0 +1,72 @@
+DOCKER_REPO?=quay.io/codeready-toolchain
```
xcoulon: why do we have `dev.mk` and `make/dev.mk` ?

alexeykazakov: Oh... good catch. I will delete dev.mk. Not sure what it’s doing here :)

alexeykazakov: Done in https://github.com/codeready-toolchain/registration-service/pull/36/commits/f2684947bcfaa9e2c8dc0b13fc20bc20ff913363

deploy/deployment_dev.yaml
```diff
@@ -0,0 +1,68 @@
+apiVersion: apps/v1
+kind: Deployment
```
xcoulon: I believe that if we use OpenShift `DeploymentConfig`, we won't need the `Service` definition below

deploy/deployment_dev_cluster.yaml
```diff
@@ -0,0 +1,87 @@
+apiVersion: image.openshift.io/v1
```
xcoulon: can't we have the same manifest for "local" and "dev cluster" deployments? we could even take advantage of this for local testing: just push a new image to the container registry and get the application redeployed with the DeployConfig + ImageStream combination.

deploy/role.yaml
```diff
@@ -0,0 +1,18 @@
...
+metadata:
+  name: registration-service
+rules:
+- apiGroups:
+  - toolchain.dev.openshift.com
+  resources:
+  - usersignups
+  verbs:
+  - create
+  - get
```
xcoulon: is there no need to `list` for now?

deploy/role.yaml
```diff
@@ -0,0 +1,18 @@
...
+- apiGroups:
+  - toolchain.dev.openshift.com
+  resources:
+    - masteruserrecord
+  verbs:
+    - get
```
xcoulon: same here: no need to `list`?

make/clean.mk
```diff
@@ -4,4 +4,6 @@ clean:
 	@cd "$(GOPATH)/src/github.com/codeready-toolchain/registration-service" && \
 		rm -f pkg/static/generated_assets.go && \
 		rm -rf $(COV_DIR) && \
-		rm -f registration-service
+		rm -f registration-service && \
```
xcoulon: shouldn't it be `./bin/registration-service` instead? The Makefile `build-dev` and `build-prod` goals generate the binary in `-o $(OUT_DIR)/bin/registration-service`

make/dev.mk
```diff
@@ -0,0 +1,72 @@
...
+## Setup service account and deploy RBAC
+deploy-rbac:
+	$(Q)-oc apply -f deploy/service_account.yaml
+	$(Q)-oc apply -f deploy/role.yaml
+	$(Q)-oc apply -f deploy/role_binding.yaml
+
+.PHONY: deploy-dev
+## Deploy Registration service on minishift
+deploy-dev: login-as-admin create-namespace deploy-rbac build image check-hosts
+# 	$(Q)-eval `minishift docker-env`
```
xcoulon: you can remove this commented-out line if it's not needed

make/dev.mk
```diff
@@ -0,0 +1,72 @@
...
+
+.PHONY: check-hosts
+## Add minishift ip to /etc/hosts
+check-hosts:
+	echo $(MINISHIFT_IP) $(HMINISHIFT_HOSTNAME) $(ETC_HOSTS)
+	if grep -q "$(MINISHIFT_IP) $(MINISHIFT_HOSTNAME)" $(ETC_HOSTS); then \
+    	echo "Hosts entry exists"; \
+	else \
+    	echo "Updating /etc/hosts (Remove old minishift.local if any and add new one)"; \
+    	# remove old entries with $(HOSTNAME) if any \
```
xcoulon: same here: you can remove these commented-out lines

deploy/role.yaml
alexeykazakov: Correct. No need for list for now.

deploy/role.yaml
alexeykazakov: Correct. Will add list later if we need. But for now let’s grand minimum permissions we need.

xcoulon: ok, sounds reasonable/pragmatic to only address today's requirements.

deploy/deployment_dev_cluster.yaml
xcoulon: ok, let's keep this PR as short as possible :)

make/clean.mk
alexeykazakov: Right. It’s a leftover from the previous binary path.

make/dev.mk
alexeykazakov: Okey :)

make/dev.mk
alexeykazakov: It’s a comment to describe what is being done here :)

xcoulon: ah right, I was too fast in reviewing this line 🤦‍♂

deploy/deployment_dev.yaml
alexeykazakov: Are you sure?

deploy/deployment_dev_cluster.yaml
```diff
@@ -0,0 +1,87 @@
...
+    spec:
+      serviceAccountName: registration-service
+      containers:
+      - name: registration-service
+        image: ''
+        command:
+        - registration-service
+        imagePullPolicy: IfNotPresent
+        env:
+        - name: WATCH_NAMESPACE
```
dipak-pawar: I think, it's not needed for registration service. do we?

deploy/deployment_dev_cluster.yaml
```diff
@@ -0,0 +1,87 @@
...
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        - name: POD_NAME
```
dipak-pawar: I think, it's not needed for registration service. do we?

deploy/deployment_dev_cluster.yaml
dipak-pawar: @xcoulon, we can't use the same setup for development, as service needs to be created as `NodePort`, Also, imagePullPolicy is different (`IfNotPresent` -> for minishift to work with `eval $(minishift docker-env)` and `Always` for production environment.

We can have some script to generate this on fly, but as we are getting rid out of this, it's not needed

deploy/deployment_dev_cluster.yaml
```diff
@@ -0,0 +1,87 @@
...
+      scheduled: true
+    name: registration-service
+---
+apiVersion: apps.openshift.io/v1
+kind: DeploymentConfig
+metadata:
+  labels:
+    app: registration-service
+  name: registration-service
+  namespace: toolchain-registration-service
```
dipak-pawar: If we deploy registration service to this namespace, then we'll be creating sa, role and rolebinding in the namespace. This service account used in registration service will create usersignup custom resource, but if host-operator is not watching the same namespace for required usesignup resource, then it will fail to reconcile

deploy/deployment_dev_cluster.yaml
```diff
@@ -0,0 +1,87 @@
...
+spec:
+  replicas: 3
```
dipak-pawar: how does this 3 number come?

deploy/deployment_dev_cluster.yaml
```diff
@@ -0,0 +1,87 @@
...
+        app: registration-service
+        deploymentconfig: registration-service
+    spec:
+      serviceAccountName: registration-service
+      containers:
+      - name: registration-service
+        image: ''
+        command:
+        - registration-service
+        imagePullPolicy: IfNotPresent
```
dipak-pawar: I think, we should use `Always` as it's running in the dev cluster, which is production for us

make/dev.mk
```diff
@@ -0,0 +1,72 @@
...
+	$(Q)-oc apply -f deploy/role_binding.yaml
+
+.PHONY: deploy-dev
+## Deploy Registration service on minishift
+deploy-dev: login-as-admin create-namespace deploy-rbac build image check-hosts
+# 	$(Q)-eval `minishift docker-env`
+	$(Q)-oc new-project $(LOCAL_TEST_NAMESPACE) || true
+	$(Q)-sed -e 's|REPLACE_IMAGE|${IMAGE_NAME}|g' ./deploy/deployment_dev.yaml  | oc apply -f -
+
+.PHONY: check-hosts
```
dipak-pawar: we can ask user to look at created route using `oc get routes registration-service` instead of updating host entries

openshift-ci/Dockerfile.tools
```diff
@@ -0,0 +1,59 @@
...
+    && echo "4b0f07428ba854174c58d2e38287e5402964c9a9355f6c359d1242efd0990da3 openshift-origin-client-tools.tar.gz" > openshift-origin-client-tools.sha256 \
+    && sha256sum -c openshift-origin-client-tools.sha256 \
+    && tar xzf openshift-origin-client-tools.tar.gz \
+    && mv /tmp/openshift-origin-client-tools-*/oc /usr/bin/oc \
+    && mv /tmp/openshift-origin-client-tools-*/kubectl /usr/bin/kubectl \
+    && rm -rf ./openshift* \
+    && oc version
+
+# download, verify and install operator-sdk
+RUN curl -L -s https://github.com/operator-framework/operator-sdk/releases/download/${OPERATOR_SDK_VERSION}/operator-sdk-${OPERATOR_SDK_VERSION}-x86_64-linux-gnu -o operator-sdk \
```
dipak-pawar: Do we need this for registration service?

openshift-ci/Dockerfile.tools
```diff
@@ -0,0 +1,59 @@
...
+    kubectl \
+    yamllint \
+    python36-virtualenv \
+    jq \
+    go-bindata \
+    && yum clean all
+
+# download, verify and install openshift client tools (oc and kubectl)
+WORKDIR /tmp
+RUN curl -L -s https://github.com/openshift/origin/releases/download/v3.11.0/openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit.tar.gz -o openshift-origin-client-tools.tar.gz \
```
dipak-pawar: Do we need this for registration service?

make/dev.mk
```diff
@@ -0,0 +1,72 @@
+MINISHIFT_IP?=$(shell minishift ip)
+MINISHIFT_HOSTNAME=minishift.local
+MINISHIFT_HOSTNAME_REGEX='minishift\.local'
+ETC_HOSTS=/etc/hosts
+
+# to watch all namespaces, keep namespace empty
+APP_NAMESPACE ?= $(LOCAL_TEST_NAMESPACE)
+LOCAL_TEST_NAMESPACE ?= "toolchain-host-operator"
```
dipak-pawar: Do we want to run it in the same namespace where we are running host operator?

deploy/deployment_dev.yaml
dipak-pawar: We don't need any service for operators, as it's not publicly exposed to the outside the cluster and also it's not rest service

deploy/deployment_dev.yaml
```diff
@@ -0,0 +1,68 @@
...
+      - name: registration-service
+        # Replace this with the built image name
+        image: REPLACE_IMAGE
+        ports:
+          - containerPort: 8080
+        command:
+        - registration-service
+        imagePullPolicy: IfNotPresent
+        env:
+        - name: WATCH_NAMESPACE
```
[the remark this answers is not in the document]
alexeykazakov: Right. We don’t need it. It’s a leftover. I will remove it.

deploy/deployment_dev.yaml
```diff
@@ -0,0 +1,68 @@
...
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        - name: POD_NAME
```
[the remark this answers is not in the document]
alexeykazakov: Another leftover :)

make/dev.mk
alexeykazakov: Yes we could. Modifying etc/hosts could be more convenient since it would not require that additional step. But it requires sudo access though...
So, what do you think would be better?

openshift-ci/Dockerfile.tools
alexeykazakov: Not right now. But it’s preparation for running e2e tests in openshif-ci. We will need operator-sdk for that.

make/dev.mk
alexeykazakov: It was my intention. In prod we will run reg service in the host operator namespace. So I thought it would make sense to do the same in dev env.

dipak-pawar: I do agree that, we need to run this service in the same namespace where we'll be running host-operator, but then we need to use the same here - https://github.com/codeready-toolchain/registration-service/pull/36/files#r329978845

make/dev.mk
dipak-pawar: I would prefer to have routes, this way we would be having the same way for dev and production and also no manual interaction which is needed for sudo

make/dev.mk
alexeykazakov: I'm going to remove deployment_dev_cluster.yaml from this PR :)

deploy/deployment_dev_cluster.yaml
alexeykazakov: Done in 8af6a1af6a8d69f5871dffae62436a7dd8b011f2

deploy/deployment_dev_cluster.yaml
alexeykazakov: Done in 8af6a1af6a8d69f5871dffae62436a7dd8b011f2

deploy/deployment_dev_cluster.yaml
alexeykazakov: Done in 8af6a1af6a8d69f5871dffae62436a7dd8b011f2

deploy/deployment_dev.yaml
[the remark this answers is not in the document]
alexeykazakov: Done in 148261b80b3db44f5cf8f1b14ee8ee35f6bdb8f0

deploy/deployment_dev.yaml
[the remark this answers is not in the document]
alexeykazakov: Done in 148261b80b3db44f5cf8f1b14ee8ee35f6bdb8f0

make/clean.mk
alexeykazakov: Done in 8af6a1af6a8d69f5871dffae62436a7dd8b011f2

make/dev.mk
alexeykazakov: We need a route anyway.

I made `/etc/hosts` modifications optional (see 547945a6f9972ba535eb98f71a88c6b6b3c700be).

alexeykazakov:
@xcoulon @dipak-pawar I've addressed all your comments. Please re-review :)

PR #38, opened 2019-10-01
Global Logging

tinakurian:


pkg/log/log.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// addContextInfo adds fields extracted from the context to the info/error
+// log messages.
+func addContextInfo(ctx *gin.Context, v ...interface{}) []interface{} {
+	if ctx != nil {
+		subject := ctx.GetString("subject")
+		if subject != ""{
+			v = append(v, fmt.Sprintf("context subject"))
+			v = append(v, fmt.Sprintf(subject))
```
golangcibot: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (from `staticcheck`)

michaelkleinhenz:
LGTM

pkg/auth/keymanager.go
```diff
@@ -140,11 +138,11 @@ func (km *KeyManager) fetchKeys(keysEndpointURL string) ([]*PublicKey, error) {
 	defer func() {
 		_, err := ioutil.ReadAll(res.Body)
 		if err != io.EOF && err != nil {
-			km.logger.Println("failed read remaining data before closing response")
+			log.Info(nil, "failed read remaining data before closing response")
```
alexeykazakov:
```suggestion
			log.Error(nil, err, "failed read remaining data before closing response")
```

pkg/auth/keymanager.go
```diff
@@ -140,11 +138,11 @@ func (km *KeyManager) fetchKeys(keysEndpointURL string) ([]*PublicKey, error) {
...
 		}
 		err = res.Body.Close()
 		if err != nil {
-			km.logger.Println("failed to close response after reading")
+			log.Info(nil, "failed to close response after reading")
```
alexeykazakov:
```suggestion
			log.Error(nil, err, "failed to close response after reading")
```

pkg/auth/keymanager.go
```diff
@@ -156,7 +154,7 @@ func (km *KeyManager) fetchKeys(keysEndpointURL string) ([]*PublicKey, error) {
 	bodyString := buf.String()
 	// if status code was not OK, bail out
 	if res.StatusCode != http.StatusOK {
-		km.logger.Println(map[string]interface{}{
+		log.Info(nil, "status not OK", map[string]interface{}{
```
alexeykazakov: `log.Info()` doesn't work with maps. It works with slices.
So it should be something like
```
		log.Info(nil, "unable to obtain public keys from remote service: status not OK",
			"response_status", res.Status,
			"response_body", bodyString,
			"url", keysEndpointURL)
```

But see my other comment: https://github.com/codeready-toolchain/registration-service/pull/38/files?file-filters%5B%5D=.go&file-filters%5B%5D=.mod#diff-1135f265d25e96e557b8114567653176R58

pkg/controller/health_check.go
```diff
@@ -56,7 +56,7 @@ func (hc *HealthCheck) GetHandler(ctx *gin.Context) {
 	}
 	err := json.NewEncoder(ctx.Writer).Encode(healthInfo)
 	if err != nil {
-		hc.logger.Println("error writing response body", err.Error())
+		log.Info(ctx, fmt.Sprintf("error writing response body: %s", err.Error()))
```
alexeykazakov: I'm going to stop reviewing all the logger usage here. Can you please verify that you use the logger properly? I see a lot of misuse here. Info instead of errors. Wrong params, etc.

tinakurian: @alexeykazakov i used info for places that we used Println. I assumed we would have used error otherwise.. but this can be easily changed...

pkg/auth/keymanager.go
tinakurian: Since we used Println, i went with info, but i didn't thoroughly review. Will change this..

tinakurian:
> This is what I see in logs when trying different log message:
> 
> ```
> {"level":"info","ts":1570236765.317385,"logger":"logger","msg":"fetching public keys from url: https://sso.prod-preview.openshift.io/auth/realms/fabric8-test/protocol/openid-connect/certs%!(EXTRA []interface {}=[])"}
> {"level":"info","ts":1570236765.6801496,"logger":"logger","msg":"1 public keys loaded%!(EXTRA []interface {}=[])"}
> [GIN-debug] GET    /api/v1/health            --> github.com/codeready-toolchain/registration-service/pkg/controller.(*HealthCheck).GetHandler-fm (3 handlers)
> [GIN-debug] GET    /api/v1/authconfig        --> github.com/codeready-toolchain/registration-service/pkg/controller.(*AuthConfig).GetHandler-fm (3 handlers)
> [GIN-debug] POST   /api/v1/signup            --> github.com/codeready-toolchain/registration-service/pkg/controller.(*Signup).PostHandler-fm (4 handlers)
> [GIN-debug] GET    /api/v1/auth_test         --> github.com/codeready-toolchain/registration-service/pkg/controller.(*HealthCheck).GetHandler-fm (4 handlers)
> {"level":"info","ts":1570236765.68035,"logger":"logger","msg":"Configured routes: ROUTE: \tRoute Path: /api/v1/authconfig\n\tMethod: GET\nROUTE: \tRoute Path: /api/v1/auth_test\n\tMethod: GET\nROUTE: \tRoute Path: /api/v1/health\n\tMethod: GET\nROUTE: \tRoute Path: /api/v1/signup\n\tMethod: POST\n%!(EXTRA []interface {}=[])"}
> {"level":"info","ts":1570236765.6805515,"logger":"logger","msg":"Service Revision d7199cbd988dbd6886c00d7cea9be5aadc79114b-dirty built on 0%!(EXTRA []interface {}=[])"}
> {"level":"info","ts":1570236765.6806612,"logger":"logger","msg":"Listening on \"0.0.0.0:8080\"...%!(EXTRA []interface {}=[])"}
> [GIN] 2019/10/04 - 17:52:49 | 401 |      91.641µs |       127.0.0.1 | GET      /api/v1/auth_test
> [GIN] 2019/10/04 - 17:52:50 | 200 |    2.755471ms |       127.0.0.1 | GET      /favicon.ico
> {"level":"error","ts":1570236779.1020632,"logger":"logger","msg":"","context host":"","error":"oopsie woopsie","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/home/igels/go/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128\ngithub.com/codeready-toolchain/registration-service/pkg/log.Error\n\t/home/igels/go/src/github.com/codeready-toolchain/registration-service/pkg/log/log.go:66\ngithub.com/codeready-toolchain/registration-service/pkg/controller.(*HealthCheck).GetHandler\n\t/home/igels/go/src/github.com/codeready-toolchain/registration-service/pkg/controller/health_check.go:63\ngithub.com/gin-gonic/gin.(*Context).Next\n\t/home/igels/go/pkg/mod/github.com/gin-gonic/gin@v1.4.0/context.go:124\ngithub.com/gin-gonic/gin.RecoveryWithWriter.func1\n\t/home/igels/go/pkg/mod/github.com/gin-gonic/gin@v1.4.0/recovery.go:83\ngithub.com/gin-gonic/gin.(*Context).Next\n\t/home/igels/go/pkg/mod/github.com/gin-gonic/gin@v1.4.0/context.go:124\ngithub.com/gin-gonic/gin.LoggerWithConfig.func1\n\t/home/igels/go/pkg/mod/github.com/gin-gonic/gin@v1.4.0/logger.go:240\ngithub.com/gin-gonic/gin.(*Context).Next\n\t/home/igels/go/pkg/mod/github.com/gin-gonic/gin@v1.4.0/context.go:124\ngithub.com/gin-gonic/gin.(*Engine).handleHTTPRequest\n\t/home/igels/go/pkg/mod/github.com/gin-gonic/gin@v1.4.0/gin.go:389\ngithub.com/gin-gonic/gin.(*Engine).ServeHTTP\n\t/home/igels/go/pkg/mod/github.com/gin-gonic/gin@v1.4.0/gin.go:351\nnet/http.serverHandler.ServeHTTP\n\t/snap/go/4517/src/net/http/server.go:2802\nnet/http.(*conn).serve\n\t/snap/go/4517/src/net/http/server.go:1890"}
> [GIN] 2019/10/04 - 17:52:59 | 200 |     217.013µs |       127.0.0.1 | GET      /api/v1/health
> [GIN] 2019/10/04 - 17:54:15 | 304 |      85.326µs |       127.0.0.1 | GET      /
> [GIN] 2019/10/04 - 17:54:15 | 200 |       77.77µs |       127.0.0.1 | GET      /api/v1/authconfig
> [GIN] 2019/10/04 - 17:54:21 | 200 |      52.666µs |       127.0.0.1 | GET      /api/v1/authconfig
> {"level":"error","ts":1570236862.2222335,"logger":"logger","msg":"","context subject":"27581614-86f7-4597-9af8-a3d00cd93f55","context host":"","error":"oopsie woopsie","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/home/igels/go/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128\ngithub.com/codeready-toolchain/registration-service/pkg/log.Error\n\t/home/igels/go/src/github.com/codeready-toolchain/registration-service/pkg/log/log.go:66\ngithub.com/codeready-toolchain/registration-service/pkg/controller.(*HealthCheck).GetHandler\n\t/home/igels/go/src/github.com/codeready-toolchain/registration-service/pkg/controller/health_check.go:63\ngithub.com/gin-gonic/gin.(*Context).Next\n\t/home/igels/go/pkg/mod/github.com/gin-gonic/gin@v1.4.0/context.go:124\ngithub.com/codeready-toolchain/registration-service/pkg/middleware.(*JWTMiddleware).HandlerFunc.func1\n\t/home/igels/go/src/github.com/codeready-toolchain/registration-service/pkg/middleware/middleware.go:90\ngithub.com/gin-gonic/gin.(*Context).Next\n\t/home/igels/go/pkg/mod/github.com/gin-gonic/gin@v1.4.0/context.go:124\ngithub.com/gin-gonic/gin.RecoveryWithWriter.func1\n\t/home/igels/go/pkg/mod/github.com/gin-gonic/gin@v1.4.0/recovery.go:83\ngithub.com/gin-gonic/gin.(*Context).Next\n\t/home/igels/go/pkg/mod/github.com/gin-gonic/gin@v1.4.0/context.go:124\ngithub.com/gin-gonic/gin.LoggerWithConfig.func1\n\t/home/igels/go/pkg/mod/github.com/gin-gonic/gin@v1.4.0/logger.go:240\ngithub.com/gin-gonic/gin.(*Context).Next\n\t/home/igels/go/pkg/mod/github.com/gin-gonic/gin@v1.4.0/context.go:124\ngithub.com/gin-gonic/gin.(*Engine).handleHTTPRequest\n\t/home/igels/go/pkg/mod/github.com/gin-gonic/gin@v1.4.0/gin.go:389\ngithub.com/gin-gonic/gin.(*Engine).ServeHTTP\n\t/home/igels/go/pkg/mod/github.com/gin-gonic/gin@v1.4.0/gin.go:351\nnet/http.serverHandler.ServeHTTP\n\t/snap/go/4517/src/net/http/server.go:2802\nnet/http.(*conn).serve\n\t/snap/go/4517/src/net/http/server.go:1890"}
> [GIN] 2019/10/04 - 17:54:22 | 200 |     971.878µs |       127.0.0.1 | GET      /api/v1/auth_test
> ```
> 
> It's a mixture of gin and our own logging. Gin logs are more readable but our logs have some extra info which we need. So:
> 
> * Can we change the Gin logger for better consistency? Not sure if it worth it though...
> * There is no any time/date in our logs. We need to add it.
> * The only user information I see is `"context subject":"27581614-86f7-4597-9af8-a3d00cd93f55"`. It's not clear what it means and it's not enough. Let's add `"user_id":"<sub>", "username":"<username>". Also we need to add "req_id":"<"X-Request-Id" header>`

To be completely honest, i had tested out logrus and it seemed tidy to me in terms of the logs that were printed. I had added timestamps that made sense (i think the "ts" field is the timestamp here..? but it seems somewhat useless...)... GIN logging seems just as readable as logrus. So i'd be in favour of switching to using GIN logging at this point. Would have been nice to use the same logger as in out controllers but... this seems better imho.

alexeykazakov:
> To be completely honest, i had tested out logrus and it seemed tidy to me in terms of the logs that were printed. I had added timestamps that made sense (i think the "ts" field is the timestamp here..? but it seems somewhat useless...)... GIN logging seems just as readable as logrus. So i'd be in favour of switching to using GIN logging at this point. Would have been nice to use the same logger as in our controllers but... this seems better for the registration app imho.

The format we see in console/terminal is not that important. Yes, it can look nice but we won't look at console logs in prod. We are going to feed all logs to some service like kibana and access it there.
There are multiple reasons for using some log aggregator instead of logs in the service/controller container:
- Much easier to access.
- Permissions.
- Logs don't diaper when the pod dies.
- One place for logs from multiple instances/replicas of the service.
- Searchability.
- etc.

So, we rather should care about the info we provide. Everything should be in key/value format.

pkg/log/log.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// Logger returns the current logger object.
+func Logger() logr.Logger {
+	return log
+}
+
+// Info logs are used for non-error messages. It will log a message with
+// the given key/value pairs as context.
+func Info(ctx *gin.Context, msg string, v ...interface{}) {
```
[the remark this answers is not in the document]
xcoulon: I see your point with regards to API consistency, @alexeykazakov. I like this way of passing args compared to the actual k8s logger (which is a bit counter-intuitive and error prone), but if we don't have a use for these `key`, it's going to feel like an overcomplicated API (compared to logrus's `Infof`, `Warnf`, etc..) for example. Yet, having log messages as "constants" and all contextual info in separate parts of the log output is valuable when we need to search for occurrences of an error for example.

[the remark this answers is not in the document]
MatousJobanek: So, first of all, if we don't need to implement some specific interface then I would definitely introduce `Infof` so we don't need to use `Sprintf` every time when we need to format the message (as we can see all over the place in this PR). When we introduce `log.Infof(ctx, msg string, args ...string)` then we cannot use the slice of strings for passing the parameters at the end of the method - we need to put it somewhere else.
The thing that I don't like in logrus style is when I don't need to specify any parameter, then I need to pass an empty map in the function anyway. Thus, I would propose having a method `log.WithValues(keysAndValues ...interface{}) Logger` that returns a logger with the key-values set and then we could call either `Error` or `Info` on that Logger instance. 

So, to sum up - let's have functions:
```go
func Infof(ctx, msg string, args ...string) {}
func Errorf(ctx, err error, msg string, args ...string) {}
func WithValues(keysAndValues ...interface{}) log.Logger {}
```
so we can then use it in this way:
```go
log.Infof(ctx, "the action %s has been processed, action)
log.Errorf(ctx, err, "the action %s has failed, action)
```
or in this way:
```go
log.WithValues(map[string]interface{}{
	"name": "john",
}).Errorf(ctx, err, "the %s action has failed", action)
```
or in this way

```go
logger := log.WithValues(map[string]interface{}{
	"name": "john",
})
if err != nil {
    logger.Errorf(ctx, err, "the %s action has failed", action)
    return err
}
logger.Infof(ctx, "the action %s has been processed, action)
```

pkg/auth/tokenparser.go
```diff
@@ -21,20 +22,18 @@ type TokenClaims struct {
...
 type TokenParser struct {
-	logger     *log.Logger
 	keyManager *KeyManager
 }
 
 // NewTokenParser creates a new TokenParser.
-func NewTokenParser(logger *log.Logger, keyManager *KeyManager) (*TokenParser, error) {
-	if logger == nil {
+func NewTokenParser(keyManager *KeyManager) (*TokenParser, error) {
+	if log.Logger() == nil {
```
MatousJobanek: I don't understand why we need this check, because:
* it should have been set in the main, so we can easily panic - if it is not set then it means that there is something really really wrong that shouldn't happen
* if we don't want to fail, then we can lazily create the instance of the logger if it is nil - in the same way as it is done in the main via `log.InitializeLogger("logger")`

pkg/log/log.go
```diff
@@ -0,0 +1,89 @@
...
+	// used), this defaults to a production zap logger.
+	//
+	// The logger instantiated here can be changed to any logger
+	// implementing the logr.Logger interface. This logger will
+	// be propagated through the whole operator, generating
+	// uniform and structured logs.
+	logf.SetLogger(zap.Logger())
+}
+
+func SetOutput(out io.Writer, isTestingMode bool, withName string) {
```
MatousJobanek: May I know what is the purpose of this function?

tinakurian:
> > To be completely honest, i had tested out logrus and it seemed tidy to me in terms of the logs that were printed. I had added timestamps that made sense (i think the "ts" field is the timestamp here..? but it seems somewhat useless...)... GIN logging seems just as readable as logrus. So i'd be in favour of switching to using GIN logging at this point. Would have been nice to use the same logger as in our controllers but... this seems better for the registration app imho.
> 
> The format we see in console/terminal is not that important. Yes, it can look nice but we won't look at console logs in prod. We are going to feed all logs to some service like kibana and access it there.
> There are multiple reasons for using some log aggregator instead of logs in the service/controller container:
> 
> * Much easier to access.
> * Permissions.
> * Logs don't diaper when the pod dies.
> * One place for logs from multiple instances/replicas of the service.
> * Searchability.
> * etc.
> 
> So, we rather should care about the info we provide. Everything should be in key/value format.

Yeah, i was just remarking on the part where you had mentioned it was more readable... 

I will make the change to add the fields you mentioned above..

tinakurian:
> There is no any time/date in our logs. We need to add it.

@alexeykazakov as mentioned the ts field is the timestamp - i'm not sure if that can be formatted differently, will need to look. I didn't add it, it was added automatically...

pkg/log/log.go
```diff
@@ -0,0 +1,89 @@
...
+
+// Info logs are used for non-error messages. It will log a message with
+// the given key/value pairs as context.
+func Info(ctx *gin.Context, msg string, v ...interface{}) {
+	log.Info(fmt.Sprintf(msg, addContextInfo(ctx, v...)))
+}
+
+// Error logs are used for logging errors. It will log the error with the given
+// message and key/value pairs as context.
+func Error(ctx *gin.Context, err error, msg string, v ...interface{}) {
```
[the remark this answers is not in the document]
tinakurian: This was taken from the golang logger: https://golang.org/pkg/log/ they used v and since i had started with that i left it the same. but keysAndValues is more clear..

pkg/log/log.go
tinakurian: it's to be able to redirect our outpit if we don't want it to go to os.Stderr. This would be useful for testing i guess. It also allows me to write tests for these functions by redirecting the output to a bytes.Buffer..

pkg/log/log.go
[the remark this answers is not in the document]
tinakurian: @MatousJobanek sounds good, this looks nice. will make this change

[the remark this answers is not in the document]
MatousJobanek: thanks @tinakurian 
btw, you could also add function:
```go
func WithValue(key, value string) log.Logger {}
```
in case you wanted to add only one key-value pair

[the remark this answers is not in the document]
alexeykazakov: > When we introduce log.Infof(ctx, msg string, args ...string) then we cannot use the slice of strings for passing the parameters at the end of the method - we need to put it somewhere else.

Yeah but this is why I mentioned above that we could just pass `nil` when we don't have any keys-values": `log.Info(ctx, nil, "some info message about %s", arg)`. But I like `WithValues(map[string]interface{}) log.Logger` better!
But we still should use `map[string]interface{}` as the param instead of `keysAndValues ...interface{}` IMO. Exactly as you did it in your examples. It will be less error prone.

alexeykazakov:
> as mentioned the ts field is the timestamp - i'm not sure if that can be formatted differently, will need to look. I didn't add it, it was added automatically...

It could be useful to have the time/date in both raw (int) and human-readable formats. So, maybe worth keeping the current built-in one and add one more - our own one with human readable time.

pkg/auth/keymanager.go
sbryzak: I think Alexey's point is that it's good to include the error object in the log :)

pkg/log/log.go
MatousJobanek: So, then we don't need an additional method in the prod logger just for the sake of using it in tests. I would rather initialize a test logger with the needed redirection/setup at the beginning of the test(s). I believe that you don't need to do this dance every time:
```go
logger.SetOutput(&buf, true, "logger_tests")
defer func() {
	logger.SetOutput(os.Stderr, false, "logger_tests")
}()
```
just create an interface for the current logger and create two implementations - the current one and one for tests.

tinakurian: Actually i didn't need to do it in each sub-test - i moved out so this is fixed here: a873ba650aef35eec3abafe2d036c905c8aeec13

pkg/controller/health_check.go
tinakurian: fixed as part of: 00bf17d961e783ddd6ad166872fdd5903f82a7c1

pkg/auth/keymanager.go
tinakurian: fixed here: 00bf17d961e783ddd6ad166872fdd5903f82a7c1

pkg/auth/keymanager.go
tinakurian: when logging with info there is no option to pass error and since println seems more like something that aligns with info, i had used info instead of error without throoughlylooking at the code. but changing it to log.Error (or now log.Errorf) allows you to pass the error :) fixed here: 00bf17d961e783ddd6ad166872fdd5903f82a7c1

pkg/auth/keymanager.go
tinakurian: fixed here: 00bf17d961e783ddd6ad166872fdd5903f82a7c1

pkg/log/log.go
```diff
@@ -0,0 +1,89 @@
...
+	log.Error(err, msg, junk...)
+}
+
+// addContextInfo adds fields extracted from the context to the info/error
+// log messages.
+func addContextInfo(ctx *gin.Context, v ...interface{}) []interface{} {
+	if ctx != nil {
+		subject := ctx.GetString("subject")
+		if subject != "" {
+			v = append(v, "context subject")
```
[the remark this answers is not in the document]
tinakurian: fixed here: 00bf17d961e783ddd6ad166872fdd5903f82a7c1

pkg/log/log.go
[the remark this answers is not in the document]
tinakurian: fixed here: 00bf17d961e783ddd6ad166872fdd5903f82a7c1

pkg/log/log.go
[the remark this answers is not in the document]
tinakurian: i think the `WithValue` func is not necessary. You can just add a single key/value pair to WithValues if you want to.. however, i still need to change the definition to accept a map..will do that change soon..

pkg/log/log.go
```diff
@@ -0,0 +1,89 @@
...
+			v = append(v, subject)
+		}
+
+		if ctx.Request != nil {
+			url := ctx.Request.URL
+			if url != nil {
+				v = append(v, "context host")
```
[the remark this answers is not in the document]
tinakurian: fixed here: 00bf17d961e783ddd6ad166872fdd5903f82a7c1

pkg/auth/keymanager.go
```diff
@@ -63,7 +62,7 @@ func NewKeyManager(logger *log.Logger, config KeyManagerConfiguration) (*KeyMana
 			km.keyMap[key.KeyID] = key.Key
 		}
 	} else {
-		logger.Println("no public key url given, not fetching keys")
+		logr.Infof(nil, "no public key url given, not fetching keys", nil)
```
golangcibot: printf: Infof call has arguments but no formatting directives (from `govet`)

pkg/auth/keymanager.go
```diff
@@ -140,11 +139,11 @@ func (km *KeyManager) fetchKeys(keysEndpointURL string) ([]*PublicKey, error) {
 	defer func() {
 		_, err := ioutil.ReadAll(res.Body)
 		if err != io.EOF && err != nil {
-			km.logger.Println("failed read remaining data before closing response")
+			logr.Errorf(nil, err, "failed read remaining data before closing response", nil)
```
golangcibot: printf: Errorf call has arguments but no formatting directives (from `govet`)

pkg/auth/keymanager.go
```diff
@@ -140,11 +139,11 @@ func (km *KeyManager) fetchKeys(keysEndpointURL string) ([]*PublicKey, error) {
...
 		}
 		err = res.Body.Close()
 		if err != nil {
-			km.logger.Println("failed to close response after reading")
+			logr.Errorf(nil, err, "failed to close response after reading", nil)
```
golangcibot: printf: Errorf call has arguments but no formatting directives (from `govet`)

pkg/auth/keymanager.go
```diff
@@ -156,11 +155,8 @@ func (km *KeyManager) fetchKeys(keysEndpointURL string) ([]*PublicKey, error) {
 	bodyString := buf.String()
 	// if status code was not OK, bail out
 	if res.StatusCode != http.StatusOK {
-		km.logger.Println(map[string]interface{}{
-			"response_status": res.Status,
-			"response_body":   bodyString,
-			"url":             keysEndpointURL,
-		}, "unable to obtain public keys from remote service")
+		err := errors.New("unable to obtain public keys from remote service")
+		logr.Errorf(nil, err, "response_status: %s", "response_body: %s", "url: %s", res.Status, bodyString, keysEndpointURL)
```
golangcibot: printf: Errorf call needs 1 arg but has 5 args (from `govet`)

pkg/auth/tokenparser.go
tinakurian: fixed here: f056abe3a4ff3aa5909f9272b03bb4bb63d732b0

pkg/log/log.go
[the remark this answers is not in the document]
tinakurian: fixed here: f1dfbdc0e3f6ec6ffb4d94d71281a675eb84d12f

pkg/log/log.go
```diff
@@ -0,0 +1,217 @@
...
+			} else {
+				headers[k] = v
+			}
+		}
+		fields = append(fields, "req_headers")
+		fields = append(fields, headers)
+	}
+	if req.ContentLength > 0 {
+		buf := new(bytes.Buffer)
+		buf.ReadFrom(req.Body)
```
golangcibot: Error return value of `buf.ReadFrom` is not checked (from `errcheck`)

pkg/log/log.go
```diff
@@ -0,0 +1,222 @@
...
+		p.lgr.Error(err, fmt.Sprintf(msg, arguments...), ctxInfo...)
+	} else {
+		p.lgr.Error(err, msg, ctxInfo...)
+	}
+
+	return p
+}
+
+// WithValues appends tags to the logger.
+func (p *Logger) WithValues(keysAndValues ...interface{}) *Logger {
```
alexeykazakov: As agreed we should use maps. Not slices.

cmd/main.go
```diff
@@ -3,18 +3,23 @@ package main
...
 	"github.com/codeready-toolchain/registration-service/pkg/configuration"
+	"github.com/codeready-toolchain/registration-service/pkg/log"
 	"github.com/codeready-toolchain/registration-service/pkg/server"
 )
 
+var logr log.Logger
+
 func main() {
+	// create logrger and registry
+	logr = *log.InitializeLogger("logger")
```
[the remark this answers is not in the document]
tinakurian: fixed here: b2848c95cac2d3558bb6653e7d0d7143bb5710b5

pkg/log/log.go
```diff
@@ -0,0 +1,222 @@
...
+	if len(keysAndValues) > 0 {
+		tags := append([]interface{}(nil), p.tags...)
+		tags = append(tags, keysAndValues...)
+		p.tags = tags
```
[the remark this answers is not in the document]
tinakurian: fixed here: 20494797a26c27aa3620222c8c9fa53a86f70b12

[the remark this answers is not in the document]
alexeykazakov: I don't see it's fixed in https://github.com/codeready-toolchain/registration-service/commit/20494797a26c27aa3620222c8c9fa53a86f70b12

pkg/log/log.go
tinakurian: fixed: ab763014216a10160ac49c4b79f34ff2e6a77b1e

pkg/log/log.go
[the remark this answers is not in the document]
tinakurian: sorry fixed here: ab763014216a10160ac49c4b79f34ff2e6a77b1e

cmd/main.go
```diff
@@ -68,14 +73,14 @@ func gracefulShutdown(hs *http.Server, logger *log.Logger, timeout time.Duration
...
-	logger.Printf("Signal received: %+v", sigReceived)
+	logr.Infof(nil, "Signal received: %+v", sigReceived.String())
 
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
-	logger.Printf("\nShutdown with timeout: %s\n", timeout)
+	logr.Infof(nil, "\nShutdown with timeout: %s\n", timeout.String())
 	if err := hs.Shutdown(ctx); err != nil {
-		logger.Printf("Shutdown error: %v\n", err)
+		logr.Errorf(nil, err, "Shutdown error: %s\n", err.Error())
```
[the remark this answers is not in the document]
tinakurian: fixed

pkg/log/log.go
```diff
@@ -0,0 +1,235 @@
...
+
+	// Use a zap logr.Logger implementation. If none of the zap
+	// flags are configured (or if the zap flag set is not being
+	// used), this defaults to a production zap logger.
+	//
+	// The logger instantiated here can be changed to any logger
+	// implementing the logr.Logger interface. This logger will
+	// be propagated through the whole operator, generating
+	// uniform and structured logs.
+	logf.SetLogger(zap.Logger())
```
alexeykazakov: Why not moving everything to `once.Do(func() {...})`?

pkg/log/log.go
```diff
@@ -0,0 +1,235 @@
...
+
+// WithValues creates a new logger with additional key-value pairs in the context
+func WithValues(keysAndValues map[string]interface{}) *Logger {
+	return logger.WithValues(keysAndValues)
+}
+
+// Info logs are used for non-error messages. It will log a message with
+// the given key/value pairs as context.
+func (l *Logger) Info(ctx *gin.Context, msg string) {
+	logger.Infof(ctx, msg)
```
alexeykazakov:
```suggestion
	l.logr.Infof(ctx, msg)
```

alexeykazakov:
> sorry fixed here: ab76301, 86e2e9c

It's still not fixed.

pkg/log/log.go
```diff
@@ -0,0 +1,235 @@
...
+		l.logr.Error(err, msg, ctxInfo...)
+	}
+}
+
+// WithValues appends tags to the logger.
+func (l *Logger) WithValues(keysAndValues map[string]interface{}) *Logger {
+	if len(keysAndValues) > 0 {
+		// ZapLoggerTo, name and WithValues all return new logger instances.
+		// The logger must be set again with the values stored in the Logger struct.
+		logger.logr = logger.logr.WithValues(slice(keysAndValues)...)
```
alexeykazakov: It's the same error. We can't modify the existing logger! Create a new one and return it. Don't modify `logger`. See https://github.com/tinakurian/registration-service/pull/5/files#diff-1135f265d25e96e557b8114567653176R119
Otherwise there will be race condition because multiple go routines will be constantly adding their values to the singleton logger.

pkg/log/log.go
tinakurian: Done!

pkg/log/log.go
tinakurian: Done!

pkg/log/log.go
tinakurian: fixed here: 352ae2c630cbef193aa3fc12308e5d048dfe423d

tinakurian:
> @tinakurian I did some changes:
> 
> * fixed a few bugs
> * refactored tests a bit
> * added more tests
> * added commit to logs
> * minor changes in other places to use logs properly

Thanks @alexeykazakov - thanks for making the change in keymanager to use withvalues - looks much better!

PR #40, opened 2019-10-03
Move image push to quay.io

alexeykazakov:
Until now travis pushed the image build from master to @michaelkleinhenz's personal repo on docker.io
This PR switches to codeready-toolchain org on quay.io

It also splits image build make targets to `make dev-image` (contains build time tag and used for minishift/dev deployments) and `make image` (for publishing on quay.io).


.travis.yml
```diff
@@ -26,16 +26,16 @@ script:
   - "make depend"
   - "make build"
   - "make test"
-  - "make image"
   - "make check-go-format"
```
xcoulon: side note: why do we check the format after building? wouldn't it be faster to do it before instead?

scripts/docker_push.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+GIT_COMMIT_ID_SHORT=`git rev-parse --short HEAD`
+if [[ -n `git status --porcelain --untracked-files=no` ]]; then
+    # uncommitted changes - dirty
+    GIT_COMMIT_ID_SHORT="$GIT_COMMIT_ID_SHORT-dirty"
+fi
+
+IMAGE_NAME="quay.io/codeready-toolchain/registration-service:$GIT_COMMIT_ID_SHORT"
```
xcoulon: what about using `quay.io/${GO_PACKAGE_ORG_NAME}/${GO_PACKAGE_REPO_NAME}` to be consistent with what's in `make/dev.mk` (https://github.com/codeready-toolchain/registration-service/pull/40/files#diff-fd8f90289cf39bfe347fc71f872e5861R72)?

scripts/docker_push.sh
```diff
@@ -0,0 +1,15 @@
...
+LATEST="quay.io/codeready-toolchain/registration-service:latest"
+
+echo "$DOCKER_PASSWORD" | docker login quay.io -u "$DOCKER_USERNAME" --password-stdin
```
xcoulon: just curious, how are these `DOCKER_USERNAME` and `DOCKER_PASSWORD` env vars set in the job?

.travis.yml
alexeykazakov: Good point. Previously we checked format after building an image (!) so I moved it one step up but yeah.. checking format should be our first think to do :)

scripts/docker_push.sh
alexeykazakov: Because I was too lazy to calculate org and repo names in docker_push.sh (we can't reuse our make here). Also docker_push.sh is purely for travis, so, I think it's OK to keep it simpler for now.

scripts/docker_push.sh
alexeykazakov: It's set manually in travis configuration for this repo - https://travis-ci.org/codeready-toolchain/registration-service/settings

.travis.yml
alexeykazakov: Done in 6b78ede71dfbcb3a9769946259b2998045888490

scripts/docker_push.sh
xcoulon: ok, fair point

scripts/docker_push.sh
xcoulon: ok, cool. I knew about the encoded env vars in the `travis.yaml` file itself (see https://docs.travis-ci.com/user/encryption-keys/#usage), but I forgot that it's also doable in the console/UI

alexeykazakov:
Thanks for the review @xcoulon.

PR #41, opened 2019-10-07
User Provisioning State Endpoint

michaelkleinhenz:
This adds a simple endpoint for polling the provisioning state. Note that this does only add the base endpoint infra, *not* the integration with the K8S api to actually retrieve the provisioning state.


pkg/server/routes.go
```diff
@@ -81,6 +82,7 @@ func (srv *RegistrationServer) SetupRoutes() error {
 		securedV1 := srv.router.Group("/api/v1")
 		securedV1.Use(authMiddleware.HandlerFunc())
 		securedV1.POST("/signup", signupCtrl.PostHandler)
+		securedV1.GET("/signupcheck", signupCheckCtrl.GetHandler)
```
alexeykazakov:
```suggestion
		securedV1.GET("/signupcheck", signupCtrl.StatusGetHandler)
```

Let's add another handler to the existing signup controller instead of introducing a new one: `/api/v1/signup/status`.
The payload it returns also could be a bit more detailed:

```
type SignupStatus struct {
    Ready bool
    Reason string
    Message *string
}
```

`Ready == true` only when the account is ready to be used by the user.
`Reason` can be: `pendingApproval`, `provisioning`, 'provisioned'
`Message` is optional. It's a human readable message if there is some problem with the account/signup.

Basically it's a simplified version of Signup/MUR status/conditions sub resource we use in Kube.

pkg/controller/signup_check.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// GetHandler returns a default heath check result.
+func (hc *SignupCheck) GetHandler(ctx *gin.Context) {
+	// Default handler for system SignupCheck
+	ctx.Writer.Header().Set("Content-Type", "application/json")
+	SignupCheckInfo := hc.getSignupCheckInfo()
+	if hc.config.IsTestingMode() {
+		// testing mode, wait 5s from the first request to flag state complete.
+		if hc.testRequestTimestamp == 0 {
```
alexeykazakov: I'm not a fan of adding too many test specifics to the production code. Let's minimize it as much as possible. It makes code not very clear and also it's error prone. There are many ways to avoid it here and still be able to test it. You could mock `getSignupCheckInfo()` in the test for example.

pkg/server/routes.go
michaelkleinhenz: We could also use the same endpoint `/api/v1/signup` but with `GET`. I'd say this would even better match the pattern. WDYT?

michaelkleinhenz: Added the above extra data in 7d2cd53.

pkg/controller/signup_check.go
michaelkleinhenz: Made that a bit more ligthweight in 7f970f2. I guess this will change anyway in the later PR where we integrate everything together and actually implementing the reporting.

pkg/controller/signup_check.go
```diff
@@ -0,0 +1,97 @@
...
+		logger: logger,
+		config: config,
+	}
+}
+
+var testRequestTimestamp int64
+
+// getTestSignupCheckInfo retrieves a test check info. Used only for tests.
+// It reports provisioning/not ready for 5s, then reports state complete.
+func (hc *SignupCheck) getTestSignupCheckInfo() *SignupCheckPayload {
```
xcoulon: what about passing this func as an argument in the `NewSignupCheck`? In that case, this will move to the `_test.go` file and not be part of the "production" code at all. And you'll also be able to get rid of the `if hc.config.IsTestingMode()` condition in  https://github.com/codeready-toolchain/registration-service/pull/41/files#diff-05008956dde833291afa7105f4e18c3aR83-R87

Something like:

```
func NewSignupCheck(logger *log.Logger, config *configuration.Registry, checker func() *SignupCheckPayload) *SignupCheck {
...
}
```

michaelkleinhenz: Cool idea. Result in 522d208.

pkg/server/routes.go
tinakurian: +1 it should be `GET /api/v1/signup `

alexeykazakov: @michaelkleinhenz it was the initial idea to use `GET /api/v1/signup` to get the signup with status. But we could move status to a subresource `GET /api/v1/signup/status` if we want to minimize the payload. I'm fine with both options.

pkg/controller/signup_check.go
```diff
@@ -0,0 +1,76 @@
...
+	// check the provisioning state.
+	return &SignupCheckPayload{
+		Ready:   true,
+		Reason:  "",
+		Message: "",
+	}
+}
+
+// GetHandler returns a default heath check result.
+func (hc *SignupCheck) GetHandler(ctx *gin.Context) {
```
alexeykazakov: Please move this handler to the existing Signup controller. No need to create a new one.

michaelkleinhenz: Fixed 8677464.

michaelkleinhenz: ...and 7f51434.

michaelkleinhenz:
Ok, I updated the tests. I am not sure if it makes sense to actually integrate the service with the endpoint in *this* PR as this has an impact on @sbryzak's PR. This should have *only* added the endpoint, not the whole shebang. @sbryzak wdyt?

michaelkleinhenz:
@alexeykazakov I am starting to think that @sbryzak's PR is actually including the stuff in here. This PR might be obsolete. After all, it was intended to not do much, just mount the endpoint and provide some tests for the envelope layer.

pkg/signup/signup_service.go
```diff
@@ -68,15 +111,33 @@ func (c *SignupServiceImpl) CreateUserSignup(username, userID string) (*crtapi.U
...
 		return nil, err
 	}
 
 	return created, nil
 }
 
-func (c *SignupServiceImpl) transformAndValidateUserName(username string) (string, error) {
+// getUserSignupImpl gets the UserSignup resource with the specified userID
+// Returns nil, nil if the resource is not found
+func (s *SignupServiceImpl) getUserSignupImpl(userID string) (*Signup, error) {
```
golangcibot: func `(*SignupServiceImpl).getUserSignupImpl` is unused (from `unused`)

michaelkleinhenz:
Ok, I looked at @sbryzak's PR and it is only dealing with the create resource, so let's move this PR from a "only endpoint" to the actual integration. I re-added the mock tests. @sbryzak we need to fill in `signup_service.go::getUserSignupImpl()` to complete this.

michaelkleinhenz:
Note: the latest commit (0005f8a) adds some integration in `signup_service.go::getUserSignupImpl()`, but this is all based on my blackbox exploration of the API. This needs more love.

pkg/signup/signup_service.go
```diff
@@ -8,30 +8,75 @@ import (
...
+	MasterUserRecords kubeclient.MasterUserRecordInterface
+	checkerFunc       func(userID string) (*Signup, error)
 }
 
-// NewSignupService creates a service object for performing user signup-related activities
-func NewSignupService(cfg SignupServiceConfiguration) (SignupService, error) {
+// NewSignupService creates a service object for performing user signup-related activities.
+func NewSignupService(cfg ServiceConfiguration) (Service, error) {
+
+	if cfg.IsTestingMode() {
```
alexeykazakov: You don't need any checkerFunc or test code here at all. I committed a small change which replace the client by the MUR interface: https://github.com/codeready-toolchain/registration-service/pull/41/commits/546d71df03959d41c9973d984ee4164afe10074a

All you need in your test is:
```
svc, err := NewSignupService(cfg)
svc.UserSignups = mockUserSignupsImpl
svc.MasterUserRecords = mockMURImpl
```

pkg/signup/signup_service.go
```diff
@@ -68,15 +117,54 @@ func (c *SignupServiceImpl) CreateUserSignup(username, userID string) (*crtapi.U
...
+// getUserSignupImpl gets the UserSignup resource with the specified userID
+// Returns nil, nil if the resource is not found
+func (s *ServiceImpl) getUserSignupImpl(userID string) (*Signup, error) {
+	// get signup resource
+	userSignup, err := s.UserSignups.Get(userID)
+	if err != nil {
+		return nil, err
+	}
+	// get MUR for it
+	mur, err := s.MasterUserRecords.Get(userSignup.GetName())
```
alexeykazakov: It will be a bit more complex than that. userSignup.Name will hold userID. And MUR.Name will hold the transformed username.
So, if:
```
original RHD IdentityProvider (IDP) username: jsmith@gmail.com
transformed Kube username: jsmith_at_gmail.com
RHD userID: 12345
```
Then:
```
UserSignup.Name: 12345
UserSignup.Spec.Username: jsmith_at_gmail.com
UserSignup.Meta.Annotations["idp_username"]: jsmith@gmail.com

MUR.Name: jsmith_at_gmail.com
MUR.Spec.UserID: 12345
MUR.Meta.Annotations["idp_username"]: jsmith@gmail.com
```

michaelkleinhenz: @sbryzak can you take a look here and fix this?

pkg/signup/signup_service.go
michaelkleinhenz: I feel that that this is all a bit overengineered. What I want to do here is to seperate the test and logic used for the controller from the logic in the service (or anything in the kube layer). This will seperate the tests as well. Currently, the test for the controller focusses solely on the controller code test, mocking the whole kube layer. 

You are suggesting to "install" the mock a layer further down in the kube layer, but that makes the testing of the controller code much more complex as it has to deal with the details of how the result from the kube api comes together. 

I don't think this is good. Yes, it will remove a few "extra turns" done for the test (like the checkerFunc mechanism), but it diverts the controller test and makes this very complex. Plus, the timed test is no longer possible without a lot of extra work.

michaelkleinhenz: And this also makes the middleware test much more complex as this one now also needs to mock that complete k8s client shebang. The signupService is added in the routes setup, so it is part of all tests that run through the routes setup (like the middleware test).

michaelkleinhenz:
@sbryzak we need help here. I worked on the mocks, but that all depends on the k8s layer to work. Here's the stuff that needs love:

1. `signup_service.go::GetUserSignup()` - my impl there is bs I guess (see comment from @alexeykazakov above). This needs to be fixed with prio 1 as everything below will fall into places then.
2. `signup_service_test.go::TestGetUserSignup()` - this is the test for the above. I added a skeleon here.
3. `controller/signup_test.go::TestSignupGetHandler()` has issues as this needs a working `GetUserSignup()`. Currently, it returns "not found" for all userIDs I feed into it. The test should be reasonable, but it does not work because of that. Note: the test is currently commented out.
4.  `middleware_test.go` has issues as this also needs a working k8s api mock now (the service is added in the routes setup that the middleware code-under-test runs through). It currently breaks with `unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined`.
We can work on this together, let me know if we should do a BJN.

alexeykazakov:
@michaelkleinhenz see my PR with tests for the controller - https://github.com/michaelkleinhenz/registration-service/pull/2
I'm pretty sure we can fix all other tests too.

michaelkleinhenz:
@alexeykazakov fyi: I talked to @sbryzak this morning, he said he will take care of the rest here  (I guess he means that we're making it worse right now ;-). So I keep my hands off this until Monday. Let's regroup then and see where we are.

alexeykazakov:
@michaelkleinhenz my last changes definitely is not making it worse:) I have talked with Shane too and he asked me to push my commits to your PR so he can now take it over.

michaelkleinhenz:
@alexeykazakov I fixed the tests after talking to @sbryzak this morning. Let's seee if we can get this merged asap.

pkg/signup/signup_service.go
```diff
@@ -4,34 +4,74 @@ import (
...
+// NewSignupService creates a service object for performing user signup-related activities.
+func NewSignupService(cfg ServiceConfiguration) (Service, error) {
+
+	if cfg.IsTestingMode() {
+		// testing mode, return default impl instance. This is needed
+		// for server and middleware tests where we need a full server
+		// initialization. In those cases, the mocking used in the
+		// signup controller tests can not be used as the initialization
+		// is happening before the test can hook into it.
+		return &ServiceImpl{}, nil
```
alexeykazakov: I would move this to `SetupRoutes()` and keep the default initialization clean.

pkg/signup/signup_service.go
```diff
@@ -68,15 +110,63 @@ func (c *SignupServiceImpl) CreateUserSignup(username, userID string) (*crtapi.U
...
+
+	// Check UserSignup status to determine whether user signup is complete
+	completed, value := condition.FindConditionByType(userSignup.Status.Conditions, crtapi.UserSignupComplete)
+	if value {
+		// If UserSignup status is complete, retrieve MasterUserRecord resource from the host cluster
+		mur, err := s.MasterUserRecords.Get(userSignup.GetName())
+		if err != nil {
+			return nil, err
+		}
+		if len(mur.Status.UserAccounts) != 1 {
```
alexeykazakov: We just need to use MUR Ready status condition here.

pkg/controller/signup.go
```diff
@@ -1,28 +1,59 @@
...
+	"github.com/codeready-toolchain/registration-service/pkg/signup"
 
 	"github.com/gin-gonic/gin"
 )
 
 // Signup implements the signup endpoint, which is invoked for new user registrations.
 type Signup struct {
-	config *configuration.Registry
-	logger *log.Logger
+	config        *configuration.Registry
```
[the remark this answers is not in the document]
alexeykazakov: Done in 6d3c4248159c28627e8a94ccbb0f9208ad4e4f4a

pkg/controller/signup.go
```diff
@@ -1,28 +1,59 @@
...
+func (s *Signup) GetHandler(ctx *gin.Context) {
+	ctx.Writer.Header().Set("Content-Type", "application/json")
+	// Get the UserSignup resource from the service by the userID
+	signupResource, err := s.signupService.GetUserSignup(ctx.GetString(middleware.SubKey))
+	if err != nil {
+		s.logger.Println("error getting UserSignup resource", err.Error())
+		errors.AbortWithError(ctx, http.StatusInternalServerError, err, "error getting UserSignup resource")
+	}
+	if signupResource == nil {
+		ctx.AbortWithStatus(http.StatusNotFound)
```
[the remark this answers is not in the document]
alexeykazakov: Done in d371452df914a7c95101fe5f84366442c0a66a19

pkg/controller/signup.go
```diff
@@ -1,28 +1,59 @@
...
+// PostHandler creates a Signup resource
 func (s *Signup) PostHandler(ctx *gin.Context) {
 	ctx.Writer.Header().Set("Content-Type", "application/json")
+	// TODO call s.signupService.CreateUserSignup() to create the actual resource in Kube API Server
+	ctx.Writer.WriteHeader(http.StatusOK)
+}
+
+// GetHandler returns the Signup resource
+func (s *Signup) GetHandler(ctx *gin.Context) {
+	ctx.Writer.Header().Set("Content-Type", "application/json")
```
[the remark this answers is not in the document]
alexeykazakov: Done in 71e0f3e08239cb73fb85cb47c1a49cf42d96ed20

pkg/controller/signup.go
```diff
@@ -1,28 +1,59 @@
...
+	}
+	if signupResource == nil {
+		ctx.AbortWithStatus(http.StatusNotFound)
+	} else {
+		ctx.Writer.WriteHeader(http.StatusOK)
+		err := json.NewEncoder(ctx.Writer).Encode(signupResource)
+		if err != nil {
+			s.logger.Println("error writing response body", err.Error())
+			errors.AbortWithError(ctx, http.StatusInternalServerError, err, "error writing response body")
+		}
```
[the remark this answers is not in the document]
alexeykazakov: Done in 71e0f3e08239cb73fb85cb47c1a49cf42d96ed20

pkg/server/server_test.go
```diff
@@ -19,6 +21,9 @@ func TestRunServerSuite(t *testing.T) {
 }
 
 func (s *TestServerSuite) TestServer() {
+	// Setting testing mode.
+	os.Setenv(configuration.EnvPrefix+"_"+"TESTINGMODE", "true")
```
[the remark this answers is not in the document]
alexeykazakov: Done in 45d1805104b360106d3dcc6655f59b5b50a01b88

pkg/server/server_test.go
```diff
@@ -28,10 +33,6 @@ func (s *TestServerSuite) TestServer() {
 	err = srv.SetupRoutes()
 	require.NoError(s.T(), err)
 
-	// Check that there are routes registered.
-	routes := srv.GetRegisteredRoutes()
-	require.NotEmpty(s.T(), routes)
```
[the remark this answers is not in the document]
alexeykazakov: Done in 94173e1a60ea72df3af18f8e5b7c526c7f4ddcde

pkg/signup/signup_service.go
alexeykazakov: Done in 45d1805104b360106d3dcc6655f59b5b50a01b88

pkg/signup/signup_service.go
alexeykazakov: Done in 59ca69bcd7a5246755ae9f2bd9c45447747aba79

pkg/middleware/middleware_test.go
```diff
@@ -91,6 +92,8 @@ func (s *TestAuthMiddlewareSuite) TestAuthMiddlewareService() {
 	err = srv.SetupRoutes()
 	require.NoError(s.T(), err)
 
+	fmt.Println(srv.Engine().Routes())
```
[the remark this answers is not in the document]
alexeykazakov: Done in d371452df914a7c95101fe5f84366442c0a66a19

alexeykazakov:
@michaelkleinhenz @sbryzak after my last refactoring (see above) the only thing is left here is missing tests. I'm pretty happy with the functionality itself now. But let's add the missing tests before merging it.

pkg/signup/signup_service.go
```diff
@@ -126,8 +128,8 @@ func (s *ServiceImpl) GetSignup(userID string) (*Signup, error) {
 	}
 
 	// Check UserSignup status to determine whether user signup is complete
-	signupCondition, complete := condition.FindConditionByType(userSignup.Status.Conditions, crtapi.UserSignupComplete)
-	if !complete {
+	signupCondition, found := condition.FindConditionByType(userSignup.Status.Conditions, crtapi.UserSignupComplete)
+	if !found || (found && signupCondition.Status != apiv1.ConditionTrue) {
```
alexeykazakov: I think something like `condition.IsTrue()` would be helpful here. I've created a PR for common - https://github.com/codeready-toolchain/toolchain-common/pull/45

sbryzak: I just refactored this code a little because I also have to handle the scenario where the condition does not even exist.  I've added a test for this also.  I'm not sure a `condition.IsTrue()` would be helpful in this case.

alexeykazakov: > I'm not sure a condition.IsTrue() would be helpful in this case.

Why not?
https://github.com/codeready-toolchain/toolchain-common/pull/45/files#diff-c3466dd19dde2ab03d3be49a50141b77R162

alexeykazakov: Just for clarification. You change is correct. And it's not about the case when the condition doesn't exist only. The original code was incorrect. It treated a found condition (even if it's set to False!) as True condition. It was a bug.
So, basically my PR just moves this functionality to common to `condition.IsTrue()` helper function.

alexeykazakov: So, when the PR is merged you could use it:
```
if condition.IsTrue(userSignup.Status.Conditions, crtapi.UserSignupComplete) {
    ...
}
```

sbryzak: You're right, the original code was incorrect which is why I changed it.  I still can't use `condition.IsTrue()` because I also need to know whether the actual condition exists or not.  If the condition exists, I populate the `Signup` object with the `Reason` and `Message` values from the condition, otherwise I set them to hard-coded values.

alexeykazakov: Ah, right. You need the condition itself to get its message and reason.

pkg/signup/signup_service.go
```diff
@@ -17,6 +17,11 @@ import (
 	"k8s.io/client-go/rest"
 )
 
+const (
+	SignupReasonNoCondition  = "UserSignupComplete condition not set"
```
alexeykazakov: Condition Reason is supposed to be one word. Something like `userSignCompleteCondtionNotSet` but in practice the missing condition means that the UserSignup resource is not handled by the controller yet. It basically equals `PendingApproval`.

pkg/signup/signup_service.go
```diff
@@ -129,15 +134,22 @@ func (s *ServiceImpl) GetSignup(userID string) (*Signup, error) {
...
 	// Check UserSignup status to determine whether user signup is complete
 	signupCondition, found := condition.FindConditionByType(userSignup.Status.Conditions, crtapi.UserSignupComplete)
-	if !found || (found && signupCondition.Status != apiv1.ConditionTrue) {
+	if !found {
+		signupResponse.Status = Status{
+			Reason:  SignupReasonNoCondition,
+			Message: SignupMessageNoCondition,
+		}
+		return signupResponse, nil
+	} else if found && signupCondition.Status != apiv1.ConditionTrue {
```
alexeykazakov: `found` is always `true` here.

pkg/signup/signup_service_test.go
```diff
@@ -215,22 +215,116 @@ func (s *TestSignupServiceSuite) TestGetSignupStatusNotComplete() {
...
+	})
+	require.NoError(s.T(), err)
+
+	response, err := svc.GetSignup(userID.String())
+	require.NoError(s.T(), err)
+	require.NotNil(s.T(), response)
+
+	require.Equal(s.T(), "bill", response.Username)
+	require.False(s.T(), response.Status.Ready)
+	require.Equal(s.T(), response.Status.Reason, signup.SignupReasonNoCondition)
```
alexeykazakov: should be `(s.T(), expected, actual)`

pkg/signup/signup_service_test.go
```diff
@@ -215,22 +215,116 @@ func (s *TestSignupServiceSuite) TestGetSignupStatusNotComplete() {
...
+	require.Equal(s.T(), response.Status.Message, signup.SignupMessageNoCondition)
```
alexeykazakov: should be `(s.T(), expected, actual)`

pkg/signup/signup_service.go
sbryzak: Good catch, fixed.

pkg/signup/signup_service_test.go
sbryzak: You're right, fixed.

pkg/signup/signup_service_test.go
sbryzak: Fixed

pkg/signup/signup_service.go
sbryzak: Changed to `PendingApproval`

pkg/controller/signup.go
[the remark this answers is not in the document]
sbryzak: Looks like this has been added.

PR #44, opened 2019-10-11
Cleanup JSON response encoding

alexeykazakov:
- Fixed `error.EncodeError()` (and also renamed it to `AbortWithError()`). It didn't properly write the response before.
- Cleaned up `HealthCheck` controller & fixed the 503 response in case of failed health check.
- Added more tests for `HealthCheck` controller

### DIFF
--- a/pkg/controller/health_check_test.go
+++ b/pkg/controller/health_check_test.go
@@ -34,11 +34,8 @@ func (s *TestHealthCheckSuite) TestHealthCheckHandler() {
 	// Create logger and registry.
 	logger := log.New(os.Stderr, "", 0)
 
-	// Check if the config is set to testing mode, so the handler may use this.
-	assert.True(s.T(), s.Config.IsTestingMode(), "testing mode not set correctly to true")
-
 	// Create health check instance.
-	healthCheckCtrl := controller.NewHealthCheck(logger, s.Config)
+	healthCheckCtrl := controller.NewHealthCheck(logger, s.Config, controller.NewHealthChecker(s.Config))
 	handler := gin.HandlerFunc(healthCheckCtrl.GetHandler)
 
 	s.Run("health in testing mode", func() {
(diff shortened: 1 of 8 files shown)


pkg/controller/health_check_test.go
```diff
@@ -76,80 +72,54 @@ func (s *TestHealthCheckSuite) TestHealthCheckHandler() {
...
+	assert.Equal(t, configuration.BuildTime, actual.BuildTime, "wrong build_time in health response")
+	assert.Equal(t, configuration.StartTime, actual.StartTime, "wrong start_time in health response")
+	assert.Equal(t, expectedTestingMode, actual.TestingMode, "wrong testing mode in health response")
+}
 
-		// Check the response body is what we expect.
-		data := &controller.Health{}
-		err := json.Unmarshal(rr.Body.Bytes(), &data)
-		require.NoError(s.T(), err)
+type mockHealthChecker struct {
```
xcoulon: what about using minimock to generate the mock impl since we have interfaces?

pkg/errors/error.go
```diff
@@ -14,19 +13,12 @@ type Error struct {
 	Details string `json:"details"`
 }
 
-// EncodeError encodes a json error response.
-func EncodeError(ctx *gin.Context, err error, code int, details string) {
-	// construct an error.
-	errorStruct := &Error{
+// AbortWithError stops the chain, writes the status code and the given error
```
xcoulon: since this code relies on github.com/gin-gonic/gin, shouldn't we move it in the `pkg/controller` pkg instead? it's not supposed to be used anywhere else, is it?

pkg/controller/health_check_test.go
alexeykazakov: To be honest I'm not a fun of minimock :) IMO sometimes it adds additional complexity instead of reducing it. It requires build generation.
Anyway in this particular case I think it would be overkill. The mock checker I'm providing here is super light.

pkg/errors/error.go
alexeykazakov: I think we will add more stuff to error package later. Maybe our own error types (for wrapping errors). Error checks, etc. So it will be used outside controllers too. So, I wouldn't move it to controller in this PR and leave it where it is already now.

alexeykazakov:
Thanks for the review @tinakurian

pkg/controller/health_check_test.go
xcoulon: ok, I understand your concern. I believe it becomes helpful when you need to alter the behaviour of the mock in the tests to verify different uses cases. In which case, having multiple implementations as you're doing here would be a less elegant approach. But we can keep it as-is for now.

pkg/errors/error.go
xcoulon: ok. I'm just concerned that this kind of error should only be used as the controller level.

pkg/controller/health_check_test.go
alexeykazakov: What multiple implications would we need here? Wouldn’t the current mock implementation cover all possible needs (as it does in these tests)?

xcoulon: ok, this mock is simple so we're fine. But I believe that minimock (for example) provides more flexibility if we need to alter the response to a method call to test so cases of errors, etc. Here, it's just a bool that is returned, and it is provided when initializing the mock, so it's good enough, indeed.

PR #46, opened 2019-10-17
Remove operator-sdk dependency

sbryzak:
Signed-off-by: Shane Bryzak <sbryzak@gmail.com>


pkg/log/log.go
```diff
@@ -55,9 +55,9 @@ func Init(withName string, out io.Writer) {
 		// be propagated through the whole operator, generating
 		// uniform and structured logs.
 		if out == nil {
-			logf.SetLogger(zap.Logger())
+			logf.SetLogger(zap.Logger(false))
```
xcoulon: the doc says that this `Logger(..)` func is now deprecated and suggest to use `New()` instead

pkg/log/log.go
```diff
@@ -55,9 +55,9 @@ func Init(withName string, out io.Writer) {
...
 		} else {
-			logf.SetLogger(zap.LoggerTo(out))
+			logf.SetLogger(zap.LoggerTo(out, false))
```
xcoulon: same here, with `New(WriteTo(writer))`

PR #50, opened 2019-10-22
Use dev keycloak by default

alexeykazakov:
Also fixes `make deploy-dev`


alexeykazakov:
> You might want to add the default values to the example config yaml as well..

Example config yaml is just an example, right? Why I would need to change it. I'm not adding any new configuration fields.

PR #55, opened 2019-10-25
Cleanup signup handler

alexeykazakov:


pkg/controller/signup.go
```diff
@@ -40,7 +36,8 @@ func (s *Signup) PostHandler(ctx *gin.Context) {
 	}
 
 	log.Infof(ctx, "UserSignup %s created", userSignup.Name)
-	ctx.Status(http.StatusOK)
+	ctx.Status(http.StatusNoContent)
```
tinakurian: I'm not quite sure that this is the right response code. 204 is really sort of a response code that shouldn't be used unless there is a good reason to do so. Probably most popularly used for DELETE. But returning this when we create a resource seems somewhat misleading - it's essentially saying we have nothing left to do, you may go back or exit or start again. If anything, we should return a 202 - Accepted. We should include the current status as well as a link to monitor the status as things change (this would be to our GET signup api)... what do you think?

alexeykazakov: IMO it's in gray area (as many http statuses :)
204 is supposed to be used when there is no content in the response. Which is exactly the case. But what concerns me is that 204 is cacheable by default. It can cause issues in some rear cases...
So, yeah, let's change it to 202. The main change here is the next line: `ctx.Writer.WriteHeaderNow()`. Without that the status is always OK no matter what.

PR #58, opened 2019-10-29
Fix transformUserName()

alexeykazakov:
This PR fixes `transformAndValidateUserName()` function which didn't take into account that `s.UserSignups.Get(transformed)` never returned nil so it caused endless loop.

Also added a limit to username transformation attempts. It will fail after 1000 attempts. Just to avoid endless loops in case of some regression bugs in the future ;) 

And this PR fixes RBAC for the namespace that the reg-service is deployed to.

### DIFF
--- a/pkg/signup/signup_service_test.go
+++ b/pkg/signup/signup_service_test.go
@@ -72,6 +72,7 @@ func (s *TestSignupServiceSuite) TestUserSignupTransform() {
 			return userSignupToBeReturnedByClient, errToBeReturnedByClient
 		}
 		userID, err := uuid.NewV4()
+		require.NoError(s.T(), err)
 
 		userSignup, err := svc.CreateUserSignup("jane.doe@redhat.com", userID.String())
 		require.NoError(s.T(), err)
@@ -104,6 +105,7 @@ func (s *TestSignupServiceSuite) TestUserSignupTransform() {
 	s.Run("unable to transform after N attempts", func() {
 		svc, userSignupsClient, _ := newSignupServiceWithFakeClient()
 		userID, err := uuid.NewV4()
+		require.NoError(s.T(), err)
 		userSignupsClient.MockGet = func(s string) (*v1alpha1.UserSignup, error) {
 			return &v1alpha1.UserSignup{}, nil // Always return some UserSignup
 		}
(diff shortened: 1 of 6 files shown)


pkg/signup/signup_service_test.go
```diff
@@ -62,29 +63,54 @@ func (s *TestSignupServiceSuite) TestCreateUserSignup() {
...
+	check := func(userSignupToBeReturnedByClient *v1alpha1.UserSignup, errToBeReturnedByClient error) {
+		svc, userSignupsClient, _ := newSignupServiceWithFakeClient()
 
-	userSignup, err := svc.CreateUserSignup("jane.doe@redhat.com", userID.String())
-	require.NoError(s.T(), err)
-	require.NotNil(s.T(), userSignup)
+		userSignupsClient.MockGet = func(s string) (*v1alpha1.UserSignup, error) {
+			return userSignupToBeReturnedByClient, errToBeReturnedByClient
+		}
+		userID, err := uuid.NewV4()
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

pkg/signup/signup_service_test.go
```diff
@@ -62,29 +63,54 @@ func (s *TestSignupServiceSuite) TestCreateUserSignup() {
...
+		check(nil, kubeerr.NewNotFound(v1alpha1.SchemeGroupVersion.WithResource(userID.String()).GroupResource(), userID.String()))
+	})
+
+	s.Run("UserSignup not found and client returns empty UserSignup", func() {
+		check(&v1alpha1.UserSignup{}, kubeerr.NewNotFound(v1alpha1.SchemeGroupVersion.WithResource(userID.String()).GroupResource(), userID.String()))
+	})
+
+	s.Run("unable to transform after N attempts", func() {
+		svc, userSignupsClient, _ := newSignupServiceWithFakeClient()
+		userID, err := uuid.NewV4()
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

pkg/signup/signup_service_test.go
alexeykazakov: Fixed in https://github.com/codeready-toolchain/registration-service/pull/58/commits/0ab9bd226f3657d946f0346295106959cde8f5a8

pkg/signup/signup_service_test.go
alexeykazakov: Fixed in https://github.com/codeready-toolchain/registration-service/pull/58/commits/0ab9bd226f3657d946f0346295106959cde8f5a8

PR #61, opened 2019-11-04
Registration service e2e tests

tinakurian:
Removing tokenmanager from registration service as it has been placed in toolchain-common. This is part of the changes required for e2e tests for  the registration-service. 

tokenmanager in toolchain-common: https://github.com/codeready-toolchain/toolchain-common/commit/e0b97f314fbbb41bc6246d994c0edbed740ff1d7


pkg/auth/tokenparser_test.go
```diff
@@ -8,67 +8,77 @@ import (
...
 
 	"github.com/dgrijalva/jwt-go"
 	uuid "github.com/satori/go.uuid"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
 )
 
-func TestTokenParser(t *testing.T) {
+type TestTokenParserSuite struct {
```
xcoulon: just curious, why do you want to use a `UnitTestSuite` here, instead of the standalone tests?

pkg/auth/tokenparser_test.go
```diff
@@ -8,67 +8,77 @@ import (
...
+func (s *TestTokenParserSuite) TestTokenParser() {
 	configRegistry := configuration.CreateEmptyRegistry()
 
 	// create test keys
-	tokengenerator := test.NewTokenManager()
+	tokengenerator := authsupport.NewTokenManager()
 	kid0 := uuid.NewV4().String()
 	_, err := tokengenerator.AddPrivateKey(kid0)
-	require.NoError(t, err)
+	require.NoError(s.T(), err)
```
xcoulon: actually, you don't have to use `s.T()` instead of `t` anymore, since https://github.com/stretchr/testify/commit/865fb2c8f5af9ccbaaadfd20d18236358dccaaff

pkg/auth/tokenparser_test.go
tinakurian: I was assigned a story awhile ago to change all the tests to use the suite instead. I guess this test slipped in afterwards. When i was fixing the tests, i thought to make it follow the convention we agreed on. 

The PR for test suite is here: https://github.com/codeready-toolchain/registration-service/commit/7bce17cec472137076dcdb2d09b53ca8683f8ac4

pkg/auth/tokenparser_test.go
tinakurian: Do you mind if we make this change in another PR? I can create one and assign to to myself - was can make the change for all the tests in one sweep in that case. Currently all the tests use s.T().

pkg/auth/tokenparser_test.go
xcoulon: ok, I see. To be honest, I'm not sure there's much value in using a `UnitTestSuite` here, but it's ok to have it, and it's probably more consistent with the other tests.

pkg/auth/tokenparser_test.go
xcoulon: ok, let's proceed like that, then

PR #63, opened 2019-11-05
E2e registration service

tinakurian:
Adding the e2etestingmode variable and using it in keymanager.

This will override fetching keys from the given url and use the keys hard-coded in the configuration file.


pkg/configuration/configuration.go
```diff
@@ -232,3 +243,8 @@ func (c *Registry) GetAuthClientPublicKeysURL() string {
...
...iEdR1zslqEOXaqWsp6HlUcu1JOuEv1LsxFuCnKc9ZvZDm0mQQJiOAl1QRvSU3pgX3IjuoefY6-6NQAYm1MQjOzWSnNkQTTEIWgIRu8QVgxko50pR3fTC7LWj6AQCv5GkW1r5zIv9OSzjoiN8A_UHASWh0Z6oy0eLeY775EhVfrg_KjYw","e":"AQAB","x5c":["MIICoTCCAYkCBgFtIFN1nTANBgkqhkiG9w0BAQsFADAUMRIwEAYDVQQDDAlkZW1vUmVhbG0wHhcNMTkwOTExMTIzNTAzWhcNMjkwOTExMTIzNjQzWjAUMRIwEAYDVQQDDAlkZW1vUmVhbG0wggEiMA0GCSqGSIb3DQEBAQUAA4IBDwAwggEKAoIBAQCLTjLFpBvt7X619wOhd7wrYNn6snbJVWMGhP1PDSM6i63nDzjHb+vm3/hoIFCOj0g7tRVZO28TRw6TVa3KRe+yjuJiH2vL2Oxm2sk1LRBfKRw4lkktXHKHW5q2tukzmF4pg3Pl+zmGn7dCHleTIKQuKAiQ6gsxc7W0uQ2+OIR1HXOyWoQ5dqpaynoeVRy7Uk64S/UuzEW4Kcpz1m9kObSZBAmI4CXVBG9JTemBfciO6h59jr7o1ABibUxCM7NZKc2RBNMQhaAhG7xBWDGSjnSlHd9MLstaPoBAK/kaRbWvnMi/05LOOiI3wD9QcBJaHRnqjLR4t5jvvkSFV+uD8qNjAgMBAAEwDQYJKoZIhvcNAQELBQADggEBADGGFneSXwWrT4Yk4PMcY14gfc6ta91Qz5xZDWiPz0ZaX1ULLEOu4k/rIKwN7tCMxBxCgnxaMj372JvAPAUqwLmRhvDxtcJDYHQwM77NqU3ZQARchqyDsd5aYW6cYFMF8D60PdFOgMRKJiRGpRbJgPt7+hFdbEw2XnWN9lnzXmbeXxCn7GZKZiKmWZU3eBa/pQVCjTb4JICs+1uBJj0VfgLNYHUbZXdvg4ismSEqXnBKX/V3lPJQWXU/yyMS6G9lHGcAisxWIthcA8C6gWUaJe1FwJrCeqDIJDABw72VAYvUaIf0pBVyXtr8A2JrBD9jdb8KOyC//X+LLiXqD1fpltw="],"x5t":"l_5tiA15SUVfBXx18njAbbs3wds","x5t#S256":"T8ef_9jOHIlDQVYCJOXPtyOkoRF-e8eJtyh7pswQclg"}]}`
```
alexeykazakov: Let's keep our test keys in https://github.com/codeready-toolchain/toolchain-common/tree/master/pkg/test/auth
We don't need them in reg-service configuration at all.
We basically need to add a hardcoded private key to https://github.com/codeready-toolchain/toolchain-common/tree/master/pkg/test/auth and a few functions to extract the public key and kid from it, generate a token (to be used in e2e tests), etc.
And then get that public key from common directly in https://github.com/codeready-toolchain/registration-service/pull/63/files#diff-5f5c78f921deb86823e059dbb7069b73R57 instead of using configuration.

Then if we need to update the key we have one place where it's stored and we wouldn't need to modify multiple repos.

pkg/configuration/configuration.go
```diff
@@ -66,6 +66,10 @@ const (
 	// DefaultTestingMode specifies whether the services should run in testing mode.
 	DefaultTestingMode = false
 
+	varE2ETestingMode = "e2etestingmode"
```
nurali-techie: all const string following diff format .. for ex, log.json, graceful_timeout, http.read_timeout
we should have same format for all of including this `e2etestingmode` .. this is also diff format where words are not separated with any delimiter.

pkg/configuration/configuration_test.go
```diff
@@ -313,6 +313,34 @@ func (s *TestConfigurationSuite) TestIsTestingMode() {
...
 }
 
+func (s *TestConfigurationSuite) TestIsE2ETestingMode() {
+	key := configuration.EnvPrefix + "_" + "E2ETESTINGMODE"
+	resetFunc := test.UnsetEnvVarAndRestore(key)
+	defer resetFunc()
+
+	s.Run("default", func() {
+		resetFunc := test.UnsetEnvVarAndRestore(key)
+		defer resetFunc()
```
nurali-techie: seems that this is redundant ?? even in next test case too.

pkg/configuration/configuration.go
tinakurian: @nurali-techie i didn't set this convention - just followed what was already in the file... (specifically was following "testingmode" in any case this will be removed shortly...

pkg/auth/keymanager_test.go
```diff
@@ -217,3 +217,33 @@ func (s *TestKeyManagerSuite) TestKeyFetching() {
 		}
 	})
 }
+
+func (s *TestKeyManagerSuite) TestE2EKeyFetching() {
+    s.Run("retrieve key for e2e-tests environment", func() {
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

pkg/configuration/configuration.go
tinakurian: this has been changed: e7b06a11435cad5ff5465e0a8b5c62f842d66d88

pkg/configuration/configuration.go
tinakurian: fixed: e7b06a11435cad5ff5465e0a8b5c62f842d66d88

pkg/configuration/configuration_test.go
tinakurian: this change is no longer part of this PR: e7b06a11435cad5ff5465e0a8b5c62f842d66d88

pkg/auth/keymanager_test.go
tinakurian: fixed here: f23f87da00e0825da65c2a25296de7c65a2510de

pkg/auth/keymanager_test.go
```diff
@@ -217,3 +217,33 @@ func (s *TestKeyManagerSuite) TestKeyFetching() {
...
+
+		for _, key := range keys {
+			// check if the keys are parsed correctly.
+			_, err = keyManager.Key(key.KeyID)
+			require.NoError(s.T(), err)
+		}
+	})
+
+	s.Run("fail to retrieve e2e keys for unit test environment", func() {
+		s.Config.GetViperInstance().Set("environment", "unit-tests")
```
alexeykazakov: We don't care much about unit-test env here.. Let's make sure e2e keys are not used in two usecases:
- Set env var explicitly to "prod", create a configuration and check the key manager.
- The env var is not set at all, so the default environment is used (which is currently "prod").

tinakurian: i had set it to unit-tests at random since this block specifically checks for e2e-tests environment. everything else (including prod) should fall into the else clause: https://github.com/codeready-toolchain/registration-service/pull/63/files#diff-5f5c78f921deb86823e059dbb7069b73R55 however, i don't mind changing it to prod.. the outcome should be no different though..

sure, i can add a test for default as well..

tinakurian: @alexeykazakov  also.. it looks like when you are running unit tests environment is already set to "unit-tests" see here: https://github.com/codeready-toolchain/registration-service/blob/master/test/testsuite.go#L24

tinakurian: changed unit-tests to prod here: bb4a4c252d6dc87253a67d997b6356e0c2dedb5d

pkg/auth/keymanager_test.go
```diff
@@ -217,3 +217,53 @@ func (s *TestKeyManagerSuite) TestKeyFetching() {
...
+		for _, key := range keys {
+			// check that key is not found as the environment
+			// is set to unit-tests.
+			_, err = keyManager.Key(key.KeyID)
+			require.Error(s.T(), err)
+			require.Equal(s.T(), err.Error(), "unknown kid")
+		}
+	})
+
+	s.Run("fail to retrieve e2e keys for unit test environment", func() {
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
	s.Run("fail to retrieve e2e keys for unit test environment", func() {
```

pkg/auth/keymanager_test.go
```diff
@@ -217,3 +217,53 @@ func (s *TestKeyManagerSuite) TestKeyFetching() {
...
+		for _, key := range keys {
+			// check that key is not found as the environment
+			// is set to unit-tests.
+			_, err = keyManager.Key(key.KeyID)
+			require.Error(s.T(), err)
+			require.Equal(s.T(), err.Error(), "unknown kid")
+		}
+	})
+
+	s.Run("fail to retrieve e2e keys for prod environment", func() {
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
	s.Run("fail to retrieve e2e keys for prod environment", func() {
```

pkg/auth/keymanager_test.go
tinakurian: fixed: 14263db118dfa67ec58f7c0eecc0b27b149e5ff8

pkg/auth/keymanager_test.go
tinakurian: fixed: 14263db118dfa67ec58f7c0eecc0b27b149e5ff8

pkg/auth/keymanager_test.go
tinakurian: fixed here: https://github.com/codeready-toolchain/registration-service/pull/63/files#diff-31b519707082aeb8272b75c7cf2460edR235

PR #67, opened 2019-11-08
redirect to https at the OpenShift route(r) level

xcoulon:
see docs on https://docs.openshift.com/container-platform/4.1/networking/routes/route-configuration.html#nw-enabling-hsts_route-configuration

### DIFF
--- a/deploy/deployment.yaml
+++ b/deploy/deployment.yaml
@@ -91,6 +91,8 @@ objects:
   - kind: Route
     apiVersion: v1
     metadata:
+      annotations:
+        haproxy.router.openshift.io/hsts_header: max-age=31536000
       labels:
         run: registration-service
       name: registration-service


xcoulon:
with this annotation in place, any request sent to `http://..` will be redirected to `https://...`

deploy/deployment.yaml
```diff
@@ -91,6 +91,8 @@ objects:
   - kind: Route
     apiVersion: v1
     metadata:
+      annotations:
+        haproxy.router.openshift.io/hsts_header: max-age=31536000
```
alexeykazakov: Hm... I don't see any difference with this annotation or without it. What am I doing wrong? :)
Even if I also add `insecureEdgeTerminationPolicy` I still don't see auto redirects if I type `http:` explicitly in my Chrome.
Testing it in minishift.

alexeykazakov: `insecureEdgeTerminationPolicy: Redirect` actually does the trick. And we already have it used in the route in toolchain-infra script for crt cluster: https://github.com/codeready-toolchain/toolchain-infra/blob/master/config/reg_service_route.yaml#L16

I think we don't have to use this redirect for the "internal" route we specify in deploy/deployment.yaml.

So, I'm closing it but please reopen if you think we still need to change something here.

xcoulon: ok, thanks @alexeykazakov

PR #68, opened 2019-11-08
removin userid

tinakurian:
removing userid to make registration service build using latest toolchain-api


pkg/signup/signup_service.go
```diff
@@ -97,7 +97,6 @@ func (s *ServiceImpl) CreateUserSignup(username, userID string) (*crtapi.UserSig
 			Namespace: s.Namespace,
 		},
 		Spec: crtapi.UserSignupSpec{
-			UserID:        userID,
```
alexeykazakov: We still need userID. But now it's used as UserSignup.matadata.name.

go.sum
```diff
@@ -1,8 +1,21 @@
+bitbucket.org/ww/goautoneg v0.0.0-20120707110453-75cd24fc2f2c/go.mod h1:1vhO7Mn/FZMgOgDVGLy5X1mE6rq1HbkBdkF/yj8zkcg=
```
alexeykazakov: Please do `go mod tidy`.

tinakurian: fixed here: ebd346dbee0e7459a4ac41f1332b5766178867b5

pkg/signup/signup_service.go
tinakurian: Hm.. is this part of the spec? i don't see it here: https://github.com/codeready-toolchain/api/blob/master/pkg/apis/toolchain/v1alpha1/usersignup_types.go#L19. 

Does this mean we will still be using it to create a usersignup?

alexeykazakov: It's not in Spec. It's resource name - https://kubernetes.io/docs/concepts/overview/working-with-objects/names/
It's in metadata - https://github.com/codeready-toolchain/api/blob/master/pkg/apis/toolchain/v1alpha1/usersignup_types.go#L74

PR #71, opened 2019-11-14
Return Web Console URL in GET /signup

alexeykazakov:
Fixes https://jira.coreos.com/browse/CRT-348

Requires - https://github.com/codeready-toolchain/api/pull/59

Related host change - https://github.com/codeready-toolchain/host-operator/pull/103

TODO:
- [x] More tests

### DIFF
--- a/pkg/signup/signup_service.go
+++ b/pkg/signup/signup_service.go
@@ -24,9 +24,8 @@ const (
 // Signup represents Signup resource which is a wrapper of K8s UserSignup
 // and the corresponding MasterUserRecord resources.
 type Signup struct {
-	// The cluster in which the user is provisioned in
-	// If not set then the target cluster will be picked automatically
-	TargetCluster string `json:"targetCluster,omitempty"`
+	// The Web Console URL of the cluster which the user was provisioned to
+	ConsoleURL string `json:"consoleURL,omitempty"`
 	// The username.  This may differ from the corresponding Identity Provider username, because of the the
 	// limited character set available for naming (see RFC1123) in K8s. If the username contains characters which are
 	// disqualified from the resource name, the username is transformed into an acceptable resource name instead.
@@ -127,8 +126,7 @@ func (s *ServiceImpl) GetSignup(userID string) (*Signup, error) {
 	}
 
 	signupResponse := &Signup{
-		Username:      userSignup.Spec.Username,
-		TargetCluster: userSignup.Spec.TargetCluster,
+		Username: userSignup.Spec.Username,
 	}
 
 	// Check UserSignup status to determine whether user signup is complete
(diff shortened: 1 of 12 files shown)


pkg/signup/signup_service.go
```diff
@@ -142,15 +146,19 @@ func (s *ServiceImpl) GetSignup(userID string) (*Signup, error) {
...
 		Ready:   ready,
 		Reason:  murCondition.Reason,
 		Message: murCondition.Message,
 	}
-	if mur.Spec.UserAccounts != nil && len(mur.Spec.UserAccounts) > 0 {
-		// TODO Set TargetCluster in UserSignup.Status. For now it's OK to get it from the first embedded UserAccount from MUR.
-		signupResponse.TargetCluster = mur.Spec.UserAccounts[0].TargetCluster
+	if mur.Status.UserAccounts != nil && len(mur.Status.UserAccounts) > 0 {
+		// TODO Set ConsoleURL in UserSignup.Status. For now it's OK to get it from the first embedded UserAccount status from MUR.
+		signupResponse.ConsoleURL = mur.Status.UserAccounts[0].Cluster.ConsoleURL
```
xcoulon: I would suggest that we "forge" for the `ConsoleURL` displayed in the UI here (ie, on the backend), rather than on the front-end (https://github.com/codeready-toolchain/registration-service/pull/71/files#diff-1c39f6c21128106bd25945f0d8391375R65). That would be easier to test. (Well, that's just my old habit of putting logic in the backend)

pkg/signup/signup_service_test.go
```diff
@@ -302,3 +286,27 @@ func newSignupServiceWithFakeClient() (signup.Service, *fake.FakeUserSignupClien
 		MasterUserRecords: fakeMURClient,
 	}, fakeClient, fakeMURClient
 }
+
+func (s *TestSignupServiceSuite) newUserSignupComplete() *v1alpha1.UserSignup {
```
xcoulon: I would rather have this function without a pointer receiver, so it can be used in other tests, and thus expose the following signature:
```
func newUserSignupComplete() (*v1alpha1.UserSignup, error)
```

pkg/signup/signup_service.go
alexeykazakov: Well. The backend provides the ConsoleURL without any modifications.
But frontend (atm) may want to adjust that URL to point user to a specific place in the console. Like specific namespace and specific view of the console (topology view for example). We will probably have more than one link in UI - dev console, che dashboard, etc.
It's not clear atm what we would like to show the user atm. So, I would just keep the base console URL in backend for now.
We might move some logic of calculating specific URL out of the console URL to backend later when we know what exactly we need.

pkg/signup/signup_service_test.go
alexeykazakov: There is no other tests in this package. And I doubt we would need to use it in other packages. But even if we did then we would need to export the function which would require as to move it to another package, etc...
I would keep it as is since it's simple and clean. Let's move it somewhere else when/if we really need it.

PR #73, opened 2019-11-23
Landing Page in Pure HTML/CSS

michaelkleinhenz:
This solves CRT-90. It uses Patternfly 4 to create a pure HTML/CSS landing page using our minimal JS client impl. It currently uses default PF4 stylings rather than RH brand stylings as I wanted to leave this to default.

It supports all states of the UX defined in https://docs.google.com/presentation/d/1tp4DC10uRtA4sqLK1NrcmHFAkiblB1RTHhW48FltmUU/edit with one execption: `inProvisioning` as this is not supported from the backend (and the provisioning takes only a short amount of time, so it may make no sense to support a seperate state for it).

It also updates and integrates the login state (login/signup button or username display in header), re-routing to keycloak etc.

It also supports retrieving the DevConsole URL and properly linking up the card in the dashboard.


cmd/main.go
```diff
@@ -2,13 +2,13 @@ package main
...
 import (
 	"context"
-	"flag"
 	"net/http"
 	"os"
 	"os/signal"
 	"syscall"
 	"time"
 
+	"github.com/spf13/pflag"
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion

```

alexeykazakov:
> with one execption: inProvisioning as this is not supported from the backend (and the provisioning takes only a short amount of time, so it may make no sense to support a seperate state for it).

Provisioning status is supported by backend. But yeah.. it’s pretty quick.

tinakurian:
Wooo hoo!! awesome work Michael :-)

michaelkleinhenz:
> Provisioning status is supported by backend. But yeah.. it’s pretty quick.

what would be the `reason` value for that? There is only this as a fixed reason in `signup_service.go`:
```
	PendingApprovalReason = "PendingApproval"
```
I guess other reasons are pulled directly from the CD response:
```
      signupResponse.Status = Status{
	    Reason: PendingApprovalReason,
      }
```
Where are the available reasons documented? @sbryzak

alexeykazakov:
@michaelkleinhenz yes. It’s coming from MUR status: https://github.com/codeready-toolchain/registration-service/blob/5589e928dbcf0d3be5f333b4251cb5f872124985/pkg/signup/signup_service.go#L156

And here is the reason you need: https://github.com/codeready-toolchain/host-operator/blob/8d2290acd943c10950462e1be185802e5950b98d/pkg/controller/masteruserrecord/masteruserrecord_controller.go#L37

BTW, @MatousJobanek , @xcoulon , @sbryzak I think it would make sense to move all our reason constants to api repo and expose them.
Also I see that we need to cleanup MUR conditions. We don’t really use anything but Ready.

michaelkleinhenz:
@abhi-kn how do you retain the keycloak session between reloads? I would expect this is all controlled by the keycloak client and settings. I would also expect this behaviour to be configurable in keycloak. Are you manually retaining the token somewhere?

michaelkleinhenz:
> @abhi-kn how do you retain the keycloak session between reloads? I would expect this is all controlled by the keycloak client and settings. I would also expect this behaviour to be configurable in keycloak. Are you manually retaining the token somewhere?

never mind, found it out, see 6ebcf50.

michaelkleinhenz:
> @michaelkleinhenz yes. It’s coming from MUR status:
> 
> https://github.com/codeready-toolchain/registration-service/blob/5589e928dbcf0d3be5f333b4251cb5f872124985/pkg/signup/signup_service.go#L156
> 
> And here is the reason you need: https://github.com/codeready-toolchain/host-operator/blob/8d2290acd943c10950462e1be185802e5950b98d/pkg/controller/masteruserrecord/masteruserrecord_controller.go#L37
> 
> BTW, @MatousJobanek , @xcoulon , @sbryzak I think it would make sense to move all our reason constants to api repo and expose them.
> Also I see that we need to cleanup MUR conditions. We don’t really use anything but Ready.

Done, added `Provisioning` state in 783e28f. I don't think this will ever display though :-)

I also default now to "Pending Approval", so if an unknown reason comes back with state 200 on the query response, it displays "Pending Approval".

PR #75, opened 2019-12-02
chore(CRT-361):  move SA and namespace name definition into deployment template

MatousJobanek:
* As the registration service needs to be created programmatically inside of the host-operator repo we need to move the SA, RBAC and the namespace name definition inside of the deployment template

* Also fixed one typo in one ConfigMap variable

* also added a target (and check) for ensuring that the deployment in host-operator repo is properly synced

Paired e2e PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/53

See these PRs for better context:
https://github.com/codeready-toolchain/host-operator/pull/111
https://github.com/codeready-toolchain/api/pull/65


deploy/deployment.yaml
```diff
@@ -3,10 +3,20 @@ apiVersion: v1
 metadata:
```
xcoulon: the file does not only contains a "templated deployment", so maybe we should rename it?

deploy/deployment.yaml
```diff
@@ -3,10 +3,20 @@ apiVersion: v1
...
+      namespace: ${NAMESPACE}
   - kind: Deployment
     apiVersion: apps/v1
     metadata:
+      labels:
+        provider: codeready-toolchain
       name: registration-service
+      namespace: ${NAMESPACE}
     spec:
       replicas: ${{REPLICAS}}
```
xcoulon: there is no need for `{{` here: `{` is enough (see https://docs.openshift.com/container-platform/3.11/dev_guide/templates.html#writing-parameters)

MatousJobanek: we need `{{` because the value that is placed there is an integer with single curly braces it puts a string - I've already tried that ;-)

deploy/deployment.yaml
MatousJobanek: do you mean to rename the file? yes, I can rename it - see my proposal here: https://github.com/codeready-toolchain/api/pull/65#discussion_r352639132

MatousJobanek: renamed 8025b63

deploy/deployment.yaml
xcoulon: ah right, I forgot about that 🤦‍♂

PR #78, opened 2019-12-05
Adding usersignups email

tinakurian:
As an admin, I want to see the user email when approving an account. This makes my work easier as I can see where the user is coming from (e.g. redhat.com).

Relates to:
Jira - https://jira.coreos.com/browse/CRT-374
API - codeready-toolchain/api#69
E2E- codeready-toolchain/toolchain-e2e#55
host - https://github.com/codeready-toolchain/host-operator/pull/115

### DIFF
--- a/pkg/controller/signup.go
+++ b/pkg/controller/signup.go
@@ -8,7 +8,6 @@ import (
 	"github.com/codeready-toolchain/registration-service/pkg/errors"
 	"github.com/codeready-toolchain/registration-service/pkg/log"
 	"github.com/codeready-toolchain/registration-service/pkg/signup"
-	"github.com/codeready-toolchain/toolchain-common/pkg/toolchain"
 
 	"github.com/gin-gonic/gin"
 )
@@ -29,8 +28,7 @@ func NewSignup(config *configuration.Registry, signupService signup.Service) *Si
 
 // PostHandler creates a Signup resource
 func (s *Signup) PostHandler(ctx *gin.Context) {
-	compliantEmailLabel := toolchain.ToValidValue(ctx.GetString(context.EmailKey))
-	userSignup, err := s.signupService.CreateUserSignup(ctx.GetString(context.UsernameKey), ctx.GetString(context.SubKey), compliantEmailLabel)
+	userSignup, err := s.signupService.CreateUserSignup(ctx.GetString(context.UsernameKey), ctx.GetString(context.SubKey), ctx.GetString(context.EmailKey))
 	if err != nil {
 		log.Error(ctx, err, "error creating UserSignup resource")
 		errors.AbortWithError(ctx, http.StatusInternalServerError, err, "error creating UserSignup resource")
--- a/pkg/controller/signup.go
+++ b/pkg/controller/signup.go
@@ -8,6 +8,7 @@ import (
 	"github.com/codeready-toolchain/registration-service/pkg/errors"
 	"github.com/codeready-toolchain/registration-service/pkg/log"
 	"github.com/codeready-toolchain/registration-service/pkg/signup"
+	"github.com/codeready-toolchain/toolchain-common/pkg/toolchain"
 
 	"github.com/gin-gonic/gin"
 )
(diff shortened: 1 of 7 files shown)


pkg/controller/signup.go
```diff
@@ -28,7 +29,8 @@ func NewSignup(config *configuration.Registry, signupService signup.Service) *Si
 
 // PostHandler creates a Signup resource
 func (s *Signup) PostHandler(ctx *gin.Context) {
-	userSignup, err := s.signupService.CreateUserSignup(ctx.GetString(context.UsernameKey), ctx.GetString(context.SubKey))
+	compliantEmailLabel := toolchain.ToValidValue(ctx.GetString(context.EmailKey))
```
alexeykazakov: It's not a label. It's annotation.
And what is `toolchain.ToValidValue()`? I don't see it in common.

tinakurian: @alexeykazakov i changed these PRs to WIP so that i can change them to be labels based off of a discussion with @xcoulon. Apparently there is no way to change a PR to a Draft once it's open =/

pkg/controller/signup_test.go
```diff
@@ -82,8 +88,8 @@ func (s *TestSignupSuite) TestSignupPostHandler() {
 			},
 		}
 
-		svc.MockCreateUserSignup = func(username, userID string) (*crtapi.UserSignup, error) {
-			assert.Equal(s.T(), expectedUserID, userID)
+		svc.MockCreateUserSignup = func(username, userID, email string) (*crtapi.UserSignup, error) {
+			assert.Equal(s.T(), expectedUserID, userID, email)
```
alexeykazakov: ?
shouldn't it be something like that instead?
```suggestion
			assert.Equal(s.T(), expectedUserID, userID)
			assert.Equal(s.T(), expectedEmail, email)
```

pkg/signup/signup_service_test.go
```diff
@@ -39,7 +39,7 @@ func (s *TestSignupServiceSuite) TestCreateUserSignup() {
 	userID, err := uuid.NewV4()
 	require.NoError(s.T(), err)
 
-	userSignup, err := svc.CreateUserSignup("jsmith", userID.String())
+	userSignup, err := svc.CreateUserSignup("jsmith", userID.String(), "jsmith@gmail.com")
```
alexeykazakov: Also in the end of this test (line 61) check that the annotation is present in the signup you get from the fake client.

pkg/controller/signup_test.go
tinakurian: sorry.. refactoring error - fixed here: f7253526905ab1e064b76a0b63b0fe5f316eb2d6

pkg/signup/signup_service_test.go
tinakurian: done: c9c8d71e8c9a03b9177866327e11a54f3b86c95f

pkg/controller/signup.go
xcoulon: @alexeykazakov why don't we convert the email to a valid label value and set it as a label on the user signup? that way, we could look-up a usersignup resource based on the email address of the user to approve... WDYT?

alexeykazakov: It would be nice to be able to look up a user by email indeed. But just transforming email to a dns compliant string won’t be enough for that. You can’t tell what was the original email just by looking at the transformed string (if we want to keep it human readable).
And we need the original email anyway. For now it will help us with approving users. Later we will need it for email notifications.
So let’s add this annotation first and deal with the lookup later.

tinakurian:
> Looks good. Just waiting for the e2e tests before approving.

Hmm maybe try refreshing? It has passed since 3hours back - see image: 
image

alexeykazakov:
I'm talking about https://github.com/codeready-toolchain/toolchain-e2e/pull/55

tinakurian:
> I'm talking about [codeready-toolchain/toolchain-e2e#55](https://github.com/codeready-toolchain/toolchain-e2e/pull/55)

@alexeykazakov that won't pass until this is merged - it requires the changes in registration-service

PR #86, opened 2019-12-16
Check all 2xx httpstatus

tinakurian:
check for all 2xx status codes

### DIFF
--- a/pkg/assets/landingpage.js
+++ b/pkg/assets/landingpage.js
@@ -157,7 +157,6 @@ function signup() {
     } else {
       hideAll();
       show('state-waiting-for-approval');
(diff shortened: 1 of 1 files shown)


pkg/assets/landingpage.js
```diff
@@ -15,7 +15,7 @@ var getJSON = function(method, url, token, callback) {
   xhr.responseType = 'json';
   xhr.onload = function() {
     var status = xhr.status;
-    if (status === 200) {
+    if (Math.floor(status / 100) == 2) {
```
xcoulon: why about using a range, such as ` status >= 200 && status < 300` ?

tinakurian: okay done

PR #87, opened 2020-01-10
Do not auto-signup if user didn't click on Get Started

alexeykazakov:
Fixes https://issues.redhat.com/browse/CRT-404

### DIFF
--- a/pkg/assets/index.html
+++ b/pkg/assets/index.html
@@ -44,7 +44,7 @@
       <div class="pf-c-page__header-tools-group">
         <div class="pf-m-user">
           <span id="user-loggedin"><span id="username"></span> <button class="logoutbutton" onclick="keycloak.logout()">Logout</button></span>
-          <span id="user-notloggedin" onclick="keycloak.login()">Login/Signup</span>
+          <span id="user-notloggedin" onclick="login()">Login/Signup</span>
         </div>
       </div>
     </div>
@@ -120,7 +120,7 @@
       </div>
       <div class="pf-c-content">
         <div class="toolchain-state" id="state-getstarted">
-          <button class="getstartedbutton" onclick="keycloak.login()">Get Started with CodeReady Toolchain</button>
+          <button class="getstartedbutton" onclick="login()">Get Started with CodeReady Toolchain</button>
           <p class="terms"><a href="#">Terms and Conditions</a> &copy; 2020, Red Hat Inc.</p>
         </div>
         <div class="toolchain-state" id="state-waiting-for-approval">
--- a/pkg/assets/landingpage.js
+++ b/pkg/assets/landingpage.js
@@ -134,8 +134,17 @@ function updateSignupState() {
     }
   }, function(err) {
     if (err === 404) {
-      // signup does not exist, but user is authorized, start signup process.
-      signup();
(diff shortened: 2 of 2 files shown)


pkg/assets/index.html
```diff
@@ -44,7 +44,7 @@
       <div class="pf-c-page__header-tools-group">
         <div class="pf-m-user">
           <span id="user-loggedin"><span id="username"></span> <button class="logoutbutton" onclick="keycloak.logout()">Logout</button></span>
-          <span id="user-notloggedin" onclick="keycloak.login()">Login/Signup</span>
+          <span id="user-notloggedin" onclick="login()">Login/Signup</span>
```
michaelkleinhenz: Shouldn't that be left at `keycloak.login()`? Calling `login()` here looks like it is doing the orginal behaviour..

MatousJobanek: You are probably right @michaelkleinhenz, but then we should rename the button to `Login` only. For now `Login/Signup` means that you login and also immediately signup for the service.
```suggestion
          <span id="user-notloggedin" onclick="keycloak.login()">Login</span>
```

alexeykazakov: This PR doesn’t change the behavior of Login/Signup and Get Started buttons.
They did well - login **and** signup. So, they now do the same.

The change is in auto-signup. When user is already logged in, we don’t auto-signup. The user has to click on the button (Login/Signup if not logged-in, or Get Started regardless of login)

michaelkleinhenz: Got it, but with this, the user will be auto signed-up when he is using the login/signup link on our page. Wouldn't it be better to "only" do login? I guess that would be also consistent to other RHD pages. 

@MatousJobanek as I'd understand this, "signup" in the header is meant to be the RHD account signup, not the Toolchain signup. This afaik is also consistent with other RHD subpages.

MatousJobanek: Good point, I was too focused on hosted toolchain so I didn't realize that the `Signup` refers to RH account signup. 

With this, I agree with @michaelkleinhenz - when users logs in to RHD account, he shouldn't be automatically registered to the hosted toolchain

tinakurian: i also agree with @michaelkleinhenz  if the user clicks login they shouldn't be automatically signed up - they should have to click the getting started button. this would make the behaviour consistent.

alexeykazakov: In RHD yes, it will just login to RHD. And that's fine since it's RHD site. But in our case, it's not RHD. It just looks like RHD. And IMO it can confuse users if login/signup doesn't actually signup for CRT. The UI can be treated as CRT UI.
Alternativly we could remove that Login/Signup button completely.

MatousJobanek: > Alternativly we could remove that Login/Signup button completely.

Or just rename it to `Login` :-)

alexeykazakov: Login to what? To RHD? But user may think that it will login them to CRT which is not the case.

alexeykazakov: My point is we should not have any RHD specific functionality in our page which can confuse users because it's not obvious that we imitating RHD web site here.

PR #102, opened 2020-04-17
Extract user information from token and store in UserSignupSpec

tinakurian:
In order to customize notification messages to users, we require additional user information which we should store in the UserSignup CRD when the user first registers. The new fields being added for the notification service messages are givenName, familyName and company. This PR will extract these fields from the token and populate the crd accordingly.

jira: https://issues.redhat.com/browse/CRT-518
related e2e-tests PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/102/files

### DIFF
--- a/pkg/context/keys.go
+++ b/pkg/context/keys.go
@@ -5,14 +5,14 @@ const (
 	UsernameKey = "username"
 	// EmailKey is the context key for the email claim
 	EmailKey = "email"
+	// NameKey is the context key for the given name claim
+	GivenName = "givenName"
+	// FamilyNameKey is the context key for the family name claim
+	FamilyNameKey = "familyName"
+	// Company is the context key for the company claim
+	Company = "company"
 	// SubKey is the context key for the subject claim
 	SubKey = "subject"
 	// JWTClaimsKey is the context key for the claims struct
 	JWTClaimsKey = "jwtClaims"
-	// FamilyNameKey is the context key for the family name claim
-	FamilyNameKey = "familyName"
-	// NameKey is the context key for the given name claim
-	GivenName = "givenName"
-	// Company is the context key for the company claim
-	Company = "company"
 )
--- a/pkg/context/keys.go
+++ b/pkg/context/keys.go
@@ -5,12 +5,12 @@ const (
 	UsernameKey = "username"
 	// EmailKey is the context key for the email claim
 	EmailKey = "email"
-	// NameKey is the context key for the given name claim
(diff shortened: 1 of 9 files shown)


pkg/context/keys.go
```diff
@@ -1,11 +1,17 @@
...
-	// EmailKey is the context key for email claim
+	// EmailKey is the context key for the email claim
 	EmailKey = "email"
-	// SubKey is the context key for subject claim
+	// NameKey is the context key for the given name claim
+	GivenName = "givenName"
+	// FamilyNameKey is the context key for the family name claim
+	FamilyNameKey = "familyName"
+	// Company is the context key for the company claim
+	Company = "company"
```
alexeykazakov, line 13:
```suggestion
	// GivenNameKey is the context key for the given name claim
	GivenNameKey = "givenName"
	// FamilyNameKey is the context key for the family name claim
	FamilyNameKey = "familyName"
	// CompanyKey is the context key for the company claim
	CompanyKey = "company"
```

pkg/controller/signup.go
```diff
@@ -28,7 +28,7 @@ func NewSignup(config *configuration.Registry, signupService signup.Service) *Si
 
 // PostHandler creates a Signup resource
 func (s *Signup) PostHandler(ctx *gin.Context) {
-	userSignup, err := s.signupService.CreateUserSignup(ctx.GetString(context.UsernameKey), ctx.GetString(context.SubKey), ctx.GetString(context.EmailKey))
+	userSignup, err := s.signupService.CreateUserSignup(ctx.GetString(context.UsernameKey), ctx.GetString(context.SubKey), ctx.GetString(context.EmailKey), ctx.GetString(context.GivenName), ctx.GetString(context.FamilyNameKey), ctx.GetString(context.Company))
```
alexeykazakov, line 31: There are too many params...
Let's maybe just pass the context as the only param and extract the required properties from the context in the service?

pkg/context/keys.go
tinakurian: fixed as part of: 809fe7555398f3cfc60debeeb9f8c8cab16ec423

pkg/controller/signup.go
tinakurian: Agreed, fixed as part of: 809fe7555398f3cfc60debeeb9f8c8cab16ec423

tinakurian:
> Looks good overall but what about e2e tests?

thanks Alexey, working on it, should have it out by tomorrow!

PR #104, opened 2020-04-23
Show error message after auth failed

rajivnathan:
**Related Issue(s)**
https://issues.redhat.com/browse/CRT-533

**What does this PR do?**

Updates the front end registration service page to show authentication errors when the problem occurs. The error was previously only visible in the browser's dev console.

**Acceptance Criteria**

- [x] Show error in the UI after auth failure
- [x] Error should disappear after successful auth (See note 2 below)
- [x] Error should still be shown in the browser's dev console
- [x] Display error message if there is one

Before:
image

After:
image

**Notes**

1. Moved the div for the error message below the button because it looked a bit odd to me that it was on top but I'm happy to change it back if that breaks some UX guidelines or something. This is how it looked with the message on the top:
image

2. Was able to verify the error disappears after auth is successful by testing with a leeway that increases after each request. This was only for testing, not a part of the change.

```
var leeway int64 = -5000

func (c *TokenClaims) Valid() error {
	c.StandardClaims.IssuedAt -= leeway
	err := c.StandardClaims.Valid()
	c.StandardClaims.IssuedAt += leeway
	leeway += 2000
	return err
}
```

### DIFF
--- a/pkg/assets/index.html
+++ b/pkg/assets/index.html
@@ -109,6 +109,18 @@
       </nav> 
     </section>
     <section class="pf-c-page__main-section">
+      <div id="state-error">
+        <div class="pf-c-alert pf-m-danger pf-m-inline" aria-label="error">
+          <div class="pf-c-alert__icon">
+            <i class="fas fa-exclamation-circle" aria-hidden="true"></i>
+          </div>
+          <h4 class="pf-c-alert__title">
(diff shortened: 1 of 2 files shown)


rajivnathan:
@alexeykazakov Thanks. Sure, I kind of think the error message should be less pronounced so made it look like this:

image

If there's no error then it looks like this:
image

PR #118, opened 2020-08-09
Add configuration params for phone verification

sbryzak:
Add new configuration parameters for phone verification.

Fixes https://issues.redhat.com/browse/CRT-773

### DIFF
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -94,6 +94,29 @@ const (
 	varNamespace = "namespace"
 	// DefaultNamespace is the default k8s namespace to use.
 	DefaultNamespace = "toolchain-host-operator"
+
+	varVerificationEnabled = "verification.enabled"
+	// DefaultVerificationEnabled is the default value for whether the phone verification feature is enabled
+	DefaultVerificationEnabled = true
+
+	varVerificationDailyLimit = "verification.daily_limit"
+	// DefaultVerificationDailyLimit is the default number of times a user may request phone verification
+	// in a 24 hour period
+	DefaultVerificationDailyLimit = 5
+
+	varVerificationAttemptsAllowed = "verification.attempts_allowed"
+	// DefaultVerificationAttemptsAllowed is the default number of maximum attempts a user may make to
+	// provide a correct verification code
+	DefaultVerificationAttemptsAllowed = 3
+
+	// varTwilioAccountSID is the constant used to read the configuration parameter for the
+	// Twilio account identifier, used for sending SMS verification codes.  Twilio is a service that
+	// provides an API for sending SMS messages anywhere in the world.  http://twilio.com
+	varTwilioAccountSID = "twilio.account_sid"
+
+	// varTwilioAuthToken is the constant used to read the configuration parameter for the
+	// Twilio authentication token, used for sending SMS verification codes
+	varTwilioAuthToken = "twilio.auth_token"
 )
 
 // Registry encapsulates the Viper configuration registry which stores the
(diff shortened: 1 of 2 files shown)


pkg/configuration/configuration_test.go
```diff
@@ -495,3 +495,149 @@ func (s *TestConfigurationSuite) TestGetNamespace() {
 		assert.Equal(s.T(), newVal, config.GetNamespace())
 	})
 }
+
+func (s *TestConfigurationSuite) TestVerificationEnabled() {
+	key := configuration.EnvPrefix + "_" + "VERIFICATION_ENABLED"
+	resetFunc := UnsetEnvVarAndRestore(s.T(), key)
+	defer resetFunc()
```
xcoulon, line 502: Even if I understand the logic, I find it a bit weird that you don't use the `resetFunc` in the last sub test

pkg/configuration/configuration_test.go
```diff
@@ -495,3 +495,149 @@ func (s *TestConfigurationSuite) TestGetNamespace() {
...
+
+	s.Run("default", func() {
+		resetFunc := UnsetEnvVarAndRestore(s.T(), key)
+		defer resetFunc()
+		config := s.getDefaultConfiguration()
+		require.Equal(s.T(), true, config.GetVerificationEnabled())
```
alexeykazakov, line 508: Or simply
```suggestion
		require.True(s.T(), config.GetVerificationEnabled())
```

pkg/configuration/configuration_test.go
```diff
@@ -495,3 +495,149 @@ func (s *TestConfigurationSuite) TestGetNamespace() {
...
+	})
+
+	s.Run("file", func() {
+		resetFunc := UnsetEnvVarAndRestore(s.T(), key)
+		defer resetFunc()
+		config := s.getFileConfiguration(`verification.enabled: "` + strconv.FormatBool(false) + `"`)
+		assert.Equal(s.T(), false, config.GetVerificationEnabled())
```
alexeykazakov, line 515:
```suggestion
		config := s.getFileConfiguration(`verification.enabled: "false"`)
		assert.False(s.T(), config.GetVerificationEnabled())
```

pkg/configuration/configuration_test.go
```diff
@@ -495,3 +495,149 @@ func (s *TestConfigurationSuite) TestGetNamespace() {
...
+	})
+
+	s.Run("env overwrite", func() {
+		newVal := strconv.FormatBool(false)
+		err := os.Setenv(key, newVal)
+		require.NoError(s.T(), err)
+		config := s.getDefaultConfiguration()
+		assert.Equal(s.T(), false, config.GetVerificationEnabled())
```
alexeykazakov, line 523:
```suggestion
		err := os.Setenv(key, "false")
		require.NoError(s.T(), err)
		config := s.getDefaultConfiguration()
		assert.False(s.T(), config.GetVerificationEnabled())
```

pkg/configuration/configuration_test.go
sbryzak: It is a bit weird, I copied this structure from previous tests though.  My guess is that the test-level one is there as a "catch-all" in case any of the subtests fail to call it.

pkg/configuration/configuration_test.go
sbryzak: Of course :)  I've fixed this and the other instances.

pkg/configuration/configuration_test.go
```diff
@@ -495,3 +495,148 @@ func (s *TestConfigurationSuite) TestGetNamespace() {
...
+		resetFunc := UnsetEnvVarAndRestore(s.T(), key)
+		defer resetFunc()
+		config := s.getDefaultConfiguration()
+		require.True(s.T(), config.GetVerificationEnabled())
+	})
+
+	s.Run("file", func() {
+		resetFunc := UnsetEnvVarAndRestore(s.T(), key)
+		defer resetFunc()
+		config := s.getFileConfiguration(`verification.enabled: "` + strconv.FormatBool(false) + `"`)
```
alexeykazakov, line 514: This one still looks a bit weird ;)

sbryzak: Agreed, fixed :)

pkg/configuration/configuration.go
```diff
@@ -94,6 +94,29 @@ const (
 	varNamespace = "namespace"
 	// DefaultNamespace is the default k8s namespace to use.
 	DefaultNamespace = "toolchain-host-operator"
+
+	varVerificationEnabled = "verification.enabled"
```
tinakurian, line 98: @sbryzak i think we missed this one: VERIFICATION_DOMAINS_EXCLUDED

PR #119, opened 2020-08-13
Add verification service for SMS verification

sbryzak:
This PR adds a verification service which the controller may use to send verification messages to users.

Fixes https://issues.redhat.com/browse/CRT-774


sbryzak:
I've currently got tests for this service in my local repo, however they contain sensitive info (auth tokens etc for Twilio) and I'm not sure how we should proceed with testing here.  I'm open to any suggestions, however one idea is to just provide a "mode" of the verification service which doesn't do anything when a message is sent.

pkg/verification/verification_service.go
```diff
@@ -0,0 +1,82 @@
...
+	toNumber := signup.Labels[v1alpha1.UserSignupPhoneNumberLabelKey]
+	client := twilio.NewClient(s.config.GetTwilioAccountSID(), s.config.GetTwilioAuthToken(), nil)
+	msg, err := client.Messages.SendMessage(s.config.GetTwilioFromNumber(), toNumber,
+		content, nil)
+	if err != nil {
+		return err
+	}
+
+	if msg.ErrorCode != 0 {
+		fmt.Printf(msg.ErrorMessage)
```
sbryzak, line 63: Just pointing out that I know I still need to fix this line :) I'm currently still thinking about how to handle it.

PR #120, opened 2020-08-13
Added verification service

sbryzak:
Fixes https://issues.redhat.com/browse/CRT-774

### DIFF
--- a/pkg/controller/signup_test.go
+++ b/pkg/controller/signup_test.go
@@ -42,8 +42,11 @@ func (s *TestSignupSuite) TestSignupPostHandler() {
 	// Create a mock SignupService
 	svc := &FakeSignupService{}
 
+	// Create a mock VerificationService
+	verifySvc := &FakeVerificationService{}
+
 	// Create signup instance.
-	signupCtrl := controller.NewSignup(s.Config, svc)
+	signupCtrl := controller.NewSignup(s.Config, svc, verifySvc)
 	handler := gin.HandlerFunc(signupCtrl.PostHandler)
 
 	userID, err := uuid.NewV4()
@@ -123,13 +126,17 @@ func (s *TestSignupSuite) TestSignupGetHandler() {
 
 	// Create a mock SignupService
 	svc := &FakeSignupService{}
+
+	// Create a mock VerificationService
+	verifyService := &FakeVerificationService{}
+
 	// Create UserSignup
 	ob, err := uuid.NewV4()
 	require.NoError(s.T(), err)
 	userID := ob.String()
 
 	// Create Signup controller instance.
-	ctrl := controller.NewSignup(s.Config, svc)
+	ctrl := controller.NewSignup(s.Config, svc, verifyService)
 	handler := gin.HandlerFunc(ctrl.GetHandler)
 
 	s.Run("signups found", func() {
@@ -216,4 +223,12 @@ func (m *FakeSignupService) GetSignup(userID string) (*signup.Signup, error) {
 
 func (m *FakeSignupService) CreateUserSignup(ctx *gin.Context) (*crtapi.UserSignup, error) {
 	return m.MockCreateUserSignup(ctx)
+}
+
+type FakeVerificationService struct {
+	MockSendVerification func(ctx *gin.Context, signup *crtapi.UserSignup) error
+}
+
+func (m *FakeVerificationService) SendVerification(ctx *gin.Context, signup *crtapi.UserSignup) error {
+	return m.SendVerification(ctx, signup)
 }
--- a/pkg/controller/signup_test.go
+++ b/pkg/controller/signup_test.go
@@ -226,7 +226,6 @@ func (m *FakeSignupService) CreateUserSignup(ctx *gin.Context) (*crtapi.UserSign
 }
 
 type FakeVerificationService struct {
-	MockSendVerification func(ctx *gin.Context, signup *crtapi.UserSignup) error
 }
 
 func (m *FakeVerificationService) SendVerification(ctx *gin.Context, signup *crtapi.UserSignup) error {
--- a/pkg/verification/verification_service.go
+++ b/pkg/verification/verification_service.go
@@ -58,8 +58,6 @@ func (s *ServiceImpl) SendVerification(ctx *gin.Context, signup *v1alpha1.UserSi
 		return err
 	}
 
-	signup.Annotations[v1alpha1.UserSignupVerificationCodeAnnotationKey] = verificationCode
-
 	content := fmt.Sprintf(s.config.GetVerificationMessageTemplate(), verificationCode)
 
 	toNumber := signup.Labels[v1alpha1.UserSignupPhoneNumberLabelKey]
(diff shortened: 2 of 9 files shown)


pkg/verification/verification_service.go
```diff
@@ -0,0 +1,82 @@
...
+	SendVerification(signup *v1alpha1.UserSignup) error
+}
+
+// ServiceImpl represents the implementation of the verification service.
+type ServiceImpl struct {
+	config ServiceConfiguration
+}
+
+// NewVerificationService creates a service object for performing user verification
+func NewVerificationService(cfg ServiceConfiguration) (Service, error) {
```
alexeykazakov, line 36: you don't need to return `error`, do you?

pkg/verification/verification_service.go
```diff
@@ -0,0 +1,82 @@
...
+	toNumber := signup.Labels[v1alpha1.UserSignupPhoneNumberLabelKey]
+	client := twilio.NewClient(s.config.GetTwilioAccountSID(), s.config.GetTwilioAuthToken(), nil)
+	msg, err := client.Messages.SendMessage(s.config.GetTwilioFromNumber(), toNumber,
+		content, nil)
+	if err != nil {
+		return err
+	}
+
+	if msg.ErrorCode != 0 {
+		fmt.Printf(msg.ErrorMessage)
```
alexeykazakov, line 63: Let's return it's as an error, not print it.

pkg/verification/verification_service.go
sbryzak: Good point, fixed.

pkg/verification/verification_service.go
sbryzak: I think this block is unnecessary, I've removed it all together and added logging to the previous error handler.

sbryzak:
@alexeykazakov This is actually my second PR, the first one I accidentally committed our Twilio auth token for one of our subaccounts which I was using in the tests.  I killed the first PR and deleted the subaccount immediately, thankfully it wasn't the token for the master account which would have been a more difficult problem to fix :)   Anyway about the tests, I was hoping to get some suggestions - there doesn't seem to be any kind of mock API available to use for Twilio, so I'm still trying to think of the best way to test the new verification service without actually having to invoke the Twilio client API.  Any ideas?

alexeykazakov:
@sbryzak the twillo client uses http client behind the scenes, I guess? In that case we could use gock for example to mock the http client.
It's trickier with kube because go-client won't initiate without kube token present in the local file system, and this is why we don't have tests for our kube services yet, but it should not be a problem for testing twillo.

sbryzak:
We can actually pass in an HttpClient to the Twilio client constructor (it's the third parameter):

`twilio.NewClient(s.config.GetTwilioAccountSID(), s.config.GetTwilioAuthToken(), nil)`

I'll see what I can do to allow us to inject a mock client here.

pkg/controller/signup_test.go
```diff
@@ -217,3 +224,11 @@ func (m *FakeSignupService) GetSignup(userID string) (*signup.Signup, error) {
 func (m *FakeSignupService) CreateUserSignup(ctx *gin.Context) (*crtapi.UserSignup, error) {
 	return m.MockCreateUserSignup(ctx)
 }
+
+type FakeVerificationService struct {
+	MockSendVerification func(ctx *gin.Context, signup *crtapi.UserSignup) error
```
alexeykazakov, line 229: This function is not used anywhere, is it?

pkg/verification/verification_service.go
```diff
@@ -0,0 +1,87 @@
...
+// ServiceImpl represents the implementation of the verification service.
+type ServiceImpl struct {
+	config     ServiceConfiguration
+	HttpClient *http.Client
+}
+
+type VerificationServiceOption func(svc *ServiceImpl)
+
+// NewVerificationService creates a service object for performing user verification
+func NewVerificationService(cfg ServiceConfiguration, opts ...VerificationServiceOption) Service {
```
alexeykazakov, line 41: You are not initiating any default http client? Looks a bit error prone. Http client is not among the required params but it is actually required.

pkg/verification/verification_service.go
```diff
@@ -0,0 +1,87 @@
...
+		config: cfg,
+	}
+
+	for _, opt := range opts {
+		opt(s)
+	}
+	return s
+}
+
+// SendVerification sends a verification message to the specified user
```
alexeykazakov, line 52: Let's also mention that the function will add the verification code annotation to the signup.

pkg/verification/verification_service_test.go
```diff
@@ -0,0 +1,148 @@
...
+		fromNumber:      fromNumber,
+		messageTemplate: configuration.DefaultVerificationMessageTemplate,
+	}
+}
+
+func (s *TestVerificationServiceSuite) TestVerify() {
+	defer gock.Off()
+	gock.New("https://api.twilio.com").
+		Reply(http.StatusNoContent).
+		BodyString("")
```
alexeykazakov, line 68: Let's check that the mock service got the expected api call: correct phone number and the verification code (the code is available in the annotation).

pkg/verification/verification_service_test.go
```diff
@@ -0,0 +1,148 @@
...
+				v1alpha1.UserSignupPhoneNumberLabelKey: "+1NUMBER",
+			},
+		},
+		Spec: v1alpha1.UserSignupSpec{
+			Username: "sbryzak@redhat.com",
+		},
+	}
+
+	err := svc.SendVerification(ctx, userSignup)
+	require.Error(s.T(), err)
```
alexeykazakov, line 125: Check the error message too.

sbryzak: Done

pkg/verification/verification_service_test.go
sbryzak: This was not easy, but I worked out a way to do it :)

pkg/verification/verification_service.go
sbryzak: Done

pkg/verification/verification_service.go
sbryzak: It's actually not required.  The very first thing that the `NewClient()` function does is create a default http client if it isn't set:

```
func NewClient(accountSid string, authToken string, httpClient *http.Client) *Client {
	if httpClient == nil {
		httpClient = defaultHttpClient
	}
```

pkg/controller/signup_test.go
sbryzak: You're correct, it's not actually needed.  I've removed it.

pkg/verification/verification_service.go
alexeykazakov: All good then!

pkg/verification/verification_service_test.go
```diff
@@ -134,6 +157,7 @@ func (s *TestVerificationServiceSuite) createVerificationService() (verification
 		"CodeReady",
 	)
 
+	//	httpClient := &ClientMock{client: &http.Client{Transport: &http.Transport{}}}
```
alexeykazakov, line 160: I guess we don't need this comment?

sbryzak: Removed

PR #122, opened 2020-08-17
secret configuration for registration-service

tinakurian:
This PR adds required configuration for twilio.

This PR is also related to allowing registration-service to have it's configurations set via a configmap and set secrets (particularily for twilio).

Related PR: codeready-toolchain/host-operator#263, PR for toolchain-infra coming soon...

Related Jira: https://issues.redhat.com/browse/CRT-775

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -46,7 +46,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	// create client that will be used for retrieving the host operator secret
+	// create client that will be used for retrieving the registration service configmaps and secrets
 	cl, err := client.New(cfg, client.Options{})
 	if err != nil {
 		panic(err.Error())
@@ -82,12 +82,12 @@ func main() {
 }
 
 func printConfig(cfg *configuration.Registry) {
-	log.Info(nil, "AuthRawSSLRequired: "+cfg.GetAuthClientConfigAuthRawSSLReuired())
-	log.Info(nil, "AuthRawServerURL: "+cfg.GetAuthClientConfigAuthRawServerURL())
-	log.Info(nil, "AuthRawResource: "+cfg.GetAuthClientConfigAuthRawResource())
-	log.Info(nil, "AuthRawRealm: "+cfg.GetAuthClientConfigAuthRawRealm())
-	log.Info(nil, "AuthRawPublicClient: "+strconv.FormatBool(cfg.GetAuthClientConfigAuthRawPublicClient()))
-	log.Info(nil, "AuthRawClientID: "+cfg.GetAuthClientConfigAuthRawClientID())
+	log.Info(nil, "AuthRawSSLRequired: "+cfg.GetAuthClientConfigRawSSLRequired())
+	log.Info(nil, "AuthRawServerURL: "+cfg.GetAuthClientConfigRawAuthServerURL())
+	log.Info(nil, "AuthRawResource: "+cfg.GetAuthClientConfigRawResource())
+	log.Info(nil, "AuthRawRealm: "+cfg.GetAuthClientConfigRawRealm())
+	log.Info(nil, "AuthRawPublicClient: "+strconv.FormatBool(cfg.GetAuthClientConfigRawPublicClient()))
+	log.Info(nil, "AuthRawClientID: "+cfg.GetAuthClientConfigRawClientID())
 	log.Info(nil, "VerificationEnabled: "+strconv.FormatBool(cfg.GetVerificationEnabled()))
 	log.Info(nil, "VerificationDailyLimit: "+strconv.Itoa(cfg.GetVerificationDailyLimit()))
 	log.Info(nil, "VerificationAttemptsAllowed: "+strconv.Itoa(cfg.GetVerificationAttemptsAllowed()))
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -82,12 +82,7 @@ func main() {
 }
 
 func printConfig(cfg *configuration.Registry) {
-	log.Info(nil, "AuthRawSSLRequired: "+cfg.GetAuthClientConfigRawSSLRequired())
-	log.Info(nil, "AuthRawServerURL: "+cfg.GetAuthClientConfigRawAuthServerURL())
-	log.Info(nil, "AuthRawResource: "+cfg.GetAuthClientConfigRawResource())
-	log.Info(nil, "AuthRawRealm: "+cfg.GetAuthClientConfigRawRealm())
-	log.Info(nil, "AuthRawPublicClient: "+strconv.FormatBool(cfg.GetAuthClientConfigRawPublicClient()))
-	log.Info(nil, "AuthRawClientID: "+cfg.GetAuthClientConfigRawClientID())
+	log.Info(nil, "AuthRawConfig: "+cfg.GetAuthClientConfigAuthRaw())
 	log.Info(nil, "VerificationEnabled: "+strconv.FormatBool(cfg.GetVerificationEnabled()))
 	log.Info(nil, "VerificationDailyLimit: "+strconv.Itoa(cfg.GetVerificationDailyLimit()))
 	log.Info(nil, "VerificationAttemptsAllowed: "+strconv.Itoa(cfg.GetVerificationAttemptsAllowed()))
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -12,6 +12,8 @@ import (
 	"github.com/codeready-toolchain/registration-service/pkg/log"
 	"github.com/codeready-toolchain/registration-service/pkg/server"
 	"github.com/spf13/pflag"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 )
 
 func main() {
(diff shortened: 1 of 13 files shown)


cmd/main.go
```diff
@@ -64,6 +81,18 @@ func main() {
 	gracefulShutdown(srv.HTTPServer(), srv.Config().GetGracefulTimeout())
 }
 
+func printConfig(cfg *configuration.Registry) {
```
alexeykazakov, line 84: Let's maybe move this function it to `cfg *configuration.Registry` as a method?

deploy/registration-service.yaml
```diff
@@ -172,8 +221,16 @@ objects:
     data:
       environment: ${ENVIRONMENT}
       auth_client.library_url: ${AUTH_CLIENT_LIBRARY_URL}
-      auth_client.config_raw: ${AUTH_CLIENT_CONFIG_RAW}
       auth_client.public_keys_url: ${AUTH_CLIENT_PUBLIC_KEYS_URL}
+      auth_client.config_raw.realm: ${AUTH_CLIENT_CONFIG_RAW_REALM}
```
alexeykazakov, line 225: Why are you spiting the raw configuration? I would still keep the raw configuration as a string. The reg-service doesn't really care what this string is about. It's just a string it needs to pass to the UI client.

deploy/registration-service.yaml
```diff
@@ -104,26 +112,67 @@ objects:
...
-                - name: REGISTRATION_AUTH_CLIENT_LIBRARY_URL
+                      key: auth_client.library_url
+                - name: REGISTRATION_AUTH_CLIENT_PUBLIC_KEYS_URL
                   valueFrom:
                     configMapKeyRef:
                       name: registration-service
-                      key: auth_client.library_url
-                - name: REGISTRATION_AUTH_CLIENT_CONFIG_RAW
+                      key: auth_client.public_keys_url
+                - name: REGISTRATION_AUTH_CLIENT_CONFIG_RAW_REALM
```
MatousJobanek, line 131: Why did you replace the `REGISTRATION_AUTH_CLIENT_CONFIG_RAW` with multiple entries? this is not how it is being configured

deploy/registration-service.yaml
```diff
@@ -183,9 +240,29 @@ parameters:
     value: '3'
   - name: ENVIRONMENT
     value: 'prod' # prod, stage, e2e-tests, dev, etc
+  - name: CONFIG_MAP_NAME
+    value: ''
```
MatousJobanek, line 244: As I have written above - there is no point in configuring the CM name

pkg/configuration/configuration.go
```diff
@@ -112,37 +123,54 @@ const (
...
 
 	// varTwilioAuthToken is the constant used to read the configuration parameter for the
 	// Twilio authentication token, used for sending SMS verification codes
-	varTwilioAuthToken = "twilio.auth_token"
+	varTwilioAuthToken = "twilio.auth.token"
 )
 
 // Registry encapsulates the Viper configuration registry which stores the
 // configuration data in-memory.
 type Registry struct {
```
MatousJobanek, line 135: Btw, we don't use the `Registry` name in any of our operators any more. I believe that it would be worth renaming here it too.

pkg/configuration/configuration.go
```diff
@@ -112,37 +123,54 @@ const (
...
 }
 
 // CreateEmptyRegistry creates an initial, empty registry.
-func CreateEmptyRegistry() *Registry {
+func CreateEmptyRegistry(cl client.Client) (*Registry, error) {
+
+	err := configuration.LoadFromConfigMap(EnvPrefix, "REGISTRATION_SERVICE_CONFIG_MAP_NAME", cl)
+	if err != nil {
+		return nil, err
+	}
```
MatousJobanek, line 146: Why loading from another CM? We already use one CM that is called `registration-service`, there is no point in introducing another one.

deploy/registration-service.yaml
```diff
@@ -104,26 +112,67 @@ objects:
                   valueFrom:
                     fieldRef:
                       fieldPath: metadata.namespace
-                - name: REGISTRATION_ENVIRONMENT
+                - name: REGISTRATION_SERVICE_CONFIG_MAP_NAME
+                  value: ${CONFIG_MAP_NAME}
```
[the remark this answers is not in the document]
tinakurian: @MatousJobanek i think it may be better to have it as a parameter so that whoever creates the configmap can configure this themselves - this isn't a configmpa that we create. it will be documented in toolchain-infra though..

deploy/registration-service.yaml
```diff
@@ -104,26 +112,67 @@ objects:
...
+                - name: REGISTRATION_SERVICE_SECRET_NAME
+                  value: ${SECRET_NAME}
+                - name: WATCH_NAMESPACE
+                  value: ${NAMESPACE}
```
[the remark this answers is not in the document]
tinakurian: hmm after testing ${NAMESPACE} seems to work just fine..

pkg/configuration/configuration.go
tinakurian: this is for someone else to override the configurations which we set by default or via the dev.yaml file. we can support someone creating their own configmap

deploy/registration-service.yaml
tinakurian: It just seemed cleaner to me, it means we don't need to decode the value if we need to access a field (this is done in the tests) and it also makes it easier when creating a configmap imo. lastly, it seems a bit inconsistent with the other fields in the dev.yaml file and now the configmap as well. in any case, i've changed it back

deploy/registration-service.yaml
```diff
@@ -18,26 +18,34 @@ objects:
...
+          - get
+          - list
+      - apiGroups:
+          - ""
+        resources:
+          - configmaps
+          - secrets
+        verbs:
+          - get
+          - list
```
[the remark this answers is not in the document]
tinakurian: thanks for the reminder @MatousJobanek i had mentioned it in the PR description but it's done here in case you missed it: https://github.com/codeready-toolchain/host-operator/pull/263/files

deploy/registration-service.yaml
tinakurian: i changed how it was being configured. as mentioned in another comment above, it seemed a bit inconsistent to have a blob of configurations. having it as separate fields also makes it easier (imo) when creating a configmap. lastly if we ever needed any of these fields we would need to decode it in order to retrieve it (see tests).. but in any case, i've changed it back..

deploy/registration-service.yaml
tinakurian: see other comment

pkg/configuration/configuration.go
tinakurian: @MatousJobanek sounds good, i guess in host-operator we use "Config" - is this fine to use?

deploy/registration-service.yaml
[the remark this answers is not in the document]
MatousJobanek: ok, so you create a logic that loads another ConfigMap with values that can override those values set in the default ConfigMap `registration-service`...? This is well... ...unexpected. 
Wouldn't it be better to incorporate it into the host-operator CM and update the default `registration-service` CM through the RegistrationService_controller?

In addition, what I'm missing is the scenario of setting/changing the `REGISTRATION_SERVICE_CONFIG_MAP_NAME` value. Don't forget that this deployment is managed by the controller.

pkg/configuration/configuration.go
MatousJobanek: Yup :+1:

pkg/configuration/configuration.go
MatousJobanek: Ok, I was missing this context. Anyway, please see my comment here: https://github.com/codeready-toolchain/registration-service/pull/122#discussion_r472997398

deploy/registration-service.yaml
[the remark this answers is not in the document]
MatousJobanek: > the registration_service_config_map_name can be set in the env files. 

but who will set it? If it is set in the runtime (when the reg-service is deployed and runnnig), then it will be replaced to the original value.

> I'm not too sure why it's unexpected?

because we set everything through the host-operator, so I'm not sure why this is separated. We already have a ConfigMap for reg-service, so I just didn't expect another one. 

>  it allows us to create configmaps and secrets the same way we do for host-operator and member-operator.

yeah, so why don't use the configmap/secret of the host-operator? The fact that registration service is a another service/deployment is just a implementation detail of the host-operator.

pkg/configuration/configuration.go
tinakurian: done as part of: 1863479426aebd4c11524c081a17b3773e64e32c

cmd/main.go
```diff
@@ -64,6 +81,18 @@ func main() {
...
+	log.Info(nil, "AuthRawSSLRequired: "+cfg.GetAuthClientConfigRawSSLRequired())
+	log.Info(nil, "AuthRawServerURL: "+cfg.GetAuthClientConfigRawAuthServerURL())
+	log.Info(nil, "AuthRawResource: "+cfg.GetAuthClientConfigRawResource())
+	log.Info(nil, "AuthRawRealm: "+cfg.GetAuthClientConfigRawRealm())
+	log.Info(nil, "AuthRawPublicClient: "+strconv.FormatBool(cfg.GetAuthClientConfigRawPublicClient()))
+	log.Info(nil, "AuthRawClientID: "+cfg.GetAuthClientConfigRawClientID())
+	log.Info(nil, "VerificationEnabled: "+strconv.FormatBool(cfg.GetVerificationEnabled()))
+	log.Info(nil, "VerificationDailyLimit: "+strconv.Itoa(cfg.GetVerificationDailyLimit()))
+	log.Info(nil, "VerificationAttemptsAllowed: "+strconv.Itoa(cfg.GetVerificationAttemptsAllowed()))
```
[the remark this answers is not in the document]
tinakurian: done as part of: 1863479426aebd4c11524c081a17b3773e64e32c

cmd/main.go
tinakurian: done as part of: 1863479426aebd4c11524c081a17b3773e64e32c

deploy/registration-service.yaml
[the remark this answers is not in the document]
tinakurian: @MatousJobanek after a call with @alexeykazakov  we have agreed to remove the configmap stuff and for now hardcodde the name of the secret in configuration.go

PR #124, opened 2020-08-22
Implement verify code endpoint

sbryzak:
This PR introduces a new endpoint for verifying the phone verification code presented by the user.

Fixes https://issues.redhat.com/browse/CRT-767

Also fixes https://issues.redhat.com/browse/CRT-764

### DIFF
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -4,6 +4,7 @@ package configuration
 
 import (
 	"strings"
+	"sync"
 	"time"
 
 	errs "github.com/pkg/errors"
@@ -134,6 +135,8 @@ const (
 	// is not required.  For example: "redhat.com,ibm.com"
 	varVerificationExcludedEmailDomains = "verification.excluded_email_domains"
 )
+
+var lock sync.Mutex
 
 // Registry encapsulates the Viper configuration registry which stores the
 // configuration data in-memory.
@@ -310,6 +313,9 @@ func (c *Registry) GetVerificationMessageTemplate() string {
 // is not required
 func (c *Registry) GetVerificationExcludedEmailDomains() []string {
 	if c.excludedDomains == nil {
+		lock.Lock()
+		defer lock.Unlock()
+
 		c.excludedDomains = strings.Split(c.v.GetString(varVerificationExcludedEmailDomains), ",")
 	}
 	return c.excludedDomains
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -312,10 +312,10 @@ func (c *Registry) GetVerificationMessageTemplate() string {
 // GetVerificationExcludedEmailDomains returns the list of email address domains for which phone verification
 // is not required
 func (c *Registry) GetVerificationExcludedEmailDomains() []string {
-	if c.excludedDomains == nil {
-		lock.Lock()
-		defer lock.Unlock()
+	lock.Lock()
+	defer lock.Unlock()
 
+	if c.excludedDomains == nil {
 		split := func(c rune) bool {
 			return c == ','
 		}
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -97,7 +97,7 @@ const (
 
 	varVerificationEnabled = "verification.enabled"
 	// DefaultVerificationEnabled is the default value for whether the phone verification feature is enabled
-	DefaultVerificationEnabled = true
+	DefaultVerificationEnabled = false
 
 	varVerificationDailyLimit = "verification.daily_limit"
 	// DefaultVerificationDailyLimit is the default number of times a user may request phone verification
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -129,6 +129,10 @@ const (
 	// DefaultVerificationMessageTemplate is the default verification message template sent to users
 	// via SMS for phone verification.  The string parameter is replaced with a random verification code
 	DefaultVerificationMessageTemplate = "Your verification code for Red Hat Developer Sandbox is: %s"
+
+	// varVerificationExcludedEmailDomains contains a comma-separated list of domain names for which phone verification
+	// is not required.  For example: "redhat.com,ibm.com"
+	varVerificationExcludedEmailDomains = "verification.excluded_email_domains"
 )
 
 // Registry encapsulates the Viper configuration registry which stores the
@@ -299,6 +303,12 @@ func (c *Registry) GetVerificationAttemptsAllowed() int {
 // phone verification
 func (c *Registry) GetVerificationMessageTemplate() string {
 	return c.v.GetString(varVerificationMessageTemplate)
+}
+
+// GetVerificationExcludedEmailDomains returns the list of email address domains for which phone verification
+// is not required
+func (c *Registry) GetVerificationExcludedEmailDomains() string {
+	return c.v.GetString(varVerificationExcludedEmailDomains)
 }
 
 // GetTwilioAccountSID is the Twilio account identifier, used for sending phone verification messages
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -316,7 +316,10 @@ func (c *Registry) GetVerificationExcludedEmailDomains() []string {
 		lock.Lock()
 		defer lock.Unlock()
 
-		c.excludedDomains = strings.Split(c.v.GetString(varVerificationExcludedEmailDomains), ",")
+		split := func(c rune) bool {
+			return c == ','
+		}
+		c.excludedDomains = strings.FieldsFunc(c.v.GetString(varVerificationExcludedEmailDomains), split)
 	}
 	return c.excludedDomains
 }
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -138,7 +138,8 @@ const (
 // Registry encapsulates the Viper configuration registry which stores the
 // configuration data in-memory.
 type Registry struct {
-	v *viper.Viper
+	v               *viper.Viper
+	excludedDomains []string
 }
 
 // CreateEmptyRegistry creates an initial, empty registry.
@@ -307,8 +308,11 @@ func (c *Registry) GetVerificationMessageTemplate() string {
 
 // GetVerificationExcludedEmailDomains returns the list of email address domains for which phone verification
 // is not required
-func (c *Registry) GetVerificationExcludedEmailDomains() string {
-	return c.v.GetString(varVerificationExcludedEmailDomains)
+func (c *Registry) GetVerificationExcludedEmailDomains() []string {
+	if c.excludedDomains == nil {
+		c.excludedDomains = strings.Split(c.v.GetString(varVerificationExcludedEmailDomains), ",")
+	}
+	return c.excludedDomains
 }
 
 // GetTwilioAccountSID is the Twilio account identifier, used for sending phone verification messages
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -4,7 +4,6 @@ package configuration
 
 import (
 	"strings"
-	"sync"
 	"time"
 
 	errs "github.com/pkg/errors"
@@ -136,8 +135,6 @@ const (
 	varVerificationExcludedEmailDomains = "verification.excluded_email_domains"
 )
 
-var lock sync.Mutex
-
 // Registry encapsulates the Viper configuration registry which stores the
 // configuration data in-memory.
 type Registry struct {
@@ -155,6 +152,9 @@ func CreateEmptyRegistry() *Registry {
 	c.v.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
 	c.v.SetTypeByDefaultValue(true)
 	c.setConfigDefaults()
+	c.excludedDomains = strings.FieldsFunc(c.v.GetString(varVerificationExcludedEmailDomains), func(c rune) bool {
+		return c == ','
+	})
 	return &c
 }
 
@@ -169,6 +169,11 @@ func New(configFilePath string) (*Registry, error) {
 		err := c.v.ReadInConfig() // Find and read the config file
 		if err != nil {           // Handle errors reading the config file.
 			return nil, errs.Wrap(err, "failed to read config file")
+		}
+		if c.v.GetString(varVerificationExcludedEmailDomains) != "" {
+			c.excludedDomains = strings.FieldsFunc(c.v.GetString(varVerificationExcludedEmailDomains), func(c rune) bool {
+				return c == ','
+			})
 		}
 	}
 	return c, nil
@@ -312,15 +317,6 @@ func (c *Registry) GetVerificationMessageTemplate() string {
 // GetVerificationExcludedEmailDomains returns the list of email address domains for which phone verification
 // is not required
 func (c *Registry) GetVerificationExcludedEmailDomains() []string {
-	lock.Lock()
-	defer lock.Unlock()
-
-	if c.excludedDomains == nil {
-		split := func(c rune) bool {
-			return c == ','
-		}
-		c.excludedDomains = strings.FieldsFunc(c.v.GetString(varVerificationExcludedEmailDomains), split)
-	}
 	return c.excludedDomains
 }
 
--- a/pkg/controller/signup.go
+++ b/pkg/controller/signup.go
@@ -60,3 +60,34 @@ func (s *Signup) GetHandler(ctx *gin.Context) {
 		ctx.JSON(http.StatusOK, signupResource)
 	}
 }
+
+// VerifyCodeHandler validates the phone verification code passed in by the user
+func (s *Signup) VerifyCodeHandler(ctx *gin.Context) {
+	code := ctx.Param("code")
+
+	userID := ctx.GetString(context.SubKey)
+	signupResource, err := s.signupService.GetUserSignup(userID)
+	if err != nil {
+		log.Error(ctx, err, "error getting UserSignup resource")
+		errors.AbortWithError(ctx, http.StatusInternalServerError, err, "error getting UserSignup resource")
+	}
+	if signupResource == nil {
+		log.Errorf(ctx, nil, "UserSignup resource for userID: %s resource not found", userID)
+		ctx.AbortWithStatus(http.StatusNotFound)
+	}
+
+	err = s.verificationService.VerifyCode(ctx, signupResource, code)
+	if err != nil {
+		log.Error(ctx, err, "error validating user verification code")
+		// TODO we need to check what type of error is returned here
+		//errors.AbortWithError(ctx)
+	}
+
+	err = s.signupService.UpdateUserSignup(signupResource)
+	if err != nil {
+		log.Error(ctx, err, "error while updating UserSignup resource")
+		errors.AbortWithError(ctx, http.StatusInternalServerError, err, "error while updating UserSignup resource")
+	}
+
+	ctx.Status(http.StatusOK)
+}
--- a/pkg/controller/signup.go
+++ b/pkg/controller/signup.go
@@ -85,7 +85,7 @@ func (s *Signup) VerifyCodeHandler(ctx *gin.Context) {
 		default:
 			errors.AbortWithError(ctx, http.StatusInternalServerError, err, "error while verifying code")
 		case *errors2.StatusError:
-			errors.AbortWithError(ctx, int(t.ErrStatus.Code), t, t.ErrStatus.Message)
+			errors.AbortWithError(ctx, int(t.ErrStatus.Code), err, t.ErrStatus.Message)
 		}
 	}
 
--- a/pkg/controller/signup.go
+++ b/pkg/controller/signup.go
@@ -66,6 +66,10 @@ func (s *Signup) GetHandler(ctx *gin.Context) {
 // VerifyCodeHandler validates the phone verification code passed in by the user
 func (s *Signup) VerifyCodeHandler(ctx *gin.Context) {
 	code := ctx.Param("code")
+	if code == "" {
+		log.Error(ctx, nil, "no code provided in request")
+		ctx.AbortWithStatus(http.StatusNotFound)
+	}
 
 	userID := ctx.GetString(context.SubKey)
 	signupResource, err := s.signupService.GetUserSignup(userID)
@@ -79,6 +83,16 @@ func (s *Signup) VerifyCodeHandler(ctx *gin.Context) {
 	}
 
 	err = s.verificationService.VerifyCode(ctx, signupResource, code)
+
+	// Regardless of whether the VerifyCode() call returns an error or not, we need to update the UserSignup instance
+	// as its state can be updated even in the case of an error.  This may result in the slight possibility that any
+	// errors returned by VerifyCode() are suppressed, as error handling for the UserSignup update is given precedence.
+	_, err2 := s.signupService.UpdateUserSignup(signupResource)
+	if err2 != nil {
+		log.Error(ctx, err2, "error while updating UserSignup resource")
+		errors.AbortWithError(ctx, http.StatusInternalServerError, err2, "error while updating UserSignup resource")
+	}
+
 	if err != nil {
 		log.Error(ctx, err, "error validating user verification code")
 		switch t := err.(type) {
@@ -87,12 +101,6 @@ func (s *Signup) VerifyCodeHandler(ctx *gin.Context) {
 		case *errors2.StatusError:
 			errors.AbortWithError(ctx, int(t.ErrStatus.Code), err, t.ErrStatus.Message)
 		}
-	}
-
-	_, err = s.signupService.UpdateUserSignup(signupResource)
-	if err != nil {
-		log.Error(ctx, err, "error while updating UserSignup resource")
-		errors.AbortWithError(ctx, http.StatusInternalServerError, err, "error while updating UserSignup resource")
 	}
 
 	ctx.Status(http.StatusOK)
--- a/pkg/controller/signup.go
+++ b/pkg/controller/signup.go
@@ -76,10 +76,13 @@ func (s *Signup) VerifyCodeHandler(ctx *gin.Context) {
 	if err != nil {
 		log.Error(ctx, err, "error getting UserSignup resource")
 		errors.AbortWithError(ctx, http.StatusInternalServerError, err, "error getting UserSignup resource")
+		return
 	}
+
 	if signupResource == nil {
 		log.Errorf(ctx, nil, "UserSignup resource for userID: %s resource not found", userID)
 		ctx.AbortWithStatus(http.StatusNotFound)
+		return
 	}
 
 	err = s.verificationService.VerifyCode(ctx, signupResource, code)
@@ -91,6 +94,7 @@ func (s *Signup) VerifyCodeHandler(ctx *gin.Context) {
 	if err2 != nil {
 		log.Error(ctx, err2, "error while updating UserSignup resource")
 		errors.AbortWithError(ctx, http.StatusInternalServerError, err2, "error while updating UserSignup resource")
+		return
 	}
 
 	if err != nil {
@@ -101,6 +105,7 @@ func (s *Signup) VerifyCodeHandler(ctx *gin.Context) {
 		case *errors2.StatusError:
 			errors.AbortWithError(ctx, int(t.ErrStatus.Code), err, t.ErrStatus.Message)
 		}
+		return
 	}
 
 	ctx.Status(http.StatusOK)
--- a/pkg/controller/signup.go
+++ b/pkg/controller/signup.go
@@ -3,6 +3,8 @@ package controller
 import (
 	"net/http"
 
+	errors2 "k8s.io/apimachinery/pkg/api/errors"
+
 	"github.com/codeready-toolchain/registration-service/pkg/verification"
 
 	"github.com/codeready-toolchain/registration-service/pkg/configuration"
@@ -79,11 +81,15 @@ func (s *Signup) VerifyCodeHandler(ctx *gin.Context) {
 	err = s.verificationService.VerifyCode(ctx, signupResource, code)
 	if err != nil {
 		log.Error(ctx, err, "error validating user verification code")
-		// TODO we need to check what type of error is returned here
-		//errors.AbortWithError(ctx)
+		switch t := err.(type) {
+		default:
+			errors.AbortWithError(ctx, http.StatusInternalServerError, err, "error while verifying code")
+		case *errors2.StatusError:
+			errors.AbortWithError(ctx, int(t.ErrStatus.Code), t, t.ErrStatus.Message)
+		}
 	}
 
-	err = s.signupService.UpdateUserSignup(signupResource)
+	_, err = s.signupService.UpdateUserSignup(signupResource)
 	if err != nil {
 		log.Error(ctx, err, "error while updating UserSignup resource")
 		errors.AbortWithError(ctx, http.StatusInternalServerError, err, "error while updating UserSignup resource")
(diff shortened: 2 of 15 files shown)


pkg/signup/signup_service.go
```diff
@@ -213,3 +239,25 @@ func (s *ServiceImpl) GetSignup(userID string) (*Signup, error) {
 
 	return signupResponse, nil
 }
+
+func (s *ServiceImpl) GetUserSignup(userID string) (*crtapi.UserSignup, error) {
```
tinakurian, line 243: can't this just be inside of `UpdateUserSignup`? then when you call UpdateUserSignup you could just pass it a userid similar to what's done in the other handlers?

pkg/controller/signup.go
```diff
@@ -60,3 +62,38 @@ func (s *Signup) GetHandler(ctx *gin.Context) {
 		ctx.JSON(http.StatusOK, signupResource)
 	}
 }
+
+// VerifyCodeHandler validates the phone verification code passed in by the user
+func (s *Signup) VerifyCodeHandler(ctx *gin.Context) {
+	code := ctx.Param("code")
```
tinakurian, line 68: if this code is empty, maybe we should just quit instead of moving ahead to retrieve the usersignup?

pkg/controller/signup.go
```diff
@@ -60,3 +62,38 @@ func (s *Signup) GetHandler(ctx *gin.Context) {
...
+
+	userID := ctx.GetString(context.SubKey)
```
tinakurian, line 70: could we also add "code" to the context used here by any chance?

pkg/controller/signup.go
```diff
@@ -60,3 +62,38 @@ func (s *Signup) GetHandler(ctx *gin.Context) {
...
+	signupResource, err := s.signupService.GetUserSignup(userID)
+	if err != nil {
+		log.Error(ctx, err, "error getting UserSignup resource")
+		errors.AbortWithError(ctx, http.StatusInternalServerError, err, "error getting UserSignup resource")
```
tinakurian, line 74: should this maybe return 404 not found instead of 500 internal server error?

pkg/verification/verification_service.go
```diff
@@ -87,3 +95,60 @@ func generateVerificationCode() (string, error) {
...
+		delete(signup.Annotations, v1alpha1.UserVerificationAttemptsAnnotationKey)
+		delete(signup.Annotations, v1alpha1.UserSignupVerificationCounterAnnotationKey)
+		delete(signup.Annotations, v1alpha1.UserSignupVerificationTimestampAnnotationKey)
+		delete(signup.Annotations, v1alpha1.UserVerificationExpiryAnnotationKey)
+		return nil
+	} else {
+		// If the code doesn't match
+		attemptsMade++
+		signup.Annotations[v1alpha1.UserVerificationAttemptsAnnotationKey] = strconv.Itoa(attemptsMade)
+		return errors2.NewBadRequest("invalid code")
```
tinakurian, line 150: i think bad request more means that the request couldn't be understood: 

```
The request could not be understood by the server due to malformed syntax. The client SHOULD NOT repeat the request without modifications.

https://www.restapitutorial.com/httpstatuscodes.html#
```

in this case, it's more that the user is unauthorized, so maybe we should send back 401 unauthorized?

pkg/verification/verification_service.go
```diff
@@ -87,3 +95,60 @@ func generateVerificationCode() (string, error) {
...
+	// If the code matches then set VerificationRequired to false, reset other verification annotations
+	if code == signup.Annotations[v1alpha1.UserSignupVerificationCodeAnnotationKey] {
+		signup.Spec.VerificationRequired = false
+		delete(signup.Annotations, v1alpha1.UserSignupVerificationCodeAnnotationKey)
+		delete(signup.Annotations, v1alpha1.UserVerificationAttemptsAnnotationKey)
+		delete(signup.Annotations, v1alpha1.UserSignupVerificationCounterAnnotationKey)
+		delete(signup.Annotations, v1alpha1.UserSignupVerificationTimestampAnnotationKey)
+		delete(signup.Annotations, v1alpha1.UserVerificationExpiryAnnotationKey)
+		return nil
+	} else {
```
tinakurian, line 146: i think we can remove this `else` nesting and remove the return nil at the end of the func.

pkg/verification/verification_service.go
```diff
@@ -87,3 +95,60 @@ func generateVerificationCode() (string, error) {
...
+
+	exp, err := time.Parse(TimestampLayout, signup.Annotations[v1alpha1.UserVerificationExpiryAnnotationKey])
+	if err != nil {
+		// If the verification expiry timestamp is corrupt or missing, then return an error
+		return errors2.NewInternalError(err)
+	}
+
+	if now.After(exp) {
+		// If it is now past the expiry timestamp for the verification code, return a 400 Bad Request error
+		return errors2.NewBadRequest("verification code expired")
```
tinakurian, line 134: i think this should also return 401 unauthorized: https://tools.ietf.org/html/rfc6750

pkg/signup/signup_service.go
```diff
@@ -154,6 +173,13 @@ func (s *ServiceImpl) CreateUserSignup(ctx *gin.Context) (*crtapi.UserSignup, er
 	return created, nil
 }
 
+func splitEmail(email string) (account, host string) {
```
tinakurian, line 176: do we need to return the account? seems like it's ignored when we call `splitEmail`

pkg/signup/signup_service.go
tinakurian, line 243: could we add a comment to this? also wondering if we should re-use this func in `GetSignup`

pkg/verification/verification_service_test.go
```diff
@@ -150,6 +159,96 @@ func (s *TestVerificationServiceSuite) TestSendVerifyMessageFails() {
 	require.Empty(s.T(), userSignup.Annotations[v1alpha1.UserSignupVerificationCodeAnnotationKey])
 }
 
+func (s *TestVerificationServiceSuite) TestVerifyCode() {
```
tinakurian, line 162: could we also add tests for the other error conditions? like if the attemps field and the expiry field are corrupt/set incorrectly/not set at all

pkg/signup/signup_service.go
```diff
@@ -213,3 +239,25 @@ func (s *ServiceImpl) GetSignup(userID string) (*Signup, error) {
...
+		if errors.IsNotFound(err) {
+			return nil, nil
+		}
+		return nil, err
+	}
+
+	return userSignup, nil
+}
+
+func (s *ServiceImpl) UpdateUserSignup(userSignup *crtapi.UserSignup) (*crtapi.UserSignup, error) {
```
tinakurian, line 256: same here, could we add a comment maybe describing what this does?

pkg/verification/verification_service.go
```diff
@@ -87,3 +95,60 @@ func generateVerificationCode() (string, error) {
...
+	}
+
+	// If the code matches then set VerificationRequired to false, reset other verification annotations
+	if code == signup.Annotations[v1alpha1.UserSignupVerificationCodeAnnotationKey] {
```
alexeykazakov, line 138: Should we make sure that "code" is not an empty string here?

pkg/verification/verification_service.go
alexeykazakov: > in this case, it's more that the user is unauthorized, so maybe we should send back 401 unauthorized?

We should be careful with 401 errors. The frontend could have a global routing for all 401 errors and redirect users to the login page.
Maybe 403 would be better than 401 here?

pkg/controller/signup.go
tinakurian: i thought at the time of creating context the idea was so that we wouldn't be hard-coding strings when trying to retrieve values from the context, an example is FamilyNameKey = "familyName" for instance..

pkg/verification/verification_service.go
tinakurian: hmm 403 may be a bit misleading: 

```
The server understood the request, but is refusing to fulfill it. Authorization will not help and the request SHOULD NOT be repeated. 
```

is there anyway for us to confirm if there is a 401 catch-all? i guess if we have no other options that's the best we can do..

pkg/controller/signup.go
alexeykazakov: Putting params to the context is a way to pass those params trough the code.
Every secured endpoint requires a user token. So, it makes sense to extract all the information from the token (id, username, email, name, etc) once and put it to the context, so, the endpoints do not have to parse the token individually.
"code" is specific to this verify endpoint. It won't be available in the other endpoints. So, I I don't think we should try to add it to the context for all endpoints.
We could create a new context inside the verify() endpoint and pass this new context with "code" to the verify service but it would be inconsistent with our existing controller/services. Currently our controllers takes care of the endpoint API and calls the corresponding services by passing all needed information directly through function arguments.

pkg/verification/verification_service.go
alexeykazakov: I still have a feeling that 401 can be even more misleading. This endpoint requires a user token. So, the user is authenticated correctly. We could confirm with RHD folks if they use any 401 redirects but we don't know if RHD is the only UI which will use our reg-service backend. We had this issue with 401 redirects previously with OSIO.

pkg/controller/signup.go
tinakurian: oh i see, i missed the part that he was using `param` as in he's retrieving a param from the endpoint /{code}/

pkg/verification/verification_service.go
tinakurian: hmm yeah, 401 may be confusing because the user may not understand whether it's the token or the phone verification that failed and if we know its been an issue in the past than maybe 403 + error message <phone verification token invalid/expired/etc> will help make it clear..

pkg/verification/verification_service.go
tinakurian: see comment above, @alexeykazakov is suggesting 403 to avoid issues with 401 catch-alls

pkg/verification/verification_service.go
tinakurian: i'm wondering if code can even be empty - since it's a path parameter you won't hit the endpoint without it being present, i don't think..

pkg/controller/signup.go
tinakurian: as mentioned in the other comment, maybe code actually can't be empty here since it's a path parameter... if it's empty the endpoint probably wasn't hit..?

sbryzak:
@tinakurian @alexeykazakov thanks for the reviews!  I do have some more work to do to improve coverage, also i'm working on the e2e tests today.

pkg/verification/verification_service.go
alexeykazakov: Yeah but this is the verification service (not the controller). So, you can still pass an empty "code" to his function. I suggest returning an invalid argument error in this case.

pkg/kubeclient/signup.go
```diff
@@ -49,3 +50,19 @@ func (c *userSignupClient) Create(obj *crtapi.UserSignup) (*crtapi.UserSignup, e
 	}
 	return result, err
 }
+
+// Update will update an existing UserSignup resource in the cluster, returning an error if something went wrong
+func (c *userSignupClient) Update(obj *crtapi.UserSignup) (*crtapi.UserSignup, error) {
+	result := &crtapi.UserSignup{}
+	err := c.client.Put().
```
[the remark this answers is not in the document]
sbryzak: I've read up on patch for the Kubernetes api, and it seems that the request has to be very specific in what is being patched.  Considering that we have a number of variations on what we might need to patch depending on the conditions, this doesn't seem practical.

pkg/verification/verification_service.go
sbryzak: I struggled to find a suitable response code for this case.  Bad Request was the best I could really come up with, however I'm still open to alternate suggestions.

pkg/verification/verification_service.go
sbryzak: Good idea, I'll fix this.

pkg/signup/signup_service.go
sbryzak: I refactored this slightly.

pkg/signup/signup_service.go
sbryzak: I've added a comment.  Using this in `GetSignup` doesn't seem to save us much, I think it would be about the same lines of code.

pkg/signup/signup_service.go
sbryzak: Comment added.

pkg/verification/verification_service_test.go
sbryzak: I've added a bunch of additional tests, I think all scenarios are now covered.

pkg/controller/signup.go
```diff
@@ -60,3 +62,38 @@ func (s *Signup) GetHandler(ctx *gin.Context) {
...
+		log.Error(ctx, err, "error validating user verification code")
+		switch t := err.(type) {
+		default:
+			errors.AbortWithError(ctx, http.StatusInternalServerError, err, "error while verifying code")
+		case *errors2.StatusError:
+			errors.AbortWithError(ctx, int(t.ErrStatus.Code), err, t.ErrStatus.Message)
+		}
+	}
+
+	_, err = s.signupService.UpdateUserSignup(signupResource)
```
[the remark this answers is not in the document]
sbryzak: I've had to rework this section a little bit.  I've added a comment with the details.

pkg/signup/signup_service.go
```diff
@@ -125,6 +132,17 @@ func (s *ServiceImpl) CreateUserSignup(ctx *gin.Context) (*crtapi.UserSignup, er
 		}
 	}
 
+	verificationRequired := s.Config.GetVerificationEnabled()
+
+	// Check if the user's email address is in the list of domains excluded for phone verification
+	_, emailHost := splitEmail(userEmail)
+	for _, d := range s.Config.GetVerificationExcludedEmailDomains() {
+		if d == emailHost {
```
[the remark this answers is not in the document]
sbryzak: Modified to case insensitive compare.

pkg/verification/verification_service.go
sbryzak: In any case I've added an empty check to the controller.

pkg/configuration/configuration.go
```diff
@@ -301,6 +306,15 @@ func (c *Registry) GetVerificationMessageTemplate() string {
 	return c.v.GetString(varVerificationMessageTemplate)
 }
 
+// GetVerificationExcludedEmailDomains returns the list of email address domains for which phone verification
+// is not required
+func (c *Registry) GetVerificationExcludedEmailDomains() []string {
+	if c.excludedDomains == nil {
```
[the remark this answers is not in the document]
sbryzak: Good point, any suggestions as to the best way to approach this?

pkg/verification/verification_service.go
sbryzak: Changed to 403 Forbidden

pkg/verification/verification_service.go
sbryzak: On further reflection I do think that 403 is a better option than 400.  I'll update the code to that for now.

pkg/signup/signup_service.go
tinakurian: yeah, just thought it would help with removing redundancy..

pkg/controller/signup.go
```diff
@@ -66,19 +66,37 @@ func (s *Signup) GetHandler(ctx *gin.Context) {
 // VerifyCodeHandler validates the phone verification code passed in by the user
 func (s *Signup) VerifyCodeHandler(ctx *gin.Context) {
 	code := ctx.Param("code")
+	if code == "" {
+		log.Error(ctx, nil, "no code provided in request")
+		ctx.AbortWithStatus(http.StatusNotFound)
```
alexeykazakov, line 71: Shouldn't it be 400? And we should also return.
```suggestion
		ctx.AbortWithStatus(http.StatusBadRequest)
		return
```

pkg/controller/signup.go
```diff
@@ -66,19 +66,37 @@ func (s *Signup) GetHandler(ctx *gin.Context) {
...
 		log.Errorf(ctx, nil, "UserSignup resource for userID: %s resource not found", userID)
 		ctx.AbortWithStatus(http.StatusNotFound)
+		return
 	}
 
 	err = s.verificationService.VerifyCode(ctx, signupResource, code)
+
+	// Regardless of whether the VerifyCode() call returns an error or not, we need to update the UserSignup instance
+	// as its state can be updated even in the case of an error.  This may result in the slight possibility that any
+	// errors returned by VerifyCode() are suppressed, as error handling for the UserSignup update is given precedence.
```
alexeykazakov, line 93: Let's at least log the error returned by VerifyCode() if the update failed too.

pkg/verification/verification_service.go
```diff
@@ -130,8 +133,11 @@ func (s *ServiceImpl) VerifyCode(ctx *gin.Context, signup *v1alpha1.UserSignup,
 	}
 
 	if now.After(exp) {
-		// If it is now past the expiry timestamp for the verification code, return a 400 Bad Request error
-		return errors2.NewBadRequest("verification code expired")
+		// If it is now past the expiry timestamp for the verification code, return a 403 Forbidden error
+		return errors2.NewForbidden(schema.GroupResource{
```
alexeykazakov, line 137: I'm curious, why do you return a kube error? Wouldn't some http error be more suitable?

pkg/verification/verification_service.go
```diff
@@ -143,12 +149,13 @@ func (s *ServiceImpl) VerifyCode(ctx *gin.Context, signup *v1alpha1.UserSignup,
...
-		attemptsMade++
-		signup.Annotations[v1alpha1.UserVerificationAttemptsAnnotationKey] = strconv.Itoa(attemptsMade)
-		return errors2.NewBadRequest("invalid code")
 	}
 
-	return nil
+	// The code doesn't match
+	attemptsMade++
+	signup.Annotations[v1alpha1.UserVerificationAttemptsAnnotationKey] = strconv.Itoa(attemptsMade)
+	return errors2.NewForbidden(schema.GroupResource{
```
alexeykazakov, line 157: Same question here.

pkg/controller/signup_test.go
```diff
@@ -212,9 +217,335 @@ func (s *TestSignupSuite) TestSignupGetHandler() {
...
+		ctx.Params = append(ctx.Params, gin.Param{
+			Key:   "code",
+			Value: "999888",
+		})
+
+		handler(ctx)
+
+		// Check the status code is what we expect.
+		require.Equal(s.T(), http.StatusOK, rr.Code)
+	})
```
alexeykazakov, line 279: Let's also check that the fake/mock verification and signup update services received the expected UserID and UserSignup as parameters.

pkg/controller/signup_test.go
```diff
@@ -212,9 +217,335 @@ func (s *TestSignupSuite) TestSignupGetHandler() {
...
+		ctx.Request = req
+		ctx.Set(context.SubKey, userID)
+		ctx.Params = append(ctx.Params, gin.Param{
+			Key:   "code",
+			Value: "111233",
+		})
+
+		handler(ctx)
+
+		// Check the status code is what we expect.
```
alexeykazakov, line 347: Lines 332-345 are mostly duplicated in many tests. Can we maybe a bit optimize it by creating a re-usable function. Something like:
ctx, rr := verify(verifyCtr,code int)

pkg/controller/signup_test.go
```diff
@@ -212,9 +217,335 @@ func (s *TestSignupSuite) TestSignupGetHandler() {
...
+							crtapi.UserSignupUserEmailAnnotationKey:        "jsmith@redhat.com",
+							crtapi.UserVerificationAttemptsAnnotationKey:   "0",
+							crtapi.UserSignupVerificationCodeAnnotationKey: "222222",
+							crtapi.UserVerificationExpiryAnnotationKey:     time.Now().Add(10 * time.Second).Format(verification.TimestampLayout),
+						},
+					},
+					Spec:   crtapi.UserSignupSpec{},
+					Status: crtapi.UserSignupStatus{},
+				}, nil
+			},
```
alexeykazakov, line 479: I wonder if can also improve the code readability by moving the UserSignup creation to a re-usable function too...

pkg/controller/signup_test.go
```diff
@@ -212,9 +217,335 @@ func (s *TestSignupSuite) TestSignupGetHandler() {
...
+		ctx.Set(context.SubKey, userID)
+		ctx.Params = append(ctx.Params, gin.Param{
+			Key:   "code",
+			Value: "222222",
+		})
+
+		handler(ctx)
+
+		// Check the status code is what we expect.
+		require.Equal(s.T(), http.StatusInternalServerError, rr.Code)
```
alexeykazakov, line 513: Do we just return a 500 error here and in the otter places or we do return some error details as a JSON so we could check the detailed error message too?

Just want to make sure we actually catch the expected (and different) errors and not the same 500 in all these tests.

pkg/signup/signup_service_test.go
```diff
@@ -403,6 +437,89 @@ func (s *TestSignupServiceSuite) TestGetSignupUnknownStatus() {
...
+		require.Equal(s.T(), us.Name, val.Name)
+	})
+
+	s.Run("getusersignup returns error", func() {
+		fakeClient.MockGet = func(s string) (userSignup *v1alpha1.UserSignup, e error) {
+			return nil, errors.New("get failed")
+		}
+
+		val, err := svc.GetUserSignup("foo")
+		require.Error(s.T(), err)
```
alexeykazakov, line 459: Check the error message

pkg/signup/signup_service_test.go
```diff
@@ -403,6 +437,89 @@ func (s *TestSignupServiceSuite) TestGetSignupUnknownStatus() {
...
+		require.Nil(s.T(), val)
+	})
+
+	s.Run("getusersignup returns not found error", func() {
```
alexeykazakov, line 463: Well... technically it doesn't return an error. Which is expected:)

pkg/signup/signup_service_test.go
```diff
@@ -403,6 +437,89 @@ func (s *TestSignupServiceSuite) TestGetSignupUnknownStatus() {
...
+	s.Run("updateusersignup returns error", func() {
+		fakeClient.MockUpdate = func(userSignup2 *v1alpha1.UserSignup) (userSignup *v1alpha1.UserSignup, e error) {
+			return nil, errors.New("update failed")
+		}
+
+		val, err := svc.GetUserSignup(us.Name)
+		require.NoError(s.T(), err)
+
+		updated, err := svc.UpdateUserSignup(val)
+		require.Error(s.T(), err)
```
alexeykazakov, line 500: Check the error message

pkg/verification/verification_service_test.go
```diff
@@ -150,6 +159,211 @@ func (s *TestVerificationServiceSuite) TestSendVerifyMessageFails() {
...
+				Username: "sbryzak@redhat.com",
+			},
+		}
+
+		err := svc.VerifyCode(ctx, userSignup, "123456")
+		require.Error(s.T(), err)
+		require.Equal(s.T(), "Internal error occurred: parsing time \"ABC\" as \"2006-01-02T15:04:05.000Z07:00\": cannot parse \"ABC\" as \"2006\"", err.Error())
+	})
+}
+
```
alexeykazakov, line 366: I guess we could move some code here to reusable functions too?

pkg/controller/signup.go
sbryzak: Updated.

pkg/verification/verification_service.go
sbryzak: Good point.  For some reason I was thinking these were gin errors, however it turns out there are no third party http errors besides the kube ones.  For this reason I've added a few new error constructors in the error package.

pkg/verification/verification_service.go
sbryzak: As above.

pkg/controller/signup.go
sbryzak: Done.

pkg/controller/signup_test.go
sbryzak: This would be dependent on what kind of error the verification service returns.   With the mock `VerifyCode` function we get the following body:

```
{"status":"Internal Server Error","code":500,"message":"some other error","details":"error while verifying code"}
```

pkg/signup/signup_service_test.go
sbryzak: Done.

pkg/signup/signup_service_test.go
sbryzak: I've modified the description.

pkg/signup/signup_service_test.go
sbryzak: Done

pkg/controller/signup_test.go
sbryzak: Done.

pkg/controller/signup_test.go
sbryzak: Done

pkg/controller/signup_test.go
sbryzak: We're not creating that many UserSignups, also I like the readability of being able to see the entire object we're creating here rather than putting it in a black box function.

pkg/verification/verification_service_test.go
sbryzak: See my other comment above.  Also, there's quite a variation of UserSignups that are created, some of them only need the email and phone number, others need all of the verification annotations.

pkg/configuration/configuration.go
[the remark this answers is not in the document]
sbryzak: Added a lock here for thread safety.

pkg/kubeclient/signup.go
```diff
@@ -49,3 +50,19 @@ func (c *userSignupClient) Create(obj *crtapi.UserSignup) (*crtapi.UserSignup, e
 	}
 	return result, err
 }
+
+// Update will update an existing UserSignup resource in the cluster, returning an error if something went wrong
+func (c *userSignupClient) Update(obj *crtapi.UserSignup) (*crtapi.UserSignup, error) {
```
tinakurian, line 55: i think we need to add permissions for `update` to actually work in the deployment?

pkg/verification/verification_service_test.go
```diff
@@ -150,6 +158,211 @@ func (s *TestVerificationServiceSuite) TestSendVerifyMessageFails() {
...
+				Name:      "123",
+				Namespace: "test",
+				Annotations: map[string]string{
+					v1alpha1.UserSignupUserEmailAnnotationKey:        "sbryzak@redhat.com",
+					v1alpha1.UserVerificationAttemptsAnnotationKey:   "0",
+					v1alpha1.UserSignupVerificationCodeAnnotationKey: "123456",
+					v1alpha1.UserVerificationExpiryAnnotationKey:     now.Add(10 * time.Second).Format(verification.TimestampLayout),
+				},
+				Labels: map[string]string{
+					v1alpha1.UserSignupPhoneNumberLabelKey: "+1NUMBER",
```
tinakurian, line 182: did we update anything to add this phone number field when a usersignup is created? what happens if phone verification is required and no phone number exists? just a thought after testing on a temp cluster. the usersignup is created for me without a phonenumber, hence sending a verification fails unless i patch my usersignup to contain my phone number..

sbryzak: The user will provide their phone number when they initiate phone verification.  It won't be present in the UserSignup until then.

alexeykazakov: Yeah.. we will need to cover the entire scenario in our e2e tests when everything is ready.

pkg/kubeclient/signup.go
alexeykazakov: Good catch @tinakurian! We do need to add `update` to https://github.com/codeready-toolchain/registration-service/blob/master/deploy/registration-service.yaml#L27

pkg/configuration/configuration.go
```diff
@@ -310,6 +313,9 @@ func (c *Registry) GetVerificationMessageTemplate() string {
 // is not required
 func (c *Registry) GetVerificationExcludedEmailDomains() []string {
 	if c.excludedDomains == nil {
+		lock.Lock()
```
alexeykazakov, line 316: Unfortunately it's not enough. We can't read and write the same variable in two different goroutines in the same time. And with this approach it's possible for one goroutine to check `if c.excludedDomains == nil` while the other one is doing `c.excludedDomains = strings.Split(c.v.GetString(varVerificationExcludedEmailDomains), ",")`
We should use RWMutex instead. But I think it will be simpler just to initialize `c.excludeDomains` in `New()` or in `CreateEmptyRegistry()`

pkg/controller/signup_test.go
```diff
@@ -276,6 +284,15 @@ func (s *TestSignupSuite) TestVerifyCodeHandler() {
 
 		// Check the status code is what we expect.
 		require.Equal(s.T(), http.StatusOK, rr.Code)
+
+		// Check that the correct userID is passed into the FakeSignupService
+		require.Equal(s.T(), userID, storedUserID)
+
+		// Check that the correct UserSignup is passed into the FakeSignupService
+		require.Equal(s.T(), userID, storedUserSignup.Name)
```
alexeykazakov, line 292: We are supposed to update the UserSignup using this FakeSignupService, right? So, just checking the name is not enough. We need to compare the entire UserSignup here. Basically we should make sure that the `Update()` receives the same UserSignup object as `Get()` returned with expected changes (annotations + verificationRequired field in the spec or whatever we update).

pkg/controller/signup_test.go
alexeykazakov: > This would be dependent on what kind of error the verification service returns. With the mock VerifyCode function we get the following body:

So, let's check that we do receive this error json.

pkg/kubeclient/signup.go
sbryzak: Done

tinakurian: @sbryzak friendly reminder - don't forget to also add it in the host-operator `registration-service.yaml`

pkg/configuration/configuration.go
sbryzak: I've put the lock() around the whole function and modified the signup service (which uses the excluded domains value) so that it will only read it once.

alexeykazakov: While it's probably not super critical in this particular case it's still a bad pattern to lock the entire function. We are basically making the entire verification service a single threaded. Let's avoid that.

alexeykazakov: Ah.. I think I got what you mean. You are modifying the configuration usage too. So, the service won't read the configuration every time. It solves the problem with multi-threading but error prone IMO. We will need to remember that we are not supposed to read this particular configuration as we do for the other configuration. Why not just initiate this configuration once in the constructor?

pkg/controller/signup_test.go
sbryzak: I've added the checks to confirm all of the set fields, however there is only one UserSignup in play here so of course all the fields will be identical.

alexeykazakov: It doesn't matter how many UserSignups in play here, does it?
If I understand the flow correctly, without that check we could miss a bug like this:
T1: signup := getUserSignup()
T2: VerifyCode(signup) // <- signup is modified here!
T3: signup.Spec.RecauiredVerification = falce // <- bug!!! We should not modify the signup here or use any other signup.
T4: UpdateSignup(signup) // <- we should make sure that the signup here is the same which is returned by VerifyCode() without any further modification.

But if we don't really can test it properly if our mock VerificationCode() doesn't actually modify the signup.

pkg/controller/signup_test.go
sbryzak: Done.

pkg/controller/signup_test.go
sbryzak: I've added a small modification to the UserSignup within the mock verification service, which the test now checks for.

pkg/configuration/configuration.go
sbryzak: Code moved to constructor.

PR #125, opened 2020-08-24
Added phone verification UI.

michaelkleinhenz:
This adds the phone verification UI to the registration service web frontend. 

Implements CRT-768 and CRT-769.


pkg/assets/landingpage.js
```diff
@@ -98,7 +113,10 @@ function getSignupState(cbSuccess, cbError) {
 // updates the signup state.
 function updateSignupState() {
   getSignupState(function(data) {
-    if (data.status.ready === true) {
+    if (data.status.verificationRequired) {
```
tinakurian, line 116: i think this would need to be something like: 
` if (data.status.ready === false && data.status.reason == 'VerificationRequired') {
     hideAll()
     show('state-initiate-phone-verification')
}
`

from the console, it looks like the this is what the state is: 
image

before making this change, all i could see in the UI was the spinner and the text: in the UI it continues to spin saying "your codeready toolchain account is waiting approval" - this was after patching the usersignup to require verification and also approving it. I guess it was falling into the `else` clause at the end.. After changing this line above, i'm able to see the request for the `country code` and `phone number` as seen below: 

image (1)

michaelkleinhenz: Ok, I guessed the response from the source. Maybe this changed since :-) Updated the PR, see 099b471.

tinakurian (approved):

PR #126, opened 2020-08-26
Initiate verification endpoint

tinakurian:
create initiate verification endpoint. this is a post endpoint which requires a body containing phone_number and country_code values: {   "country_code":"61",   "phone_number":"0411222333" }

jira: https://issues.redhat.com/browse/CRT-765
e2e-tests are in progress

#### Text message received via endpoint: 
IMG_2590

------------------------------
#### screenshot of postman used to invoke endpoint:
image


pkg/signup/signup_service.go
```diff
@@ -238,6 +242,45 @@ func (s *ServiceImpl) GetSignup(userID string) (*Signup, error) {
...
+	if annotationCounter != "" {
+		counter, err = strconv.Atoi(annotationCounter)
+		if err != nil {
+			return nil, http.StatusInternalServerError, err
+		}
+	}
+
+	// check if counter has exceeded the limit of daily limit - if at limit error out
+	if counter > dailyLimit {
+		return nil, http.StatusForbidden, errors2.New("daily limit has been exceeded")
```
alexeykazakov, line 266: Let's log the current counter and the limit.

pkg/controller/signup.go
```diff
@@ -46,6 +47,50 @@ func (s *Signup) PostHandler(ctx *gin.Context) {
...
+	}
+
+	err = s.verificationService.SendVerification(ctx, userSignup)
+	if err != nil {
+		log.Errorf(ctx, nil, "Verification for %s could not be sent", userID)
+		ctx.AbortWithError(http.StatusInternalServerError, err)
+		return
+	}
+
+	log.Infof(ctx, "phone verification has passed for userID %s", userID)
```
alexeykazakov, line 90: This method does not verify the code, does it? It just sends it to the user.

pkg/signup/signup_service.go
```diff
@@ -238,6 +242,45 @@ func (s *ServiceImpl) GetSignup(userID string) (*Signup, error) {
 	return signupResponse, nil
 }
 
+// PostVerification returns userSignup resource from the host cluster.
+// Returns the usersignup, an http code and an error if present.
+func (s *ServiceImpl) PostVerification(dailyLimit int, responseBody map[string]string, userID, code string) (*crtapi.UserSignup, int, error) {
```
alexeykazakov, line 247: The method name is confusing IMO. UpdateWithVerificationCode()?

Also, why you send the entire requestBody (it's not response body btw as it's named now) from the controller to the service. Why not extracting the phone number from the request in the controller and send it to the service as a parameter? It should not return any http statuses either.

Controllers deal with our REST API and HTTP. Services implement business logic and transport/api agnostic.

tinakurian: you're right, it shouldn't be called POST and the endpoint shouldn't be a "POST" either. it should really be a PUT, since that's what we're doing... maybe this should be changed?

pkg/signup/signup_service.go
```diff
@@ -238,6 +242,45 @@ func (s *ServiceImpl) GetSignup(userID string) (*Signup, error) {
...
+	// Retrieve UserSignup resource from the host cluster
+	userSignup, err := s.UserSignups.Get(userID)
+	if err != nil {
+		return nil, http.StatusNotFound, err
+	}
+
+	// get the annotation counter
+	annotationCounter := userSignup.Annotations[crtapi.UserSignupVerificationCounterAnnotationKey]
```
sbryzak, line 255: This logic should be in the verification service.

pkg/signup/signup_service.go
```diff
@@ -238,6 +242,45 @@ func (s *ServiceImpl) GetSignup(userID string) (*Signup, error) {
...
+	// check if counter has exceeded the limit of daily limit - if at limit error out
+	if counter > dailyLimit {
+		return nil, http.StatusForbidden, errors2.New("daily limit has been exceeded")
+	}
+
+	// set the usersignup annotations
+	userSignup.Annotations[crtapi.UserSignupVerificationCounterAnnotationKey] = strconv.Itoa(counter + 1)
+	userSignup.Annotations[crtapi.UserSignupPhoneNumberLabelKey] = responseBody["country_code"] + responseBody["phone_number"]
+	userSignup.Annotations[crtapi.UserSignupVerificationCodeAnnotationKey] = code
+	userSignup.Annotations[crtapi.UserSignupVerificationTimestampAnnotationKey] = time.Now().Format(verification.TimestampLayout)
```
sbryzak, line 273: All of these verification-related annotations should be managed by the verification service.

pkg/signup/signup_service.go
sbryzak: I don't think we need this method at all, the UpdateUserSignup() function should be all we need from this service.

sbryzak (changes requested):
There seems to be a couple of things still missing from the workflow:

- Checking that the phone number the user provides isn't already in use by another user.
- Checking that the phone number hasn't been banned.

I would also take the opportunity in this PR to address CRT-800, and take care of hashing the user's phone number here.  The design as it stands is quite a departure from what was established in the verify code endpoint.  The verification service should manage all verification-related state; it should be a black box that is simply passed in the current UserSignup and which takes care of the entire verification process.  The different services are there to provide a clean separation of concerns, so there shouldn't be any other verification code in other places.  In fact, if anything I would even change the verify code function to load/save the usersignup itself via the signup service, rather than have the controller do it.  That way the controller is solely responsible for unmarshalling the values from the request, and passing them to the verification service which then does all the work.

tinakurian:
> There seems to be a couple of things still missing from the workflow:
> 
> * Checking that the phone number the user provides isn't already in use by another user.
> * Checking that the phone number hasn't been banned.
> 
> I would also take the opportunity in this PR to address CRT-800, and take care of hashing the user's phone number here. The design as it stands is quite a departure from what was established in the verify code endpoint. The verification service should manage all verification-related state; it should be a black box that is simply passed in the current UserSignup and which takes care of the entire verification process. The different services are there to provide a clean separation of concerns, so there shouldn't be any other verification code in other places. In fact, if anything I would even change the verify code function to load/save the usersignup itself via the signup service, rather than have the controller do it. That way the controller is solely responsible for unmarshalling the values from the request, and passing them to the verification service which then does all the work.

@sbryzak thanks, i didn't realize i needed to check those, makes sense. will add this!

I didn't think that the usersignup should be modified in the verification service. the other functions in signup_service handle modifying/creating the usersignup resource, so i thought i was following the same design that was already there. this is currently how the create function is implemented. the usersignup is created in the signup_service and the verification functions are just called as a service.

pkg/signup/signup_service.go
tinakurian: Hm, i don't really understand this flow in that case? The verifyCodeHandler calls UpdateUserSignup() - how would this work?

pkg/signup/signup_service.go
tinakurian: hmm not sure i understand why? the signup resource is created/modified in the signup_service. by doing this all in the verification service we are sort of bringing in some inconsistencies.

pkg/controller/signup.go
```diff
@@ -1,19 +1,20 @@
...
-
 	"github.com/codeready-toolchain/registration-service/pkg/configuration"
 	"github.com/codeready-toolchain/registration-service/pkg/context"
 	"github.com/codeready-toolchain/registration-service/pkg/errors"
 	"github.com/codeready-toolchain/registration-service/pkg/log"
 	"github.com/codeready-toolchain/registration-service/pkg/signup"
+	"github.com/codeready-toolchain/registration-service/pkg/verification"
 
 	"github.com/gin-gonic/gin"
+	errors2 "k8s.io/apimachinery/pkg/api/errors"
```
xcoulon, line 17: `apierrors` may be a better alias than `errors2`

pkg/controller/signup.go
```diff
@@ -46,6 +47,51 @@ func (s *Signup) PostHandler(ctx *gin.Context) {
...
+// UpdateVerificationHandler stars the verification process and updates a usersignup resource
+func (s *Signup) UpdateVerificationHandler(ctx *gin.Context) {
+	userID := ctx.GetString(context.SubKey)
+	signup, err := s.signupService.GetUserSignup(userID)
+	if err != nil {
+		if errors2.IsNotFound(err) {
+			log.Error(ctx, err, "usersignup not found")
+			errors.AbortWithError(ctx, http.StatusNotFound, err, "usersignup not found")
+		}
+		log.Error(ctx, err, "error retreiving usersignup")
```
xcoulon, line 59:
```suggestion
		log.Error(ctx, err, "error retrieving usersignup")
```

pkg/controller/signup.go
```diff
@@ -46,6 +47,51 @@ func (s *Signup) PostHandler(ctx *gin.Context) {
...
+		errors.AbortWithError(ctx, http.StatusInternalServerError, err, fmt.Sprintf("error retreiving usersignup: %s", userID))
```
xcoulon, line 60:
```suggestion
		errors.AbortWithError(ctx, http.StatusInternalServerError, err, fmt.Sprintf("error retrieving usersignup: %s", userID))
```

pkg/controller/signup.go
```diff
@@ -46,6 +47,51 @@ func (s *Signup) PostHandler(ctx *gin.Context) {
...
+		return
+	}
+
+	// Read the Body content
+	var bodyBytes []byte
+	if ctx.Request.Body != nil {
+		bodyBytes, _ = ioutil.ReadAll(ctx.Request.Body)
+	}
+
```
xcoulon, line 69:
```suggestion
	defer func() {
		if err := res.Body.Close(); err != nil {
			log.Error(nil, err, "failed to close response after reading")
		}
	}
```

xcoulon: you also need wrap the call `res.Body.Close()` in a deferred func

pkg/controller/signup.go
```diff
@@ -73,26 +119,25 @@ func (s *Signup) VerifyCodeHandler(ctx *gin.Context) {
...
 
 	// The VerifyCode() call here MAY make changes to the specified signupResource
-	err = s.verificationService.VerifyCode(ctx, signupResource, code)
+	signup, err = s.verificationService.VerifyCode(signup, code)
 
 	// Regardless of whether the VerifyCode() call returns an error or not, we need to update the UserSignup instance
 	// as its state can be updated even in the case of an error.  This may result in the slight possibility that any
 	// errors returned by VerifyCode() are suppressed, as error handling for the UserSignup update is given precedence.
-	_, err2 := s.signupService.UpdateUserSignup(signupResource)
+	_, err2 := s.signupService.UpdateUserSignup(signup)
```
xcoulon, line 140: There is no need for a second `err2` var: you could use `err` again here, couldn't you?

pkg/controller/signup_test.go
```diff
@@ -526,6 +686,7 @@ type FakeSignupService struct {
 	MockCreateUserSignup func(ctx *gin.Context) (*crtapi.UserSignup, error)
 	MockGetUserSignup    func(userID string) (*crtapi.UserSignup, error)
 	MockUpdateUserSignup func(userSignup *crtapi.UserSignup) (*crtapi.UserSignup, error)
+	//MockUpdateWithVerificationCode func(dailyLimit int, responseBody map[string]string, userID, code string) (*crtapi.UserSignup, error)
```
xcoulon, line 689: I believe you can remove this commented out line if it's not useful anymore

pkg/verification/verification_service.go
```diff
@@ -57,28 +58,65 @@ func NewVerificationService(cfg ServiceConfiguration, opts ...VerificationServic
...
+	}
+
+	dailyLimit := s.config.GetVerificationDailyLimit()
+	counter, err = strconv.Atoi(annotationCounter)
 	if err != nil {
-		return err
+		// We shouldn't get an error here, but if we do, we should probably set verification attempts to daily limit
+		// so that we at least now have a valid value
+		signup.Annotations[v1alpha1.UserSignupVerificationCounterAnnotationKey] = strconv.Itoa(dailyLimit)
+		return signup, errors3.NewInternalError(err, fmt.Sprintf("error when retrieving counter annotation for UserSignup %s, set to daily limit", signup.GetName()))
```
xcoulon, line 87: the `errors3 "github.com/codeready-toolchain/registration-service/pkg/errors"` alias should be removed. `errors` (ie, default pkg name) is enough here

pkg/verification/verification_service.go
```diff
@@ -57,28 +58,65 @@ func NewVerificationService(cfg ServiceConfiguration, opts ...VerificationServic
...
+	}
+
+	// check if counter has exceeded the limit of daily limit - if at limit error out
+	if counter > dailyLimit {
+		err := errors3.NewForbiddenError("daily limit exceeded", fmt.Sprintf("%d attempts made. the daily limit of %d has been exceeded", counter, dailyLimit))
+		log.Error(ctx, err, fmt.Sprintf("%d attempts made. the daily limit of %d has been exceeded", counter, dailyLimit))
```
xcoulon, line 93: I believe there is a bit of overlap in the error messages, isn't it? How is it going to appear in the logs?

pkg/controller/signup.go
```diff
@@ -46,6 +47,50 @@ func (s *Signup) PostHandler(ctx *gin.Context) {
 	ctx.Writer.WriteHeaderNow()
 }
 
+// PostVerificationHandler creates a verification and updates a usersignup resource
```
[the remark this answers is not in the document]
tinakurian: fixed as part of: https://github.com/codeready-toolchain/registration-service/pull/126/commits/47d3acf94777a662549db066dbb691caf22852f3

pkg/controller/signup.go
```diff
@@ -46,6 +47,50 @@ func (s *Signup) PostHandler(ctx *gin.Context) {
...
+	code, err := s.verificationService.GenerateVerificationCode()
+	if err != nil {
+		log.Errorf(ctx, nil, "verification code could not be generated")
+		ctx.AbortWithError(http.StatusInternalServerError, err)
+		return
+	}
+
+	userSignup, httpCode, err := s.signupService.PostVerification(s.config.GetVerificationDailyLimit(), m, userID, code)
+	if err != nil {
+		log.Errorf(ctx, nil, "phone verification has failed: %s", err.Error())
```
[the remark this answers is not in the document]
tinakurian: fixed as part of: https://github.com/codeready-toolchain/registration-service/pull/126/commits/47d3acf94777a662549db066dbb691caf22852f3

pkg/controller/signup.go
tinakurian: fixed here: https://github.com/codeready-toolchain/registration-service/blob/60ca7377aabe427ed8f8e3d642380b1b8396454e/pkg/controller/signup.go#L113

pkg/signup/signup_service.go
tinakurian: this has been fixed here: https://github.com/codeready-toolchain/registration-service/blob/60ca7377aabe427ed8f8e3d642380b1b8396454e/pkg/controller/signup.go#L51

pkg/signup/signup_service.go
tinakurian: fixed here: https://github.com/codeready-toolchain/registration-service/blob/60ca7377aabe427ed8f8e3d642380b1b8396454e/pkg/controller/signup.go#L51

pkg/signup/signup_service.go
tinakurian: fixed here: https://github.com/codeready-toolchain/registration-service/blob/60ca7377aabe427ed8f8e3d642380b1b8396454e/pkg/controller/signup.go#L51

pkg/controller/signup.go
tinakurian: fixed as part of: 60ca7377aabe427ed8f8e3d642380b1b8396454e

pkg/controller/signup.go
tinakurian: fixed as part of: 60ca7377aabe427ed8f8e3d642380b1b8396454e

pkg/controller/signup.go
tinakurian: fixed as part of: 60ca7377aabe427ed8f8e3d642380b1b8396454e

pkg/verification/verification_service.go
tinakurian: fixed as part of: https://github.com/codeready-toolchain/registration-service/pull/126/commits/60ca7377aabe427ed8f8e3d642380b1b8396454e

pkg/verification/verification_service.go
tinakurian: fixed as part of: https://github.com/codeready-toolchain/registration-service/pull/126/commits/60ca7377aabe427ed8f8e3d642380b1b8396454e

pkg/controller/signup.go
tinakurian: so the reason why we are doing this is because we are temporarily ignoring the errors in err in favour of updating the usersignup. if the update fails we will return err2, if not and there was an err we will return that. It's described in this comment above: 
```
// Regardless of whether the VerifyCode() call returns an error or not, we need to update the UserSignup instance
// as its state can be updated even in the case of an error.  This may result in the slight possibility that any
// errors returned by VerifyCode() are suppressed, as error handling for the UserSignup update is given precedence.
```

i was thinking that what we could do instead is return a array of json errors, that way we aren't skipping or missing any errors that occur.. WDYT?

pkg/controller/signup_test.go
tinakurian: removed here: https://github.com/codeready-toolchain/registration-service/blob/60ca7377aabe427ed8f8e3d642380b1b8396454e/pkg/controller/signup_test.go#L681-L686

pkg/controller/signup.go
tinakurian: fixed here: 7908cfab4b879d2a292e2d8dafb882a6bbe3b967 and here: https://github.com/codeready-toolchain/registration-service/blob/7908cfab4b879d2a292e2d8dafb882a6bbe3b967/pkg/controller/signup.go#L70-L85

pkg/signup/signup_service.go
tinakurian: fixed here: https://github.com/codeready-toolchain/registration-service/blob/7908cfab4b879d2a292e2d8dafb882a6bbe3b967/pkg/verification/verification_service.go#L103

pkg/verification/verification_service.go
```diff
@@ -61,10 +62,7 @@ func NewVerificationService(cfg ServiceConfiguration, opts ...VerificationServic
 // specified UserSignup parameter, setting the verification code annotation label to the value of the
 // generated verification code
 func (s *ServiceImpl) SendVerification(ctx *gin.Context, signup *v1alpha1.UserSignup) error {
```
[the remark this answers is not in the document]
tinakurian: this was re-purposed based on discussions with @sbryzak to be `InitVerification` updated the description here: https://github.com/codeready-toolchain/registration-service/blob/7908cfab4b879d2a292e2d8dafb882a6bbe3b967/pkg/verification/verification_service.go#L63-L64

is this okay?

pkg/signup/signup_service.go
```diff
@@ -240,25 +242,42 @@ func (s *ServiceImpl) GetSignup(userID string) (*Signup, error) {
...
+	}
+	if len(bannedUserList.Items) > 0 {
+		return errors3.NewForbiddenError("cannot re-register with phone number", "phone number already in use")
+	}
+
+	userSignupList, err := s.UserSignups.ListByPhoneNumber(countryCode + phoneNumber)
+	if err != nil {
+		return errors3.NewInternalError(err, "failed listing userSignups")
+	}
+	if len(userSignupList.Items) > 0 {
```
sbryzak, line 278: Where do we check if the UserSignup found is not the user that is currently verifying?

pkg/verification/verification_service.go
```diff
@@ -57,28 +60,72 @@ func NewVerificationService(cfg ServiceConfiguration, opts ...VerificationServic
...
+		// Set a new timestamp
+		signup.Annotations[v1alpha1.UserSignupVerificationInitTimestampAnnotationKey] = now.Format(TimestampLayout)
+		signup.Annotations[v1alpha1.UserSignupVerificationCounterAnnotationKey] = "0"
+	}
+
+	// get the annotation counter
+	annotationCounter := signup.Annotations[v1alpha1.UserSignupVerificationCounterAnnotationKey]
+	var counter int
+	if annotationCounter == "" {
+		counter = 0
```
sbryzak, line 88: Does `counter` need to be set here?  It seems that in just a few lines down (line 92) it will be overridden anyway.

pkg/verification/verification_service.go
```diff
@@ -57,28 +60,72 @@ func NewVerificationService(cfg ServiceConfiguration, opts ...VerificationServic
...
+	counter, err = strconv.Atoi(annotationCounter)
+	if err != nil {
+		// We shouldn't get an error here, but if we do, we should probably set verification attempts to daily limit
+		// so that we at least now have a valid value
+		signup.Annotations[v1alpha1.UserSignupVerificationCounterAnnotationKey] = strconv.Itoa(dailyLimit)
+		return signup, errors.NewInternalError(err, fmt.Sprintf("error when retrieving counter annotation for UserSignup %s, set to daily limit", signup.GetName()))
+	}
+
+	// check if counter has exceeded the limit of daily limit - if at limit error out
+	if counter > dailyLimit {
```
sbryzak, line 101: shouldn't this be:

`if counter >= dailyLimit {`

sbryzak (changes requested):
Looking pretty good, just a few minor comments to address.

pkg/verification/verification_service.go
tinakurian: okay, didn't realize it was inclusive. updated here: 81068f378a5c0a0537e45a8d350576ca62f87dfa

pkg/signup/signup_service.go
tinakurian: good point! fixed here: 707ae8b14f2c6d40e0c39c59f4ac656d4d19c320

pkg/signup/signup_service.go
```diff
@@ -240,25 +242,46 @@ func (s *ServiceImpl) GetSignup(userID string) (*Signup, error) {
...
 	return userSignup, nil
 }
+
+func (s *ServiceImpl) CheckIfUserIsKnown(userID, countryCode, phoneNumber string) error {
+	bannedUserList, err := s.BannedUsers.ListByPhoneNumber(countryCode + phoneNumber)
+	if err != nil {
+		return errors3.NewInternalError(err, "failed listing banned users")
+	}
+	for _, bannedUser := range bannedUserList.Items {
+		if bannedUser.Name == userID {
```
sbryzak, line 271: If the phone number exists _at all_ in BannedUser then the user should not be able to proceed, so this check seems unnecessary.

pkg/signup/signup_service.go
```diff
@@ -240,25 +242,46 @@ func (s *ServiceImpl) GetSignup(userID string) (*Signup, error) {
...
+			return errors3.NewForbiddenError("cannot re-register with phone number", "phone number already in use")
+		}
+	}
+
+	userSignupList, err := s.UserSignups.ListByPhoneNumber(countryCode + phoneNumber)
+	if err != nil {
+		return errors3.NewInternalError(err, "failed listing userSignups")
+	}
+	for _, signup := range userSignupList.Items {
+		if signup.Name == userID {
```
sbryzak, line 281: This still looks wrong.  We need to confirm that no _other_ users are already using the provided phone number.  If the phone number already belongs to the current user who is attempting to verify then that is fine.

sbryzak (changes requested):
Looking better, still needs a small change to `CheckIfUserIsKnown`.  Also, I would consider renaming this function, possibly to something like `CheckPhoneNumberAvailable` or `CheckPhoneNumberValid`.

pkg/signup/signup_service.go
tinakurian: ohhh okay, so what i had earlier should be fine, i've updated as part of this commit: 4ebb9c3029dabcb8590ae9e0e0984c3b6f348b7d

pkg/signup/signup_service.go
tinakurian: doh! sorry! i've updated it to be `not equal` as part of this commit: 4ebb9c3029dabcb8590ae9e0e0984c3b6f348b7d

tinakurian:
> Looking better, still needs a small change to `CheckIfUserIsKnown`. Also, I would consider renaming this function, possibly to something like `CheckPhoneNumberAvailable` or `CheckPhoneNumberValid`.

@sbryzak thanks for having another look. i've updated the name of the func as part of this commit: 4ebb9c3029dabcb8590ae9e0e0984c3b6f348b7d

pkg/signup/signup_service.go
```diff
@@ -240,25 +242,47 @@ func (s *ServiceImpl) GetSignup(userID string) (*Signup, error) {
...
 		return nil, err
 	}
 
 	return userSignup, nil
 }
+
+// CheckPhoneNumberValid checks if the phone number has been banned. If so, return
+// an internal server error. If not, check if a signup with a different userID
+// exists. If so, return an internal server error. Otherwise, return without error.
+func (s *ServiceImpl) CheckPhoneNumberValid(userID, countryCode, phoneNumber string) error {
```
alexeykazakov, line 268: Maybe something like `PhoneNumberAlreadyInUse()` would be a better name?

pkg/controller/signup.go
```diff
@@ -46,6 +47,80 @@ func (s *Signup) PostHandler(ctx *gin.Context) {
...
+		log.Errorf(ctx, nil, "Request body could not be read")
+		ctx.AbortWithError(http.StatusInternalServerError, err)
+		return
+	}
+
+	countryCode := m["country_code"]
+	phoneNumber := m["phone_number"]
+	err = s.signupService.CheckPhoneNumberValid(userID, countryCode, phoneNumber)
+	if err != nil {
+		if apierrors.IsInternalError(err) {
```
alexeykazakov, line 99: I would rather check apierrors.ForbiddenError and return http.StatusForbiden. In the case of any other error return StatusInternalServerError.

go.mod
```diff
@@ -38,6 +39,7 @@ require (
 
 replace (
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.2+incompatible // Required by OLM
+	github.com/codeready-toolchain/api => github.com/tinakurian/api v0.0.0-20200908183345-4f12c69c119d
```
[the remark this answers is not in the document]
tinakurian: done! added as part of this commit: a7be0d46781f1feea1afdfca64e727a883b05d47

pkg/controller/signup.go
```diff
@@ -46,6 +47,80 @@ func (s *Signup) PostHandler(ctx *gin.Context) {
...
+		if err != nil {
+			ctx.AbortWithError(http.StatusInternalServerError, err)
+			return
+		}
+	}
+
+	m := make(map[string]string)
+	err = json.Unmarshal(bodyBytes, &m)
+	if err != nil {
+		log.Errorf(ctx, nil, "Request body could not be read")
```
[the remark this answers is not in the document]
tinakurian: fixed as part of: a7be0d46781f1feea1afdfca64e727a883b05d47

pkg/controller/signup.go
```diff
@@ -46,6 +47,80 @@ func (s *Signup) PostHandler(ctx *gin.Context) {
 	ctx.Writer.WriteHeaderNow()
 }
 
+// UpdateVerificationHandler starts the verification process and updates a usersignup resource
+func (s *Signup) UpdateVerificationHandler(ctx *gin.Context) {
```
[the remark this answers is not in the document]
tinakurian: done as part of: a7be0d46781f1feea1afdfca64e727a883b05d47

pkg/controller/signup.go
tinakurian: ohhh good idea, updated as part of: a7be0d46781f1feea1afdfca64e727a883b05d47

pkg/signup/signup_service.go
tinakurian: sure, updated as part of: a7be0d46781f1feea1afdfca64e727a883b05d47

pkg/signup/signup_service.go
```diff
@@ -240,25 +242,47 @@ func (s *ServiceImpl) GetSignup(userID string) (*Signup, error) {
...
 		return nil, err
 	}
 
 	return userSignup, nil
 }
+
+// PhoneNumberAlreadyInUse checks if the phone number has been banned. If so, return
+// an internal server error. If not, check if a signup with a different userID
+// exists. If so, return an internal server error. Otherwise, return without error.
+func (s *ServiceImpl) PhoneNumberAlreadyInUse(userID, countryCode, phoneNumber string) error {
```
sbryzak, line 268: Ths function logic looks good now, however I'm wondering if this protection could be defeated by the user by simply inserting spaces into the phone number.  Perhaps we should "normalize" these values by removing all spaces.  Possibly we should also enforce some form of validation to ensure the user can only enter digits, spaces and brackets `()`, and then to normalize remove everything that's not a digit?  WDYT?

sbryzak (approved):
Looking good, just one minor change which I should have thought of earlier.

alexeykazakov (changes requested):

pkg/controller/signup.go
```diff
@@ -46,6 +47,82 @@ func (s *Signup) PostHandler(ctx *gin.Context) {
...
+	// Read the Body content
+	var bodyBytes []byte
+	defer func() {
+		if err := ctx.Request.Body.Close(); err != nil {
+			log.Error(nil, err, "failed to close response after reading")
+			return
+		}
+	}()
+
+	if ctx.Request.Body != nil {
```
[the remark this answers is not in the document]
tinakurian: @alexeykazakov hmm, i'm a bit confused here, this is the controller (pkg/controller/signup). i had fixed this in a previous discussion where you had asked me to extract the country_code/phone_number in the controller and pass it to the service: https://github.com/codeready-toolchain/registration-service/pull/126#discussion_r479582502

pkg/signup/signup_service.go
tinakurian: this is already addressed..if spaces are given this error is shown:
example input: 
{"country_code":"     ", "phone_number":"          "}
```
{
    "status": "Internal Server Error",
    "code": 500,
    "message": "A 'To' phone number is required.",
    "details": "error while verifying code"
}
```

tinakurian: in the case that letters are given, this is the error message: 
{"country_code":"", "phone_number":"abadsdsdfsfsfdsdffd"}
```
{
    "status": "Internal Server Error",
    "code": 500,
    "message": "The 'To' number  is not a valid phone number.",
    "details": "error while verifying code"
}
```

tinakurian: there are different variations of this error message depending on what is entered (for instance is the country_code not valid and/or the phone_number is not valid)

tinakurian: i have tested this via postman, @sbryzak if you could please also test and make sure everything is fine, that would be a great help. i have tested different variations and nothing comes back as "positive" when invalid data is given. these error messages don't come from us, they come from twilio.

pkg/controller/signup.go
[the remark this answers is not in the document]
alexeykazakov: I think I need more sleep :) I thought I'm looking at the service! But it's the controller. Sorry, my mistake!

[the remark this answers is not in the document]
alexeykazakov: BTW gin already provides some functionality for reading the json body. You can use something like this:

```
type Phone struct {
    CountryCode     string `form:"country_code" json:"country_code" binding:"required"`
    PhoneNumber string `form:"phone_number" json:"phone_number" binding:"required"`
}
...
    var phone Phone
    if err := ctx.BindJSON(&phone); err != nil {
        // Abort with BadRequest
        ...
    }
    err := s.signupService.PhoneNumberAlreadyInUse(userID, phone.CountryCode, phone.PhoneNumber)
    ...
```
No need to close the request or check other errors. Gin will take care of it.

pkg/signup/signup_service.go
alexeykazakov: Are you saying this is validated by twilio? So, if the user entered a phone number with spaces then it will fail?
Shouldn't we normalize the phone number (delete all spaces) as Shane suggested?

tinakurian: yeah, twilio already performs basic phone number validation.

> Twilio's API has some very basic phone number validation logic built in. We try to save you money by rejecting messages to numbers we are certain are incorrect. However, we trust that our users know more about the phone number they are trying to send a message to than we do. Our validation logic will only affect numbers which are unquestionably incorrect. It is your responsibility to provide accurate data. 

[source: https://support.twilio.com/hc/en-us/articles/223181828-Does-Twilio-check-to-see-if-phone-numbers-can-receive-SMS-]

however, we could add our own basic phone number validation to be extra safe - WDYT? The things we don't validate against currently are things like validating if a number from a certain region is correct - do we want to add this?

the UI should block a lot of the basic problematic phone number entries.. currently the code in the UI does not handle this

pkg/controller/signup.go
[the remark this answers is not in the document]
tinakurian: lol! no worries! ohhhh that's much nicer.. okay i'll change it to that :+1:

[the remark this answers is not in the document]
tinakurian: @alexeykazakov does this look better? 1416a883fbf747a701454c405f6bb715bb0d681e

pkg/kubeclient/banneduser.go
```diff
@@ -53,3 +54,33 @@ func (c *bannedUserClient) List(email string) (*crtapi.BannedUserList, error) {
...
+	list, err := intf.Resource(r).Namespace(c.ns).List(context.TODO(), listOptions)
+	if err != nil {
+		return nil, err
+	}
+
+	result := &crtapi.BannedUserList{}
+
+	err = c.crtClient.scheme.Convert(list, result, nil)
+	return result, err
+}
```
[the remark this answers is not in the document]
tinakurian: @alexeykazakov is this more what you had in mind? f7a607c13356a51a07c936a1be4f51302aafb51f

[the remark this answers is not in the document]
tinakurian: renamed the func here to what you had suggested - sorry missed the suggestion: 03541fc47918e8345f2c50e29407ea363bfaeffe

pkg/kubeclient/banneduser.go
```diff
@@ -21,17 +21,17 @@ type bannedUserClient struct {
 }
 
 type BannedUserInterface interface {
-	List(email string) (*crtapi.BannedUserList, error)
+	ListByHashedLabel(value, label string) (*crtapi.BannedUserList, error)
 }
 
-// List returns a BannedUserList containing any BannedUser resources that have a label matching the specified email address
-func (c *bannedUserClient) List(email string) (*crtapi.BannedUserList, error) {
+// ListByHashedLabel returns a BannedUserList containing any BannedUser resources that have a label matching the specified label
```
alexeykazakov, line 27: What I actually meant is something like this:

```
type BannedUserInterface interface {
	ListByEmail(email string) (*crtapi.BannedUserList, error)
        ListByPhone(phone string) (*crtapi.BannedUserList, error)
}
...
func (c *bannedUserClient) ListByEmail(email string) (*crtapi.BannedUserList, error) {
     return c.listByHashedLabel("<email_label_key>", email)
}
func (c *bannedUserClient) ListByPhone(phone string) (*crtapi.BannedUserList, error) {
     return c.listByHashedLabel("<phone_label_key>", phone)
}
func (c *bannedUserClient) listByHashedLabel(labelKey, labelValue string) (*crtapi.BannedUserList, error) {
     ...
}
```

pkg/kubeclient/signup.go
```diff
@@ -68,3 +76,33 @@ func (c *userSignupClient) Update(obj *crtapi.UserSignup) (*crtapi.UserSignup, e
 	}
 	return result, nil
 }
+
+// ListByHashedLabel returns a UserSignupList containing any UserSignup resources that have a label matching the specified label
+func (c *userSignupClient) ListByHashedLabel(value, label string) (*crtapi.UserSignupList, error) {
```
alexeykazakov, line 81: Same here as above.

pkg/signup/signup_service.go
alexeykazakov: It's not only about validation.
What if the user enter the phone number as:
- `(925) 123-4567`
- `925 123-45-67`
- `925 123 4567`
We will fail, right? So, we need to "normalize" them. Either in UI (then our backend should expect only numbers and fail with anything else) or in the backend. I would do the normalization in our backend. Just delete all not digit symbols from the phone string. So all the examples above will be normalized to the same string: `9251234567`. And it must be done in the controller level. So we use the normalized number everywhere.

pkg/signup/signup_service.go
```diff
@@ -240,25 +242,47 @@ func (s *ServiceImpl) GetSignup(userID string) (*Signup, error) {
...
+		return errors3.NewInternalError(err, "failed listing userSignups")
+	}
+	for _, signup := range userSignupList.Items {
+		if signup.Name != userID {
+			return errors3.NewForbiddenError("cannot re-register with phone number", "phone number already in use")
+		}
+	}
+
+	return nil
+}
```
alexeykazakov, line 288: Any tests for PhoneNumberAlreadyInUse()?

pkg/verification/verification_service.go
```diff
@@ -57,28 +60,72 @@ func NewVerificationService(cfg ServiceConfiguration, opts ...VerificationServic
...
+		err := errors.NewForbiddenError("daily limit exceeded", "cannot generate new verification code")
+		log.Error(ctx, err, fmt.Sprintf("%d attempts made. the daily limit of %d has been exceeded", counter, dailyLimit))
+		return signup, err
+	}
+
+	// generate verification code
+	code, err := generateVerificationCode()
 	if err != nil {
-		return err
+		log.Errorf(ctx, nil, "verification code could not be generated")
```
alexeykazakov, line 110:
```suggestion
		log.Errorf(ctx, err, "verification code could not be generated")
```

pkg/signup/signup_service.go
tinakurian: no, these cases are valid with twilio. i tried each one of them with my own phone number and it works.. i guess as the document states it will only do basic validation, so as long as the number looks correct it will not fail, hence no need to normalize. they already do the normalization and validation on their end..

tinakurian: also Michael added some validation to the UI as well using regex so it's covered on that end as well

pkg/controller/signup.go
```diff
@@ -46,6 +50,61 @@ func (s *Signup) PostHandler(ctx *gin.Context) {
...
+		}
+		log.Error(ctx, err, "error retrieving usersignup")
+		errors.AbortWithError(ctx, http.StatusInternalServerError, err, fmt.Sprintf("error retrieving usersignup: %s", userID))
+		return
+	}
+
+	// check that verification is required before proceeding
+	if signup.Spec.VerificationRequired == false {
+		log.Errorf(ctx, nil, "phone verification not required for usersignup: %s", userID)
+		ctx.AbortWithStatus(http.StatusBadRequest)
```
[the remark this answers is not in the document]
tinakurian: sure, updated here: a100bd39b021ccdc5cc9701ddeb4a0cddf814da4 and here: 3b886fc07d0edc7da639c9b7d35940c466cf50ef

pkg/signup/signup_service.go
alexeykazakov: OK. But then if the same user tries the same number, first time as 1234567 and the second time as 123 4567, then we will treat them as two different numbers. But we shouldn't. So we have to normalize them on our backend side.

pkg/controller/signup.go
```diff
@@ -46,6 +50,61 @@ func (s *Signup) PostHandler(ctx *gin.Context) {
...
+	}
+
+	// Read the Body content
+	var phone Phone
+	if err := ctx.BindJSON(&phone); err != nil {
+		errors.AbortWithError(ctx, http.StatusBadRequest, err, "error reading request body")
```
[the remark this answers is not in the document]
tinakurian: logged here: 47404b2c47f1f213eb2f0243c9e3f94db04bd426

pkg/controller/signup.go
```diff
@@ -46,6 +50,61 @@ func (s *Signup) PostHandler(ctx *gin.Context) {
...
+		default:
+			errors.AbortWithError(ctx, http.StatusInternalServerError, err, "error while verifying code")
+		case *apierrors.StatusError:
+			errors.AbortWithError(ctx, int(t.ErrStatus.Code), err, t.ErrStatus.Message)
+		}
+		return
+	}
+
+	log.Infof(ctx, "phone verification has been sent for userID %s", userID)
+	ctx.Status(http.StatusOK)
```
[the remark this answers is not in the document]
tinakurian: done as part of: a100bd39b021ccdc5cc9701ddeb4a0cddf814da4 and 3b886fc07d0edc7da639c9b7d35940c466cf50ef

pkg/controller/signup_test.go
```diff
@@ -218,6 +222,169 @@ func (s *TestSignupSuite) TestSignupGetHandler() {
...
+		req, err = http.NewRequest(http.MethodPost, "/api/v1/signup/verification", bytes.NewBuffer(data))
+		ctx.Request = req
+		ctx.Set(context.SubKey, userID)
+
+		handler(ctx)
+
+		// Check the status code is what we expect.
+		assert.Equal(s.T(), http.StatusBadRequest, rr.Code, "handler returned wrong status code")
+
+		// Check that the correct UserSignup is passed into the FakeSignupService for update
```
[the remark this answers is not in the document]
tinakurian: These tests were pretty messy, i tried to make them more in-line with how we did the other tests for the GET verification endpoint: 956c1f0560a41490abd17b51dd429e4d8f7bdaf7 and 7a21751dd5572329832fd519d4a3bb1e54030da5 is this better?

pkg/controller/signup_test.go
```diff
@@ -218,6 +222,169 @@ func (s *TestSignupSuite) TestSignupGetHandler() {
...
+		ctx.Request = req
+		ctx.Set(context.SubKey, userID)
+
+		handler(ctx)
+
+		// Check the status code is what we expect.
+		assert.Equal(s.T(), http.StatusForbidden, rr.Code, "handler returned wrong status code")
+
+		// Check that the correct UserSignup is passed into the FakeSignupService for update
+		require.Equal(s.T(), userID, expected.Name)
```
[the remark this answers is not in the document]
tinakurian: These tests were pretty messy, i tried to make them more in-line with how we did the other tests for the GET verification endpoint: 956c1f0 and 7a21751 is this better?

pkg/kubeclient/banneduser.go
tinakurian: fixed as part of these commits: 956c1f0 and 7a21751

pkg/kubeclient/signup.go
tinakurian: fixed as part of: 956c1f0 and 7a21751

pkg/signup/signup_service.go
tinakurian: added as part of: 956c1f0 and 7a21751

pkg/verification/verification_service.go
tinakurian: fixed: 6bc0a5643b52795636f9453134177b2cd9c167e5

pkg/signup/signup_service.go
tinakurian: ohhh right, because of the hash, i've updated it as part of: 43237d7c8084c491f3f9539a329e2fe6a7d8cb3f is this okay?

pkg/controller/signup_test.go
```diff
@@ -218,6 +222,169 @@ func (s *TestSignupSuite) TestSignupGetHandler() {
...
+		require.Equal(s.T(), "jsmith@redhat.com", expected.Annotations[crtapi.UserSignupUserEmailAnnotationKey])
+		require.Equal(s.T(), "1", expected.Annotations[crtapi.UserSignupVerificationCounterAnnotationKey])
+		require.Equal(s.T(), dataMap["country_code"]+dataMap["phone_number"], expected.Annotations[crtapi.UserSignupUserPhoneHashLabelKey])
+		require.Equal(s.T(), verificationInitTimeStamp, expected.Annotations[crtapi.UserSignupVerificationInitTimestampAnnotationKey])
+	})
+}
+
 func (s *TestSignupSuite) TestVerifyCodeHandler() {
```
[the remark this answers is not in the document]
tinakurian: i've added some more tests, is there a scenario i missed?

pkg/controller/signup.go
```diff
@@ -46,6 +50,61 @@ func (s *Signup) PostHandler(ctx *gin.Context) {
...
+			errors.AbortWithError(ctx, http.StatusNotFound, err, "usersignup not found")
+		}
+		log.Error(ctx, err, "error retrieving usersignup")
+		errors.AbortWithError(ctx, http.StatusInternalServerError, err, fmt.Sprintf("error retrieving usersignup: %s", userID))
+		return
+	}
+
+	// check that verification is required before proceeding
+	if signup.Spec.VerificationRequired == false {
+		log.Errorf(ctx, nil, "phone verification not required for usersignup: %s", userID)
```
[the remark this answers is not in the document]
tinakurian: yeah, better not to, fixed it here: 35d5380f320740928205f64f8c09999fdf665ce2

pkg/configuration/configuration.go
```diff
@@ -139,6 +139,10 @@ const (
 	// varVerificationExcludedEmailDomains contains a comma-separated list of domain names for which phone verification
 	// is not required.  For example: "redhat.com,ibm.com"
 	varVerificationExcludedEmailDomains = "verification.excluded_email_domains"
+
+	// varVerificationExpiry is used to set the amount of time a verification code is active for before expiring
```
alexeykazakov, line 143: Amount of time? In what units? Minutes I guess?
Let's call it `verification.code_expires_in_min` so it's clear what this is about.

pkg/errors/error.go
```diff
@@ -54,3 +54,21 @@ func NewInternalError(err error, details string) *Error {
...
+func NewNotFoundError(err error, details string) *Error {
+	return &Error{
+		Status:  http.StatusText(http.StatusNotFound),
+		Code:    http.StatusNotFound,
+		Message: err.Error(),
+		Details: details,
+	}
+}
+
+func NewBadRequest(message, details string) *Error {
```
alexeykazakov, line 67: Please add a test for this one too.

pkg/controller/signup.go
```diff
@@ -46,6 +52,85 @@ func (s *Signup) PostHandler(ctx *gin.Context) {
...
+		return
+	}
+
+	// normalize phone number
+	r := strings.NewReplacer("(", "",
+		")", "",
+		" ", "",
+		"-", "")
+	countryCode := r.Replace(phone.CountryCode)
+	phoneNumber := r.Replace(phone.PhoneNumber)
```
alexeykazakov, line 92: Please add tests for different valid cases like `(123)`, `1 2 3`, etc. As far as I can see you have only tests for the invalid symbols if I'm not missing anything.

pkg/configuration/configuration.go
tinakurian: sorry, i added this in the Get func comment but guess i forgot to add it here.. anyways i've updated it here: 97cfceeac77f1fb7ff25e3b81945a35b07093d4d and 873f3c16c9e09e21ab31ef29e475c9e2ca115d14

pkg/errors/error.go
tinakurian: updated here: 3cd7858e9204ccb4210d579bd1d341567ee9a2b0

pkg/controller/signup.go
tinakurian: updated here: 3cd7858e9204ccb4210d579bd1d341567ee9a2b0

pkg/controller/signup_test.go
```diff
@@ -328,6 +328,99 @@ func (s *TestSignupSuite) TestUpdateVerificationHandler() {
...
+		assert.Equal(s.T(), http.StatusNoContent, rr.Code, "handler returned wrong status code")
+
+		// Check that the correct UserSignup is passed into the FakeSignupService for update
+		require.Equal(s.T(), userID, storedUserSignup.Name)
+		require.Equal(s.T(), "jsmith@redhat.com", storedUserSignup.Annotations[crtapi.UserSignupUserEmailAnnotationKey])
+		require.Equal(s.T(), "1", storedUserSignup.Annotations[crtapi.UserSignupVerificationCounterAnnotationKey])
+		require.Equal(s.T(), storedVerificationCode, storedUserSignup.Annotations[crtapi.UserSignupVerificationCodeAnnotationKey])
+		require.Equal(s.T(), verificationInitTimeStamp, storedUserSignup.Annotations[crtapi.UserSignupVerificationInitTimestampAnnotationKey])
+		require.Equal(s.T(), "12268213044", storedUserSignup.Labels[crtapi.UserSignupUserPhoneHashLabelKey])
+	})
```
alexeykazakov, line 423: All these four tests seem to be almost identical. You could just create a re-usable function like `verifyInitVerificationSuccess(rawPhoneNumber, expectedPhoneNumber string)`
and call it four times with `verifyInitVerificationSuccess("2 2 6 8 2 1 3 0 4 4", "2268213044")`, `verifyInitVerificationSuccess("(226) 821-3044", "2268213044")`, etc.

alexeykazakov: @tinakurian to make the code even cleaner/more readable you could extract the actual phone normalization test to a custom assert function within the same test to cover all the cases.
Similar to what we do in other places. For example https://github.com/codeready-toolchain/member-operator/blob/master/pkg/controller/idler/idler_controller_test.go#L341-L373
so some local function like `assertPhoneNormalized(rawPhoneNumber, expectedPhoneNumber)`

tinakurian: @alexeykazakov i re-organized the test this way - is this okay? e8239d20e14faa8e1dc02e3dbe798aa6ff704e6a

alexeykazakov (approved):

pkg/controller/signup_test.go
alexeykazakov: It's better. Thanks.

PR #127, opened 2020-09-02
Upgrade to latest operator-sdk

tinakurian:


alexeykazakov (commented):
go.mod changes look good. But all these ctx changes do not seem to be related to the kube update?

tinakurian:
> go.mod changes look good. But all these ctx changes do not seem to be related to the kube update?

for some reason, when i upgrade to the latest toolchain/api this breaks and the client requires Do() to be changed to Do(ctx) - i guess the api changed... none of the tests will pass without this change..

so now in kubeclient->mur.go->Get(...) for instance it would need to be the following:
```
	err := c.client.Get().
		Namespace(c.ns).
		Resource(masterUserRecordResourcePlural).
		Name(name).
		Do(ctx).
		Into(result)
```

alexeykazakov:
Ah OK. But I don't think we need to pass the gin context. `context.TODO()` should work fine I think.

tinakurian:
> Ah OK. But I don't think we need to pass the gin context. `context.TODO()` should work fine I think.

ohh okay so more like this then: 249f11272467859413329cbba512ef15ca70cab1 ?

alexeykazakov (approved):
Yes. Looks good!

PR #130, opened 2020-09-09
update reg-service.yaml

tinakurian:
Updating reg-service.yaml to be identical to what is found in host-operator: https://github.com/codeready-toolchain/host-operator/blob/master/deploy/registration-service/registration-service.yaml


alexeykazakov (approved):

MatousJobanek (approved):
Thanks :+1:

PR #133, opened 2020-09-25
Minor fixes for phone verification

sbryzak:
This PR contains minor fixes for the phone verification workflow.

### DIFF
--- a/pkg/controller/signup.go
+++ b/pkg/controller/signup.go
@@ -90,8 +90,8 @@ func (s *Signup) UpdateVerificationHandler(ctx *gin.Context) {
 		"-", "")
 	countryCode := r.Replace(phone.CountryCode)
 	phoneNumber := r.Replace(phone.PhoneNumber)
-	countryValid, err := regexp.MatchString(`\+?[0-9]+`, countryCode)
-	phoneValid, err := regexp.MatchString(`[0-9]+`, phoneNumber)
+	countryValid, _ := regexp.MatchString(`^\+?[0-9]+$`, countryCode)
+	phoneValid, _ := regexp.MatchString(`^[0-9]+$`, phoneNumber)
 
 	// if phone number contains odd characters, return error
 	if !countryValid || !phoneValid {
--- a/pkg/controller/signup.go
+++ b/pkg/controller/signup.go
@@ -90,11 +90,11 @@ func (s *Signup) UpdateVerificationHandler(ctx *gin.Context) {
 		"-", "")
 	countryCode := r.Replace(phone.CountryCode)
 	phoneNumber := r.Replace(phone.PhoneNumber)
-	matchedCode, err := regexp.MatchString("[^0-9]+", countryCode)
-	matchedPhone, err := regexp.MatchString("[^0-9]+", phoneNumber)
+	countryValid, err := regexp.MatchString(`\+?[0-9]+`, countryCode)
+	phoneValid, err := regexp.MatchString(`[0-9]+`, phoneNumber)
 
 	// if phone number contains odd characters, return error
-	if matchedCode || matchedPhone {
+	if !countryValid || !phoneValid {
 		log.Errorf(ctx, errors.NewBadRequest("bad request", "invalid request"), "phone number entered contains invalid characters")
 		ctx.AbortWithStatus(http.StatusBadRequest)
 		return
@@ -115,11 +115,22 @@ func (s *Signup) UpdateVerificationHandler(ctx *gin.Context) {
 	}
 
 	signup, err = s.verificationService.InitVerification(ctx, signup, phone.CountryCode, phone.PhoneNumber)
+	_, err2 := s.signupService.UpdateUserSignup(signup)
+	if err2 != nil {
+		log.Error(ctx, err2, "error while updating UserSignup resource")
+		errors.AbortWithError(ctx, http.StatusInternalServerError, err2, "error while updating UserSignup resource")
+
+		if err != nil {
+			log.Error(ctx, err, "error initiating user verification")
+		}
+		return
+	}
+
 	if err != nil {
 		log.Errorf(ctx, nil, "Verification for %s could not be sent", userID)
 		switch t := err.(type) {
 		default:
-			errors.AbortWithError(ctx, http.StatusInternalServerError, err, "error while verifying code")
+			errors.AbortWithError(ctx, http.StatusInternalServerError, err, "error while initiating verification")
 		case *apierrors.StatusError:
 			errors.AbortWithError(ctx, int(t.ErrStatus.Code), err, t.ErrStatus.Message)
 		}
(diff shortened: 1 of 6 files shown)


pkg/server/server.go
```diff
@@ -23,8 +23,15 @@ type RegistrationServer struct {
 
 // New creates a new RegistrationServer object with reasonable defaults.
 func New(config *configuration.Config) *RegistrationServer {
+
+	ginRouter := gin.New()
+	ginRouter.Use(
+		gin.LoggerWithWriter(gin.DefaultWriter, "/api/v1/health"),
+		gin.Recovery(),
+	)
```
alexeykazakov, line 31: What exactly does this code do? I guess you need some special logging for that endpoint. Can you add some comments? What and why you are doing this.

alexeykazakov (commented):
It looks like you are fixing a few bugs in the code but I don't see any tests. Does it mean that this code is not covered by any tests? Shouldn't we add tests for this code then too?

sbryzak:
@alexeykazakov some of the changes are covered by unit tests, however all of them are covered by accompanying e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/167

pkg/controller/signup.go
```diff
@@ -90,11 +90,11 @@ func (s *Signup) UpdateVerificationHandler(ctx *gin.Context) {
 		"-", "")
 	countryCode := r.Replace(phone.CountryCode)
 	phoneNumber := r.Replace(phone.PhoneNumber)
-	matchedCode, err := regexp.MatchString("[^0-9]+", countryCode)
-	matchedPhone, err := regexp.MatchString("[^0-9]+", phoneNumber)
+	countryValid, _ := regexp.MatchString(`^\+?[0-9]+$`, countryCode)
+	phoneValid, _ := regexp.MatchString(`^[0-9]+$`, phoneNumber)
```
MatousJobanek, line 94: Could you please add unit tests verifying the regex?
Ideally, extract the verification logic to functions and run a set of test cases against it. Thanks to the tests it will be easier to understand which numbers/country codes are valid and  which not.

pkg/controller/signup.go
```diff
@@ -115,11 +115,22 @@ func (s *Signup) UpdateVerificationHandler(ctx *gin.Context) {
 	}
 
 	signup, err = s.verificationService.InitVerification(ctx, signup, phone.CountryCode, phone.PhoneNumber)
+	_, err2 := s.signupService.UpdateUserSignup(signup)
```
MatousJobanek, line 118: I'm curious - why do we need to return the `UserSignup` instances from the methods when we pass the same `UserSignup` instances as a pointer

pkg/signup/signup_service_test.go
```diff
@@ -450,6 +450,7 @@ func (s *TestSignupServiceSuite) TestGetSignupStatusOK() {
 	assert.True(s.T(), response.Status.Ready)
 	assert.Equal(s.T(), response.Status.Reason, "mur_ready_reason")
 	assert.Equal(s.T(), response.Status.Message, "mur_ready_message")
+	assert.False(s.T(), response.Status.VerificationRequired)
```
MatousJobanek, line 453: shouldn't we check this param also in the other cases?

pkg/controller/signup.go
sbryzak: There's already a test for this - https://github.com/codeready-toolchain/registration-service/blob/master/pkg/controller/signup_test.go#L460

I'm not going to touch the structure of the code in this PR, that is planned for a separate PR.  This PR is intended as a targeted set of changes to enable the e2e tests to pass, and due to the restrictions of the e2e test setup not being capable of executing test across 3 branches this PR will need to be merged in order for the corresponding PRs in toolchain-e2e and host-operator to pass.

pkg/controller/signup.go
sbryzak: It's not necessary, however I'm not going to change this here (see my other comment above).  I can address this in a separate PR, we are planning an architectural refactor to clean up the entire verification workflow.

pkg/signup/signup_service_test.go
sbryzak: There's already an explicit test for this property, see `TestUserWithExcludedDomainEmailSignsUp()` further up.  The prior test `TestCreateUserSignup()` tests it also.

pkg/controller/signup.go
MatousJobanek: Ok, my point was that you changed the regex, then I would expect that we have a test case that covers this change, otherwise, we don't know if the regex is finally OK-ish or requires some additional changes. Apart from that, existing unit test would ensure that any future refactoring changes won't break anything.
In addition, such tests would show what the regex expects, what is valid and what not - it would be really helpful for dev who are not strong at regexes ;-)

pkg/controller/signup.go
MatousJobanek: ok, I didn't know that you are planning an architectural refactor to clean up the code :+1:

pkg/signup/signup_service_test.go
MatousJobanek: Ok, I missed that. But correct me if I'm wrong - none of those tests that you mentioned verifies the `GetSignup` method, right? In other words, they don't verify getting of the already existing `UserSignup` resource, but they verify creation of the new one.

pkg/controller/signup.go
sbryzak: I changed the regex because an existing test (from master) was failing, however when running the tests from the master branch the tests pass.... so really weird behaviour.  In any case, the regex was wrong, it was searching for any match even if the string as a whole contained invalid characters.  If I had more time I'd dig deeper into why the tests seem to be passing in master however I've not had a spare moment to look at this as of yet.

pkg/signup/signup_service_test.go
sbryzak: Good catch, you're absolutely right.  I've updated the tests that check the Signup status returned from `GetSignup` to also confirm that `verificationRequired` is correctly set.

pkg/controller/signup.go
tinakurian: @sbryzak this regex should only match strings that are anything other than the numbers 0-9. in the regex you've provided you're doing the reverse. whichever way is fine, but i don't understand how/why the regex was broken. you can test it out here: https://regex101.com/

pkg/signup/signup_service_test.go
MatousJobanek: awesome, thanks :+1:

pkg/controller/signup.go
MatousJobanek: To be honest, I would rather prefer having a large set of strings that are executed against the regex instead of using third-party service for verifying that the regex works. We could either store them as global variable and use them is special test cases or move them into separated function doing the only function that will verify the validity and test the string against the function. Something similar as we have for username transformation: https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/usersignup/usersignup_controller_test.go#L2010-L2035

alexeykazakov: Fully agree with @MatousJobanek. We have to have a list of explicit assertions testing different variants of phone numbers.
We can't say that this code works until we have those tests in place. We wouldn't have this argument regarding whether the new/old regex is correct or not if we had proper tests in place.

MatousJobanek: @tinakurian btw, Shane is correct with his fix - the original regex was wrong: `"[^0-9]+"`
* for specifying regex, we don't use the normal "double apostrophe" but the backquote (but it's just a detail)
* AFAIK the char `^` in the square brackets means negation so your regex accepted all strings that didn't contain a number
* if you want to accept only the numbers you need to use: `[0-9]+`

tinakurian: @MatousJobanek i fully intended the regex to match on everything other than numbers. so it was doing what i expected and wanted. as for the third party tool, i had only suggested it for testing out what exactly the regex works against in terms of understanding it since @sbryzak had mentioned that it was broken.. again, i had intended it to do what it does - match on characters that aren't numbers..

MatousJobanek (approved):

PR #134, opened 2020-10-05
Architectural refactoring

sbryzak:
This PR does the following:

- Introduces a managed services layer
- Improves separation of controller/model
- Modifies tests to use actual services rather than business logic-heavy mocks

Fixes https://issues.redhat.com/browse/CRT-825


pkg/verification/verification_service.go
```diff
@@ -147,56 +189,78 @@ func generateVerificationCode() (string, error) {
 
 // VerifyCode validates the user's phone verification code.  It updates the specified UserSignup value, so even
 // if an error is returned by this function the caller should still process changes to it
-func (s *ServiceImpl) VerifyCode(signup *v1alpha1.UserSignup, code string) (*v1alpha1.UserSignup, error) {
+func (s *ServiceImpl) VerifyCode(ctx *gin.Context, userID string, code string) (verificationErr error) {
+
+	signup, verificationErr := s.Services().SignupService().GetUserSignup(userID)
+	if verificationErr != nil {
+		log.Error(ctx, verificationErr, "error retrieving usersignup")
+		return
```
tinakurian, line 197: is there a reason why the error is not being returned here?

pkg/verification/verification_service.go
```diff
@@ -147,56 +189,78 @@ func generateVerificationCode() (string, error) {
...
+		// If the code matches then set VerificationRequired to false, reset other verification annotations
+		signup.Spec.VerificationRequired = false
+		delete(signup.Annotations, v1alpha1.UserSignupVerificationCodeAnnotationKey)
+		delete(signup.Annotations, v1alpha1.UserVerificationAttemptsAnnotationKey)
+		delete(signup.Annotations, v1alpha1.UserSignupVerificationCounterAnnotationKey)
+		delete(signup.Annotations, v1alpha1.UserSignupVerificationTimestampAnnotationKey)
+		delete(signup.Annotations, v1alpha1.UserSignupVerificationInitTimestampAnnotationKey)
+		delete(signup.Annotations, v1alpha1.UserVerificationExpiryAnnotationKey)
+	} else {
+		log.Error(ctx, verificationErr, "error validating verification code")
```
tinakurian, line 255: is there a reason why we don't return the error?

pkg/verification/verification_service.go
```diff
@@ -147,56 +189,78 @@ func generateVerificationCode() (string, error) {
...
-	delete(signup.Annotations, v1alpha1.UserVerificationExpiryAnnotationKey)
+	// Update changes made to the UserSignup
+	_, updateErr := s.Services().SignupService().UpdateUserSignup(signup)
+	if updateErr != nil {
+		log.Error(ctx, updateErr, "error updating UserSignup")
+		verificationErr = updateErr
+	}
 
-	return signup, nil
+	return
```
tinakurian, line 265: it looks like this function swallows the errors, it would probably be beneficial to return it?

pkg/verification/verification_service.go
```diff
@@ -147,56 +189,78 @@ func generateVerificationCode() (string, error) {
...
+		// so that we at least now have a valid value, and let the workflow continue to the
+		// subsequent attempts check
+		attemptsMade = s.config.GetVerificationAttemptsAllowed()
+		signup.Annotations[v1alpha1.UserVerificationAttemptsAnnotationKey] = strconv.Itoa(attemptsMade)
 	}
 
 	// If the user has made more attempts than is allowed per day, return an error
 	if attemptsMade >= s.config.GetVerificationAttemptsAllowed() {
-		return signup, errors.NewTooManyRequestsError("too many verification attempts", "")
+		verificationErr = errors.NewTooManyRequestsError("too many verification attempts", "")
```
tinakurian, line 221: instead of assigning this error, why not just return it?

pkg/verification/verification_service.go
```diff
@@ -147,56 +189,78 @@ func generateVerificationCode() (string, error) {
...
 
-	if now.After(exp) {
-		// If it is now past the expiry timestamp for the verification code, return a 403 Forbidden error
-		return signup, errors.NewForbiddenError("expired", "verification code expired")
+	if verificationErr == nil {
+		if code != signup.Annotations[v1alpha1.UserSignupVerificationCodeAnnotationKey] {
+			// The code doesn't match
+			attemptsMade++
+			signup.Annotations[v1alpha1.UserVerificationAttemptsAnnotationKey] = strconv.Itoa(attemptsMade)
+			verificationErr = errors.NewForbiddenError("invalid code", "the provided code is invalid")
```
tinakurian, line 241: without returning the error and overwriting the error, we not only loose the previous error but we also continue on as though no error happened - is this intended?

pkg/verification/verification_service.go
```diff
@@ -63,13 +64,43 @@ func NewVerificationService(cfg ServiceConfiguration, opts ...VerificationServic
...
 // InitVerification sends a verification message to the specified user.  If successful,
 // the user will receive a verification SMS.
-func (s *ServiceImpl) InitVerification(ctx *gin.Context, signup *v1alpha1.UserSignup, countryCode, phoneNumber string) (*v1alpha1.UserSignup, error) {
-	// get phone number hash
+func (s *ServiceImpl) InitVerification(ctx *gin.Context, userID string, e164PhoneNumber string) error {
+	signup, err := s.Services().SignupService().GetUserSignup(userID)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Error(ctx, err, "usersignup not found")
+			errors.AbortWithError(ctx, http.StatusNotFound, err, "usersignup not found")
```
tinakurian, line 72: this code seems a bit inconsistent we could use `errors.NewNotFoundError` or `errors.NewForbiddenError` but here we rely on `errors.AbortWithError` - is there a reason for this difference?

pkg/verification/verification_service.go
```diff
@@ -63,13 +64,43 @@ func NewVerificationService(cfg ServiceConfiguration, opts ...VerificationServic
...
+			return
+		}
+		log.Error(ctx, err, "error retrieving usersignup")
+		errors.AbortWithError(ctx, http.StatusInternalServerError, err, fmt.Sprintf("error retrieving usersignup: %s", userID))
```
tinakurian, line 76: same here, we could use `errors.NewInternalError(...)` but here we use `errors.AbortWithError(...)` is there a reason? or could we stick to a more unified approach?

pkg/verification/verification_service.go
```diff
@@ -63,13 +64,43 @@ func NewVerificationService(cfg ServiceConfiguration, opts ...VerificationServic
...
+		log.Errorf(ctx, errors.NewForbiddenError("forbidden request", "verification code will not be sent"), "phone verification not required for usersignup: %s", userID)
+		ctx.AbortWithStatus(http.StatusBadRequest)
+		return
+	}
+
+	err = s.Services().SignupService().PhoneNumberAlreadyInUse(userID, e164PhoneNumber)
+	if err != nil {
+		if apierrors.IsForbidden(err) {
+			log.Errorf(ctx, err, "phone number already in use, cannot register using phone number: %s", e164PhoneNumber)
+			errors.AbortWithError(ctx, http.StatusForbidden, err, fmt.Sprintf("phone number already in use, cannot register using phone number: %s", e164PhoneNumber))
```
tinakurian, line 91: same as above

pkg/verification/verification_service.go
```diff
@@ -63,13 +64,43 @@ func NewVerificationService(cfg ServiceConfiguration, opts ...VerificationServic
...
+			return
+		}
+		log.Error(ctx, err, "error while looking up users by phone number")
+		errors.AbortWithError(ctx, http.StatusInternalServerError, err, "could not lookup users by phone number")
```
tinakurian, line 95: same as above

pkg/verification/verification_service.go
```diff
@@ -147,56 +189,78 @@ func generateVerificationCode() (string, error) {
 
 // VerifyCode validates the user's phone verification code.  It updates the specified UserSignup value, so even
 // if an error is returned by this function the caller should still process changes to it
-func (s *ServiceImpl) VerifyCode(signup *v1alpha1.UserSignup, code string) (*v1alpha1.UserSignup, error) {
+func (s *ServiceImpl) VerifyCode(ctx *gin.Context, userID string, code string) (verificationErr error) {
```
tinakurian, line 192: this function doesn't appear to ever return an error - should this be removed?

pkg/verification/verification_service.go
```diff
@@ -63,13 +64,43 @@ func NewVerificationService(cfg ServiceConfiguration, opts ...VerificationServic
...
+		}
+		log.Error(ctx, err, "error retrieving usersignup")
+		errors.AbortWithError(ctx, http.StatusInternalServerError, err, fmt.Sprintf("error retrieving usersignup: %s", userID))
+		return
+	}
+
+	// check that verification is required before proceeding
+	if signup.Spec.VerificationRequired == false {
+		log.Errorf(ctx, errors.NewForbiddenError("forbidden request", "verification code will not be sent"), "phone verification not required for usersignup: %s", userID)
+		ctx.AbortWithStatus(http.StatusBadRequest)
```
tinakurian, line 83: same as above

pkg/verification/verification_service.go
```diff
@@ -94,22 +125,22 @@ func (s *ServiceImpl) InitVerification(ctx *gin.Context, signup *v1alpha1.UserSi
...
-		return signup, err
+		return err
 	}
 
 	// generate verification code
 	code, err := generateVerificationCode()
 	if err != nil {
 		log.Errorf(ctx, err, "verification code could not be generated")
-		return signup, ctx.AbortWithError(http.StatusInternalServerError, err)
+		return ctx.AbortWithError(http.StatusInternalServerError, err)
```
tinakurian, line 143: maybe this should be changed to use `errors.NewInternalError(...)`

pkg/verification/verification_service.go
```diff
@@ -119,15 +150,26 @@ func (s *ServiceImpl) InitVerification(ctx *gin.Context, signup *v1alpha1.UserSi
...
 	if err != nil {
 		log.Error(ctx, err, fmt.Sprintf("error while sending, code: %d message: %s", msg.ErrorCode, msg.ErrorMessage))
-		return signup, err
+		return err
+	}
+
+	_, err2 := s.Services().SignupService().UpdateUserSignup(signup)
+	if err2 != nil {
+		log.Error(ctx, err2, "error while updating UserSignup resource")
+		errors.AbortWithError(ctx, http.StatusInternalServerError, err2, "error while updating UserSignup resource")
```
tinakurian, line 163: maybe this should be changed to use `errors.NewInternalError(...)`

pkg/verification/verification_service.go
```diff
@@ -119,15 +150,26 @@ func (s *ServiceImpl) InitVerification(ctx *gin.Context, signup *v1alpha1.UserSi
...
+
+		if err != nil {
+			log.Error(ctx, err, "error initiating user verification")
```
tinakurian, line 166: hmm, isn't this error already logged above on line 154?

pkg/verification/verification_service.go
```diff
@@ -119,15 +150,26 @@ func (s *ServiceImpl) InitVerification(ctx *gin.Context, signup *v1alpha1.UserSi
...
+		}
+
+		return
```
tinakurian, line 169: any reason why we have this return here?

pkg/verification/verification_service.go
```diff
@@ -147,56 +189,78 @@ func generateVerificationCode() (string, error) {
...
-	exp, err := time.Parse(TimestampLayout, signup.Annotations[v1alpha1.UserVerificationExpiryAnnotationKey])
-	if err != nil {
-		// If the verification expiry timestamp is corrupt or missing, then return an error
-		return signup, errors.NewInternalError(err, "error parsing expiry timestamp")
+	if verificationErr == nil {
+		// Parse the verification expiry timestamp
+		exp, parseErr := time.Parse(TimestampLayout, signup.Annotations[v1alpha1.UserVerificationExpiryAnnotationKey])
+		if parseErr != nil {
+			// If the verification expiry timestamp is corrupt or missing, then return an error
+			verificationErr = errors.NewInternalError(parseErr, "error parsing expiry timestamp")
```
tinakurian, line 229: same as above

pkg/verification/verification_service.go
```diff
@@ -147,56 +189,78 @@ func generateVerificationCode() (string, error) {
...
+		} else if now.After(exp) {
+			// If it is now past the expiry timestamp for the verification code, return a 403 Forbidden error
+			verificationErr = errors.NewForbiddenError("expired", "verification code expired")
```
tinakurian, line 232: same as above

tinakurian (changes requested):

pkg/controller/signup.go
```diff
@@ -52,28 +49,10 @@ func (s *Signup) PostHandler(ctx *gin.Context) {
 	ctx.Writer.WriteHeaderNow()
 }
 
-// UpdateVerificationHandler starts the verification process and updates a usersignup resource. The
+// InitVerificationHandler starts the verification process and updates a usersignup resource. The
```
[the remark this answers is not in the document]
alexeykazakov: Yeah... I wonder if we better rename the other functions here and instead of using http methods in the names we use more meaningful names. For example we could rename `PostHandler()` to `CreateSignup()`.

pkg/controller/signup.go
```diff
@@ -52,28 +47,10 @@ func (s *Signup) PostHandler(ctx *gin.Context) {
...
 // ctx should contain a JSON body in the request with a country_code and a phone_number string
```
alexeykazakov, line 51: Can we add a bit more details about what this function does.

pkg/controller/signup_test.go
```diff
@@ -222,147 +235,150 @@ func (s *TestSignupSuite) TestSignupGetHandler() {
...
+		updatedUserSignup, err := s.FakeUserSignupClient.Get(userSignup.Name)
+		require.NoError(s.T(), err)
 
-		assertVerification(s.T(), storedUserSignup, http.StatusNoContent, rr.Code, userID, "12268213044",
-			"jsmith@redhat.com", "1", storedVerificationCode, verificationInitTimeStamp)
+		require.NotEmpty(s.T(), updatedUserSignup.Annotations[crtapi.UserSignupVerificationCodeAnnotationKey])
+		require.NotEmpty(s.T(), updatedUserSignup.Annotations[crtapi.UserSignupVerificationInitTimestampAnnotationKey])
+		require.NotEmpty(s.T(), updatedUserSignup.Annotations[crtapi.UserVerificationExpiryAnnotationKey])
+		require.Equal(s.T(), "4", updatedUserSignup.Annotations[crtapi.UserSignupVerificationCounterAnnotationKey])
+		require.Equal(s.T(), "fd276563a8232d16620da8ec85d0575f", updatedUserSignup.Labels[crtapi.UserSignupUserPhoneHashLabelKey])
```
alexeykazakov, line 346: The four tests above are almost identical, right? The only difference is the phone format. Can we use some custom assertion for that instead of duplicating the code?

Something like:
```
assertInitVerificationSuccess := func(phone string) {
    gock.New(....)
    ...
    setup()...
    ...
    require....
}

assertInitVerificationSuccess("2 2 6 ...")
assertInitVerificationSuccess("2-26 ...")
....
```

Aslo, I would rename `setup()` to something like `initVerification()` because this function is not only about setting up some test env. It actually does the action we want to test.

pkg/server/in_cluster_application.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/codeready-toolchain/registration-service/pkg/application"
+	"github.com/codeready-toolchain/registration-service/pkg/application/service"
+	servicecontext "github.com/codeready-toolchain/registration-service/pkg/application/service/context"
+	"github.com/codeready-toolchain/registration-service/pkg/application/service/factory"
+	"github.com/codeready-toolchain/registration-service/pkg/configuration"
+	"github.com/codeready-toolchain/registration-service/pkg/kubeclient"
+	"k8s.io/client-go/rest"
+)
+
+func NewInClusterApplication(config configuration.Configuration, options ...factory.Option) (application.Application, error) {
```
alexeykazakov, line 13: Could you add some comments about this particular application implementation.

pkg/verification/service/verification_service.go
```diff
@@ -0,0 +1,265 @@
...
+			log.Error(ctx, err, "usersignup not found")
+			return errors.NewNotFoundError(err, "usersignup not found")
+		}
+		log.Error(ctx, err, "error retrieving usersignup")
+		return errors.NewInternalError(err, fmt.Sprintf("error retrieving usersignup: %s", userID))
+	}
+
+	// check that verification is required before proceeding
+	if signup.Spec.VerificationRequired == false {
+		log.Errorf(ctx, errors.NewForbiddenError("forbidden request", "verification code will not be sent"), "phone verification not required for usersignup: %s", userID)
```
alexeykazakov, line 81: Why logging it as Forbidden? Why not as a bad request which is actually returned to the client?

pkg/verification/service/verification_service.go
```diff
@@ -0,0 +1,265 @@
...
+	var counter int
+	dailyLimit := s.config.GetVerificationDailyLimit()
+	if verificationCounter == "" {
+		counter = 0
+	} else {
+		counter, err = strconv.Atoi(verificationCounter)
+		if err != nil {
+			// We shouldn't get an error here, but if we do, we should probably set verification counter to the daily
+			// limit so that we at least now have a valid value
+			signup.Annotations[v1alpha1.UserSignupVerificationCounterAnnotationKey] = strconv.Itoa(dailyLimit)
```
alexeykazakov, line 130: Let's log the error? Just in case.

pkg/verification/service/verification_service.go
```diff
@@ -0,0 +1,265 @@
...
+
+			// If we get an error here then just die, don't bother updating the UserSignup
+			return errors.NewInternalError(err, "error while sending verification code")
+		}
+	}
+
+	_, updateErr := s.Services().SignupService().UpdateUserSignup(signup)
+	if updateErr != nil {
+		log.Error(ctx, err, "error while updating UserSignup resource")
+		initError = errors.NewInternalError(updateErr, "error while updating UserSignup resource")
```
alexeykazakov, line 166: Still don't like the named returns (IMO it hurts code readability) but we can keep it if you really prefer using them ;)

pkg/verification/service/verification_service.go
```diff
@@ -0,0 +1,265 @@
...
+		// Set a new timestamp
+		signup.Annotations[v1alpha1.UserSignupVerificationTimestampAnnotationKey] = now.Format(TimestampLayout)
+		signup.Annotations[v1alpha1.UserVerificationAttemptsAnnotationKey] = "0"
+	}
+
+	attemptsMade, convErr := strconv.Atoi(signup.Annotations[v1alpha1.UserVerificationAttemptsAnnotationKey])
+	if convErr != nil {
+		// We shouldn't get an error here, but if we do, we will set verification attempts to max allowed
+		// so that we at least now have a valid value, and let the workflow continue to the
+		// subsequent attempts check
```
alexeykazakov, line 213: Same here. Maybe it's worth logging the error just in case.

sbryzak: Log message added.

pkg/verification/service/verification_service.go
sbryzak: I did some more reading up on named return parameters and I don't think there's anything to be afraid of when using them, in fact most people said they helped by documenting the meaning of the return values.  Anyway, I've changed this to use an error variable that's declared up front instead.

pkg/verification/service/verification_service.go
sbryzak: Log message added.

pkg/verification/service/verification_service.go
sbryzak: I'm not actually sure.. this code was taken from here: 

https://github.com/codeready-toolchain/registration-service/blob/master/pkg/controller/signup.go#L73

I'll change it to BadRequest.

sbryzak: Actually, we can just log this here.  No need to log an actual error.

pkg/server/in_cluster_application.go
sbryzak: Done.

pkg/controller/signup.go
sbryzak: I've added more detail here, and also on the `InitVerification()` function in the Verification service.

pkg/controller/signup_test.go
sbryzak: Good suggestions, implemented.

pkg/verification/service/verification_service.go
alexeykazakov: This is why I didn't insist :) Not sure if my code readability point is justified. Maybe it's just a personal preference.

alexeykazakov (approved):

MatousJobanek (changes requested):
Wow, this is huge refactoring - I have gone through only one small part of the code even though I have spent a few hours with it :grimacing:. But it's probably mainly because I'm not really familiar with the original code, so sorry if my questions are not related or just dumb :sweat_smile:. I'll continue on Monday. In the inline comments, you can see my first observations.
To be honest, I'm not really a fan of defining/using options in prod code when they should be used only in the tests - it's pretty hard to understand what we could expect there.

pkg/configuration/configuration.go
```diff
@@ -145,23 +145,51 @@ const (
 	DefaultVerificationCodeExpiresInMin = 5
 )
 
+type Configuration interface {
```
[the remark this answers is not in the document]
sbryzak: The reason for introducing this interface was twofold; firstly it simplified some aspects of the tests, but mainly I was thinking ahead to the enhancements we intend to make for configuration in general.  If we wish to introduce the planned enhancements for configuration (loading config from CRDs etc) then we will need a new implementation, and by that stage it starts to make more sense in my estimation for the consumers of the configuration to work off an interface.

If we were to simply override the environment variables at the beginning of the tests, then wouldn't that need to be rewritten once the configuration was refactored to be based on CRDs?

pkg/auth/keymanager_test.go
```diff
@@ -106,11 +108,11 @@ func (s *TestKeyManagerSuite) TestKeyFetching() {
 		require.NoError(s.T(), err)
 
 		// Set the config for testing mode, the handler may use this.
-		s.Config.GetViperInstance().Set("auth_client.public_keys_url", ts.URL)
-		assert.Equal(s.T(), s.Config.GetAuthClientPublicKeysURL(), ts.URL, "key url not set correctly for testing")
+		s.ViperConfig().GetViperInstance().Set("auth_client.public_keys_url", ts.URL)
```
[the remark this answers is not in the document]
sbryzak: The configuration and application are reset for each test, see:

https://github.com/codeready-toolchain/registration-service/pull/134/files#diff-d74b3288c7ba388e049b90c3afba16411cbdce1ffed9e81892f16140de3a33edR47

pkg/application/service/factory/service_factory.go
```diff
@@ -0,0 +1,78 @@
...
+
+type serviceContextImpl struct {
+	kubeClient kubeclient.CRTClient
+	services   service.Services
+}
+
+func NewServiceContext(kubeClient kubeclient.CRTClient, config configuration.Configuration, options ...Option) servicecontext.ServiceContext {
+	ctx := &serviceContextImpl{kubeClient: kubeClient}
+	var sc servicecontext.ServiceContext
+	sc = ctx
```
tinakurian, line 22: hmm it's a bit strange to create the ctx variable only to re-assign it to a servicecontext

pkg/server/in_cluster_application.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/codeready-toolchain/registration-service/pkg/kubeclient"
+	"k8s.io/client-go/rest"
+)
+
+// NewInClusterApplication creates a new in-cluster application with the specified configuration and options.  This
+// application type is intended to run inside a Kubernetes cluster, where it makes use of the rest.InClusterConfig()
+// function to determine which Kubernetes configuration to use to create the REST client that interacts with the
+// Kubernetes service endpoints.
+func NewInClusterApplication(config configuration.Configuration, options ...factory.Option) (application.Application, error) {
+	app := new(InClusterApplication)
```
tinakurian, line 18: do we really need to create this here? if we error out on line 22 or 27 this was created and never used

pkg/verification/service/verification_service.go
```diff
@@ -0,0 +1,274 @@
...
+		opt(s)
+	}
+	return s
+}
+
+// InitVerification sends a verification message to the specified user, using the Twilio service.  If successful,
+// the user will receive a verification SMS.  The UserSignup resource is updated with a number of annotations in order
+// to manage the phone verification process and protect against system abuse.
+func (s *ServiceImpl) InitVerification(ctx *gin.Context, userID string, e164PhoneNumber string) error {
+	var initError error
```
tinakurian, line 70: i'm in favour or just returning the error rather than re-assigning it and waiting until the end to return it. but in any case this is only used on line 144 the first time. maybe we could define it there instead?

pkg/verification/service/verification_service.go
```diff
@@ -0,0 +1,274 @@
...
+		signup.Annotations[v1alpha1.UserSignupVerificationCounterAnnotationKey] = "0"
+	}
+
+	// get the verification counter (i.e. the number of times the user has initiated phone verification within
+	// the last 24 hours)
+	verificationCounter := signup.Annotations[v1alpha1.UserSignupVerificationCounterAnnotationKey]
+	var counter int
+	dailyLimit := s.config.GetVerificationDailyLimit()
+	if verificationCounter == "" {
+		counter = 0
```
tinakurian, line 127: can't we set counter to 0 and remove some of this nesting? 
```
counter := 0
if verifiationCounter != "" {
   counter, err = strconv.Atoi(verificationCounter)
   ...
}
```

pkg/verification/service/verification_service.go
```diff
@@ -0,0 +1,274 @@
...
+
+			// If we get an error here then just die, don't bother updating the UserSignup
+			return errors.NewInternalError(err, "error while sending verification code")
+		}
+	}
+
+	_, updateErr := s.Services().SignupService().UpdateUserSignup(signup)
+	if updateErr != nil {
+		log.Error(ctx, err, "error while updating UserSignup resource")
+		initError = errors.NewInternalError(updateErr, "error while updating UserSignup resource")
```
tinakurian, line 172: nothing seems to happen after this error occurs, we just end up returning it later. any reason why we can't just return this error here instead of assigning it to initError?

pkg/verification/service/verification_service.go
```diff
@@ -0,0 +1,274 @@
...
+		// If the code matches then set VerificationRequired to false, reset other verification annotations
+		signup.Spec.VerificationRequired = false
+		delete(signup.Annotations, v1alpha1.UserSignupVerificationCodeAnnotationKey)
+		delete(signup.Annotations, v1alpha1.UserVerificationAttemptsAnnotationKey)
+		delete(signup.Annotations, v1alpha1.UserSignupVerificationCounterAnnotationKey)
+		delete(signup.Annotations, v1alpha1.UserSignupVerificationTimestampAnnotationKey)
+		delete(signup.Annotations, v1alpha1.UserSignupVerificationInitTimestampAnnotationKey)
+		delete(signup.Annotations, v1alpha1.UserVerificationExpiryAnnotationKey)
+	} else {
+		log.Error(ctx, verificationErr, "error validating verification code")
```
tinakurian, line 263: we log this error here but we never return it - is there a reason for this? later, verificationErr is re-assigned so the fact that this error even occurred will be lost..

pkg/verification/service/verification_service.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+
+	// Update changes made to the UserSignup
+	_, updateErr := s.Services().SignupService().UpdateUserSignup(signup)
+	if updateErr != nil {
+		log.Error(ctx, updateErr, "error updating UserSignup")
+		verificationErr = errors.NewInternalError(updateErr, "error updating UserSignup")
```
tinakurian, line 270: why do we re-assign the error at this point? it doesn't look like anything is done with verificationErr after it's re-assigned.

tinakurian (changes requested):

pkg/configuration/configuration.go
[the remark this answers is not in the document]
MatousJobanek: Ok, I didn't think about the CRDs usage. That's a good point. 
We will see what it will look like and if we will query the `HostOperatorConfig` CR from the registration service or if we will still use the `ConfigMap` for this particular deployment. 
We will need to change the code in host-operator deployment anyway and make the usage of the new CR consistent in both repos (host-operator & reg-service), so I was thinking that if we had the code similar now, then the future changes would be easier.

Let's keep it as it is - the code is already done, working and looks fine - we can change it later when we start using the CRD

pkg/auth/keymanager_test.go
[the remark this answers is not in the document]
MatousJobanek: yeah, but doesn't apply for subtests. My point is that the subtests that are at the same level shouldn't influence each other.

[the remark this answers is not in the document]
sbryzak: Ah I see - I didn't add this code though, this test is legacy from before the refactoring.  See: 

https://github.com/codeready-toolchain/registration-service/blob/master/pkg/auth/keymanager_test.go#L109

pkg/server/server.go
```diff
@@ -7,22 +7,35 @@ import (
...
 
+type ServerOption = func(server *RegistrationServer)
+
+func WithApplication(app application.Application) ServerOption {
+	return func(server *RegistrationServer) {
+		server.applicationProducerFunc = func() application.Application {
+			return app
+		}
+	}
+}
```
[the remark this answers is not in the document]
sbryzak: I've gone one step further here and now just pass in the application to the registration server constructor.  I think there's probably no need for a producer function here.

pkg/server/in_cluster_application.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/codeready-toolchain/registration-service/pkg/configuration"
+	"github.com/codeready-toolchain/registration-service/pkg/kubeclient"
+	"k8s.io/client-go/rest"
+)
+
+// NewInClusterApplication creates a new in-cluster application with the specified configuration and options.  This
+// application type is intended to run inside a Kubernetes cluster, where it makes use of the rest.InClusterConfig()
+// function to determine which Kubernetes configuration to use to create the REST client that interacts with the
+// Kubernetes service endpoints.
+func NewInClusterApplication(config configuration.Configuration, options ...factory.Option) (application.Application, error) {
```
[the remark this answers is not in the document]
sbryzak: Done

pkg/application/service/factory/service_factory.go
```diff
@@ -0,0 +1,78 @@
...
+	servicecontext "github.com/codeready-toolchain/registration-service/pkg/application/service/context"
+	"github.com/codeready-toolchain/registration-service/pkg/configuration"
+)
+
+type serviceContextImpl struct {
+	kubeClient kubeclient.CRTClient
+	services   service.Services
+}
+
+func NewServiceContext(kubeClient kubeclient.CRTClient, config configuration.Configuration, options ...Option) servicecontext.ServiceContext {
```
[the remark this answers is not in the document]
sbryzak: Good catch, removed.

pkg/server/in_cluster_application.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	kubeClient, err := kubeclient.NewCRTRESTClient(k8sConfig, config.GetNamespace())
+	if err != nil {
+		return nil, err
+	}
+
+	app.serviceFactory = factory.NewServiceFactory(func() servicecontext.ServiceContext {
+		return factory.NewServiceContext(kubeClient, config)
+	}, config, options...)
```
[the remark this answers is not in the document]
sbryzak: You're right, and I actually have a chicken and egg problem here.. I'll think of some way I can reuse the same `ServiceFactory` instance.

pkg/verification/service/verification_service.go
```diff
@@ -0,0 +1,274 @@
...
+		signup.Labels = map[string]string{}
+	}
+	signup.Labels[v1alpha1.UserSignupUserPhoneHashLabelKey] = phoneHash
+
+	// read the current time
+	now := time.Now()
+
+	// If 24 hours has passed since the verification timestamp, then reset the timestamp and verification attempts
+	ts, parseErr := time.Parse(TimestampLayout, signup.Annotations[v1alpha1.UserSignupVerificationInitTimestampAnnotationKey])
+	if parseErr != nil || (parseErr == nil && now.After(ts.Add(24*time.Hour))) {
```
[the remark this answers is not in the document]
sbryzak: Good catch, fixed.

pkg/verification/service/verification_service.go
```diff
@@ -0,0 +1,274 @@
...
+		signup.Annotations[v1alpha1.UserSignupVerificationCodeAnnotationKey] = verificationCode
+		signup.Annotations[v1alpha1.UserVerificationExpiryAnnotationKey] = now.Add(
+			time.Duration(s.config.GetVerificationCodeExpiresInMin()) * time.Minute).Format(TimestampLayout)
+
+		// Generate the verification message with the new verification code
+		content := fmt.Sprintf(s.config.GetVerificationMessageTemplate(), verificationCode)
+		client := twilio.NewClient(s.config.GetTwilioAccountSID(), s.config.GetTwilioAuthToken(), s.HttpClient)
+		msg, err := client.Messages.SendMessage(s.config.GetTwilioFromNumber(), e164PhoneNumber, content, nil)
+		if err != nil {
+			log.Error(ctx, err, fmt.Sprintf("error while sending, code: %d message: %s", msg.ErrorCode, msg.ErrorMessage))
```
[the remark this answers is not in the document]
sbryzak: It might be nil... I've added a conditional here with an alternative log message.

pkg/verification/service/verification_service.go
```diff
@@ -0,0 +1,274 @@
...
+		log.Error(ctx, err, "error while updating UserSignup resource")
+		initError = errors.NewInternalError(updateErr, "error while updating UserSignup resource")
+	}
+
+	return initError
+}
+
+func generateVerificationCode() string {
+	buf := make([]byte, codeLength)
+	rand.Read(buf)
```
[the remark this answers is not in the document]
sbryzak: There's no conditions that can cause an error here, so it's not handled.

pkg/verification/service/verification_service.go
```diff
@@ -0,0 +1,274 @@
...
+		}
+		log.Error(ctx, lookupErr, "error retrieving usersignup")
+		return errors.NewInternalError(lookupErr, fmt.Sprintf("error retrieving usersignup: %s", userID))
+	}
+
+	now := time.Now()
+
+	// If 24 hours has passed since the verification timestamp, then reset the timestamp and verification attempts
+	ts, parseErr := time.Parse(TimestampLayout, signup.Annotations[v1alpha1.UserSignupVerificationTimestampAnnotationKey])
+	if parseErr != nil || (parseErr == nil && now.After(ts.Add(24*time.Hour))) {
```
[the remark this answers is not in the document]
sbryzak: Fixed, thanks.

pkg/verification/service/verification_service.go
```diff
@@ -0,0 +1,274 @@
...
+			// If we get an error here then just die, don't bother updating the UserSignup
+			return errors.NewInternalError(err, "error while sending verification code")
+		}
+	}
+
+	_, updateErr := s.Services().SignupService().UpdateUserSignup(signup)
+	if updateErr != nil {
+		log.Error(ctx, err, "error while updating UserSignup resource")
+		initError = errors.NewInternalError(updateErr, "error while updating UserSignup resource")
+	}
```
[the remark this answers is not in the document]
sbryzak: We don't want to update `UserSignup` if the verification code fails to send.  Likewise, we would also prefer not to send the verification code if the `UserSignup` fails to update, however this scenario is much less likely than the verification code failing to send.

pkg/verification/service/verification_service.go
```diff
@@ -0,0 +1,274 @@
...
+			signup.Annotations[v1alpha1.UserSignupVerificationCounterAnnotationKey] = strconv.Itoa(dailyLimit)
+			counter = dailyLimit
+		}
+	}
+
+	// check if counter has exceeded the limit of daily limit - if at limit error out
+	if counter >= dailyLimit {
+		log.Error(ctx, err, fmt.Sprintf("%d attempts made. the daily limit of %d has been exceeded", counter, dailyLimit))
+		initError = errors.NewForbiddenError("daily limit exceeded", "cannot generate new verification code")
+	}
```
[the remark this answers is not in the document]
sbryzak: No, because the `UserSignup` may have had changes made to it requiring an update.

pkg/application/service/factory/service_factory.go
sbryzak: I agree, but this quirk is required for this code to work.

pkg/server/in_cluster_application.go
sbryzak: Good point, I've moved it further down.

pkg/verification/service/verification_service.go
sbryzak: No problem, moved it down further in the code.

pkg/verification/service/verification_service.go
sbryzak: Sure, fixed.

pkg/verification/service/verification_service.go
sbryzak: The error is returned further down, assuming that the `UserSignup` update doesn't return an error itself.

pkg/verification/service/verification_service.go
sbryzak: Errors while updating `UserSignup` take priority, which is why it's re-assigned before being returned at the end of the function.

pkg/verification/service/verification_service.go
sbryzak: Isn't this your code? ;)  All I did here was change the variable name.

pkg/application/service/base/base_service.go
```diff
@@ -0,0 +1,11 @@
...
+
+import "github.com/codeready-toolchain/registration-service/pkg/application/service/context"
+
+type BaseService struct {
+	context.ServiceContext
+}
+
+func NewBaseService(context context.ServiceContext) BaseService {
+	return BaseService{context}
+}
```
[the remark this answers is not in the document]
sbryzak: It does just hold the service context for now... give me some time to think over whether we really need this.

pkg/verification/service/verification_service.go
tinakurian: does it matter whose code it is :-) i think it can be done better and if we're re-factoring may as well do so now, right?

pkg/verification/service/verification_service.go
tinakurian: oh, i forgot, this is because we want to update the usersignup even if the verification failed. instead of having error messages take precedence, maybe we really should be returning an array of errors? as a user, if the validation failed, i'd want to know that because it's something i can fix on my end. the usersignup update failing isn't something that's in my control though...

pkg/verification/service/verification_service.go
tinakurian: i guess my point is, since this error takes precedence, why even bother re-assigning? may as well just return?

pkg/auth/keymanager_test.go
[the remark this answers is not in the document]
MatousJobanek: yeah, I know... let's discuss it later and keep it as it is :+1:

pkg/verification/service/verification_service.go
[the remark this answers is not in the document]
MatousJobanek: ok, fair enough :+1:

pkg/verification/service/verification_service.go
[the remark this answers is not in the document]
MatousJobanek: ah, right. I missed that part :man_facepalming:

pkg/verification/service/verification_service.go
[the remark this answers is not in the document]
MatousJobanek: yeah, you are probably right, but since the error is handled also in the example in the goland doc https://golang.org/pkg/crypto/rand/ then I thought that it could be worth doing it here as well.
But it's up to you

pkg/verification/service/verification_service.go
```diff
@@ -0,0 +1,277 @@
...
+
+	// If 24 hours has passed since the verification timestamp, then reset the timestamp and verification attempts
+	ts, parseErr := time.Parse(TimestampLayout, signup.Annotations[v1alpha1.UserSignupVerificationTimestampAnnotationKey])
+	if parseErr != nil || (now.After(ts.Add(24 * time.Hour))) {
+		// Set a new timestamp
+		signup.Annotations[v1alpha1.UserSignupVerificationTimestampAnnotationKey] = now.Format(TimestampLayout)
+		signup.Annotations[v1alpha1.UserVerificationAttemptsAnnotationKey] = "0"
+	}
+
+	attemptsMade, convErr := strconv.Atoi(signup.Annotations[v1alpha1.UserVerificationAttemptsAnnotationKey])
```
MatousJobanek, line 218: I maybe got lost in the usage of the annotations. 
In the method `InitVerification` that initiates the verification phase and sets/sends the expected verification code, there you use:
* UserSignupVerificationInitTimestampAnnotationKey
* UserSignupVerificationCounterAnnotationKey

The purpose is clear - we don't want to let users to generate the code too many times in a day.

In this method `VerifyCode` that expects the generated code from the previous step, there you use:
* UserSignupVerificationTimestampAnnotationKey
* UserVerificationAttemptsAnnotationKey

I guess that the purpose is allowing the user to verify the same code only a few times and not infinitely, right? If yes then I have these questions:
1. why don't we reset the counter `UserVerificationAttemptsAnnotationKey` when we generate a new verification code?
2. why do we check the 24 hours timeout when the code has validity only a few minutes?
3. why do we have the `UserSignupVerificationTimestampAnnotationKey` at all? The time limitation should be done by the code expiry, shouldn't it?

pkg/verification/service/verification_service.go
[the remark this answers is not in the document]
sbryzak: I'd prefer to leave this out because it would be impossible to test, therefore it couldn't be included in the test coverage.

pkg/application/service/factory/service_factory.go
sbryzak: I've done some refactoring here which has allowed me to remove this code altogether.

pkg/verification/service/verification_service.go
sbryzak: My point here isn't about whose code this is, it's that this is not new code that has been added for the purposes of this PR.  Making this change here is out of scope of the PR however I've gone ahead and simplified it by removing the first condition.

pkg/application/service/base/base_service.go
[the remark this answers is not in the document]
sbryzak: After some thought I'd like to keep this here as a formality.  Every single service requires the `ServiceContext`, and it's good practice to extend the `BaseService` just in case there's some framework-level changes required in the future.

pkg/server/in_cluster_application.go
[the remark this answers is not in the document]
sbryzak: This was a tricky one, I had to spend most of yesterday and today working through this problem because there was no easy solution.  I've ended up refactoring a significant amount of code as a result, and now there is only a single service factory created and reused for the service contexts.

pkg/configuration/configuration.go
```diff
@@ -145,23 +145,51 @@ const (
...
+type ViperConfig struct {
 	v               *viper.Viper
 	secretValues    map[string]string
 	excludedDomains []string
 }
 
 // CreateEmptyConfig creates an initial, empty config.
-func CreateEmptyConfig(cl client.Client) (*Config, error) {
+func CreateEmptyConfig(cl client.Client) (*ViperConfig, error) {
 	os.Setenv("HOST_OPERATOR_SECRET_NAME", "host-operator-secret")
```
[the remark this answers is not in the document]
sbryzak: Agreed, let's deal with it in the configuration enhancement.  Resolving this conversation for now.

pkg/verification/service/verification_service.go
sbryzak: We limit the total number of attempts allowed per day, regardless of how many times the user has requested a verification code.  The `UserSignupVerificationTimestampAnnotationKey` annotation is used in conjunction with `UserVerificationAttemptsAnnotationKey` in order to track the number of verification attempts a user has made within a 24 hour period.  Once they exceed the maximum allowed, they can no longer attempt to verify until the 24 hour period has lapsed.

pkg/verification/service/verification_service.go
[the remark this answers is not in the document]
MatousJobanek: I wouldn't really care about the test coverage in this case - IMHO proper error handling has higher priority

pkg/application/service/base/base_service.go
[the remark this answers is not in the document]
MatousJobanek: Ok

pkg/verification/service/verification_service.go
MatousJobanek: ok, but wouldn't we achieve the same result if we allowed for example one attempt per verification code?

Apart from that - looking at the default values in the config:
`DefaultVerificationDailyLimit = 5`
`DefaultVerificationAttemptsAllowed = 3`
which means that I can request the verification code 5 times in a day, but can try to verify the code only 3 times...? Imagine this situation:
1. I request a verification code
2. I try to verify the code 3 times but for some reason it fails every time
3. I can still request a new code - actually four new codes - but cannot verify any of them

That doesn't make any sense :man_shrugging:

sbryzak: I think these values should probably be swapped around, that would seem to make more sense then.

MatousJobanek: but still, it will be the same UX - you can do multiple verification of the same code - even if we swap the values - still you could run out of the attempts for the first one or two codes for the day, then no attempt will be left for the last generated code.

We should either use the same annotation/constant for both cases (generation and attempts) or limit the number of attempts per code, not per day.

alexeykazakov: Yeah.. I think it's reasonable to either:
- Count the overall number of verification code generations/sms per day.
- Count the number of verification attempts per code (not overall).
Or:
- Simplify the logic and use one parameter for both the code generation and overall attempts per day.

The first one is preferable but I think it's fine to keep it simple if it really simplify the code.

sbryzak: Guys, let's deal with this in a followup issue as we've gotten off track here.  This PR is about the architectural refactoring only, and while I agree that we should look more closely at the verification workflow business logic it is out of scope of this PR.

alexeykazakov: Sure. Let's address this in a separate PR.

pkg/verification/service/verification_service.go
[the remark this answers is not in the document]
sbryzak: It makes sense that the Golang doc example shows error handling, because that is exactly what would be required in a typical scenario.  With the way we are using it though the code is 100% deterministic.  If we want to build in error handling for some impossible scenario, then we would also need to propagate it within the service layer, and then provide special handling for it there along with a specific HTTP return code, all of which will never be called and can't be tested for.  If we can't trust the Go language to act in a deterministic fashion within this limited scope, then we really can't trust it anywhere.

pkg/verification/service/verification_service.go
MatousJobanek: ok, I'm fine with doing it in separate PR. 
@sbryzak Just please create jira issue for that, so we won't forget, and we can plan it. Thanks

pkg/verification/service/verification_service.go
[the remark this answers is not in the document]
MatousJobanek: Yeah, that make sense. Thanks for this explanation and for the discussion :-) :+1:

pkg/verification/service/verification_service.go
sbryzak: https://issues.redhat.com/browse/CRT-839

pkg/application/service/factory/service_factory.go
```diff
@@ -0,0 +1,117 @@
...
+			f.serviceContextOptions = append(f.serviceContextOptions, opt)
+		}
+	}
+}
+
+func WithServiceContextProducer(producer servicecontext.ServiceContextProducer) func(f *ServiceFactory) {
+	return func(f *ServiceFactory) {
+		f.contextProducer = producer
+	}
+}
```
MatousJobanek, line 78: Correct me if I'm wrong, but this is not used anywhere

pkg/application/service/factory/service_factory.go
```diff
@@ -0,0 +1,117 @@
...
+
+// Option an option to configure the Service Factory
+type Option func(f *ServiceFactory)
+
+func WithServiceContextOptions(opts ...ServiceContextOption) func(f *ServiceFactory) {
+	return func(f *ServiceFactory) {
+		for _, opt := range opts {
+			f.serviceContextOptions = append(f.serviceContextOptions, opt)
+		}
```
MatousJobanek, line 70: I guess that you can replace it with one simple append command:
```suggestion
		f.serviceContextOptions = append(f.serviceContextOptions, opts...)
```

MatousJobanek (approved):
The refactoring looks good - the code is much cleaner and easier to read. Good job @sbryzak :+1: :1st_place_medal: and thanks a lot for taking extra time and addressing my comments :pray:. 
There are still a few things I'd like to change, but let's move it for later PRs.
Please don't forget to create a Jira for fixing the number of verification attempts.

MatousJobanek:
> Please don't forget to create a Jira for fixing the number of verification attempts.

you already did it - I didn't update the page, so didn't see your comment. Thanks a lot :+1:

pkg/application/service/factory/service_factory.go
sbryzak: You're right, I think it's safe to remove this now.

pkg/application/service/factory/service_factory.go
sbryzak: Nice suggestion, I didn't realise the append function was a variadic.

PR #138, opened 2020-10-26
Adjust workflow business rules for verification

sbryzak:
As per our previous discussion [here](https://github.com/codeready-toolchain/registration-service/pull/134#discussion_r508872883), I've adjusted the verification workflow so that attempts made are only counted towards the currently generated verification code.

Fixes https://issues.redhat.com/browse/CRT-839


MatousJobanek (approved):
Looks good - thanks for changing it. But please adjust the tests - there are still some that use the `UserSignupVerificationTimestampAnnotationKey` constant.

PR #140, opened 2020-11-16
adding preflight check

tinakurian:
Adding a preflight check to avoid issues when endpoints are being accessed from different domains.


pkg/server/server.go
```diff
@@ -35,6 +35,13 @@ func New(config configuration.Configuration, application application.Application
 	ginRouter.Use(
 		gin.LoggerWithWriter(gin.DefaultWriter, "/api/v1/health"),
 		gin.Recovery(),
+		cors.New(cors.Config{
+			AllowOrigins:     []string{"*"},
```
alexeykazakov, line 39: You can use AllowAllOrigins for that too:
```suggestion
			AllowAllOrigins:     true,
```

pkg/server/server.go
```diff
@@ -35,6 +35,13 @@ func New(config configuration.Configuration, application application.Application
...
+			AllowMethods:     []string{"*"},
```
alexeykazakov, line 40: Hm... "*" doesn't seem to be always treated as equivalent of all methods... https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Methods
I wonder if we should use 
```suggestion
			AllowMethods:     []string{"PUT", "PATCH", "POST", "GET", "DELETE", "OPTIONS"},
```
instead?

pkg/server/server.go
```diff
@@ -35,6 +35,13 @@ func New(config configuration.Configuration, application application.Application
...
+			AllowHeaders:     []string{"*"},
```
alexeykazakov, line 41: Same here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers
It seems that we have to explicitly list at least `Authorization` header.

pkg/server/server.go
```diff
@@ -35,6 +35,13 @@ func New(config configuration.Configuration, application application.Application
...
+			ExposeHeaders:    []string{"*"},
```
alexeykazakov, line 42: Same here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers

alexeykazakov (commented):
I'm not sure about using wildcard `*`. See my comments.
In our devcluster we use the following configuration: https://github.com/codeready-toolchain/devcluster/blob/master/pkg/server/server.go#L53-L58 and it seems to be working fine with cors. It can be improved too though.

pkg/server/server.go
tinakurian: ohhh okay, nice!

pkg/server/server.go
```diff
@@ -35,6 +35,13 @@ func New(config configuration.Configuration, application application.Application
 	ginRouter.Use(
 		gin.LoggerWithWriter(gin.DefaultWriter, "/api/v1/health"),
 		gin.Recovery(),
+		cors.New(cors.Config{
+			AllowAllOrigins:     true,
```
alexeykazakov, line 39:
```suggestion
			AllowAllOrigins:  true,
```

alexeykazakov (commented):
Looks good. Just need to fix the formatting.
And can we add some tests for that?

tinakurian:
> Looks good. Just need to fix the formatting.
> And can we add some tests for that?

@alexeykazakov Hmm I'm not too sure how we would do that, any ideas? I used a package called cors.

alexeykazakov:
@tinakurian I was thinking if we can start an actual gin server in the test and call OPTIONS method if it's not too tricky.

tinakurian:
> @tinakurian I was thinking if we can start an actual gin server in the test and call OPTIONS method if it's not too tricky.

@alexeykazakov  good idea! will give that a try!

pkg/server/server_test.go
```diff
@@ -36,3 +37,41 @@ func (s *TestServerSuite) TestServer() {
 	// Check that Engine() returns the router object.
 	require.NotNil(s.T(), srv.Engine())
 }
+
+func (s *TestServerSuite) TestCrossCors() {
```
alexeykazakov, line 41:
```suggestion
func (s *TestServerSuite) TestCors() {
```

pkg/server/server_test.go
```diff
@@ -36,3 +37,41 @@ func (s *TestServerSuite) TestServer() {
...
+	// Setting up the routes.
+	err := srv.SetupRoutes()
+	require.NoError(s.T(), err)
+
+	// Check that there are routes registered.
+	routes := srv.GetRegisteredRoutes()
+	require.NotEmpty(s.T(), routes)
+
+	// Check that Engine() returns the router object.
+	require.NotNil(s.T(), srv.Engine())
```
alexeykazakov, line 55: This is just a copy of `TestServer()`.
Let's add the cors test as a subtest of `TestServer()`:

```
func (s *TestServerSuite) TestServer() {
    ...
    s.T().Run("CORS", func(t *testing.T) {
        // Tests CORS assuming that the server and the routes already setup and tested above
        ...
    })
)

File: pkg/server/server_test.go [hunk 8]
```diff
@@ -36,3 +37,41 @@ func (s *TestServerSuite) TestServer() {
...
+
+	go srv.Engine().Run()
```
alexeykazakov, line 57: Shouldn't we wait for the server to be available or something before proceeding with the tests?

tinakurian: I don't think so: https://github.com/gin-gonic/gin/blob/master/gin.go#L304

pkg/server/server_test.go
tinakurian: fixed here: 5cc1588595e2fc5e68c5173d46d3c495d55db363

pkg/server/server_test.go
alexeykazakov: You are starting a new goroutine which will be blocked. But the test goroutine will proceed. This is exactly why you need to start a new goroutine here so the test is not blocked. But it takes some time to start the server and it might happen then when you make that OPTIONS request the server is not started yet.

pkg/server/server_test.go
```diff
@@ -20,7 +21,7 @@ func TestRunServerSuite(t *testing.T) {
 	suite.Run(t, &TestServerSuite{test.UnitTestSuite{}})
 }
 
-func (s *TestServerSuite) TestServer() {
+func (s *TestServerSuite) TestCors() {
```
alexeykazakov, line 24: Since the cors test is just a subtest of that bigger server test it would make sense to keep the original test name :)

```suggestion
func (s *TestServerSuite) TestServer() {
```

tinakurian: accidentally happened when i merged :)

alexeykazakov (approved):
looks good!

PR #141, opened 2020-11-17
Load Console and Che URLs from ToolchainStatus

alexeykazakov:
After we switched from Che to CRW in our crt-stage clusters we stuck with old Che route/URL in MUR status.
With this PR we stop using MUR status for retrieving those URLs and use ToolchainSatus instead.

There is no e2e tests to update since we do not have Che installed in our e2e tests. The toolchain-status retrieval by the reg-service is automatically covered by the existing e2e tests.

### DIFF
--- a/pkg/signup/service/signup_service.go
+++ b/pkg/signup/service/signup_service.go
@@ -174,7 +174,7 @@ func (s *ServiceImpl) GetSignup(userID string) (*signup.Signup, error) {
 		// Retrieve Console and Che dashboard URLs from the status of the corresponding member cluster
 		status, err := s.CRTClient().V1Alpha1().ToolchainStatuses().Get()
 		if err != nil {
-			return nil, errors2.Wrap(err, fmt.Sprintf("error when retrieving ToolchainStatus to set Che Dashboard for completed UserSignup %s", userSignup.GetName()))
+			return nil, errors2.Wrapf(err, "error when retrieving ToolchainStatus to set Che Dashboard for completed UserSignup %s", userSignup.GetName())
 		}
 		for _, member := range status.Status.Members {
 			if member.ClusterName == mur.Status.UserAccounts[0].Cluster.Name {
--- a/pkg/signup/service/signup_service.go
+++ b/pkg/signup/service/signup_service.go
@@ -171,9 +171,18 @@ func (s *ServiceImpl) GetSignup(userID string) (*signup.Signup, error) {
 		VerificationRequired: userSignup.Spec.VerificationRequired,
 	}
 	if mur.Status.UserAccounts != nil && len(mur.Status.UserAccounts) > 0 {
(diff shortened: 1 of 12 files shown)


rajivnathan (approved):
Nice!

alexeykazakov:
Tests will fail until https://github.com/codeready-toolchain/host-operator/pull/320 is merged.

pkg/signup/service/signup_service.go
```diff
@@ -171,9 +171,18 @@ func (s *ServiceImpl) GetSignup(userID string) (*signup.Signup, error) {
 		VerificationRequired: userSignup.Spec.VerificationRequired,
 	}
 	if mur.Status.UserAccounts != nil && len(mur.Status.UserAccounts) > 0 {
-		// TODO Set ConsoleURL in UserSignup.Status. For now it's OK to get it from the first embedded UserAccount status from MUR.
-		signupResponse.ConsoleURL = mur.Status.UserAccounts[0].Cluster.ConsoleURL
-		signupResponse.CheDashboardURL = mur.Status.UserAccounts[0].Cluster.CheDashboardURL
+		// Retrieve Console and Che dashboard URLs from the status of the corresponding member cluster
+		status, err := s.CRTClient().V1Alpha1().ToolchainStatuses().Get()
+		if err != nil {
+			return nil, errors2.Wrap(err, fmt.Sprintf("error when retrieving ToolchainStatus to set Che Dashboard for completed UserSignup %s", userSignup.GetName()))
```
xcoulon, line 177: I believe you could use `errors2.Wrapf` to avoid the use of `fmt.Sprintf`, but it's not critical

test/fake/toolchainstatus_client.go
```diff
@@ -0,0 +1,72 @@
...
+	namespace string
+	MockGet   func() (*crtapi.ToolchainStatus, error)
+}
+
+func NewFakeToolchainStatusClient(namespace string, initObjs ...runtime.Object) *FakeToolchainStatusClient {
+	clientScheme := runtime.NewScheme()
+	err := crtapi.SchemeBuilder.AddToScheme(clientScheme)
+	if err != nil {
+		log.Error(err, "Error adding to scheme")
+		os.Exit(1)
```
xcoulon, line 26: are you sure you want to `os.Exit`? Should you call something like `t.Fatal` instead?

pkg/signup/service/signup_service.go
alexeykazakov: Yep. Changed it in https://github.com/codeready-toolchain/registration-service/pull/141/commits/7f1d753c9461c654e2fe787a9164e73c45d9225a

test/fake/toolchainstatus_client.go
alexeykazakov: It was a copy-paste. Did some small refactoring for that: https://github.com/codeready-toolchain/registration-service/pull/141/commits/7f1d753c9461c654e2fe787a9164e73c45d9225a

PR #144, opened 2020-11-20
UserSignup reactivation

alexeykazakov:
`POST /api/v1/signup` should reactivate the usersignup if it's currently deactivated.
If there is no usersignup at all then it creates a new one.

**Bonus:** for some weird reason keycloak.IdToken is reset by keycloak.js in our landing page after successful authentication. This PR stores the token so it can be used in the subsequent calls. Otherwise phone verification fails because Authorization header is not added by UI.

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/197


pkg/signup/service/signup_service.go
```diff
@@ -116,9 +124,49 @@ func extractEmailHost(email string) string {
...
+		return s.reactivateUserSignup(ctx)
+	}
+
+	username := ctx.GetString(context.UsernameKey)
+	return nil, errors2.Errorf("unable to create UserSignup [id: %s; username: %s] because there is already an active UserSignup with such ID", userID, username)
+}
+
+// reactivateUserSignup reactivates the deactivated UserSignup resource with the specified username and userID
+func (s *ServiceImpl) reactivateUserSignup(ctx *gin.Context) (*v1alpha1.UserSignup, error) {
+	userSignup, err := s.newUserSignup(ctx)
```
MatousJobanek, line 154: If I understand it correctly - we need to create the new version of UserSignup here and cannot use the retrieved from this line [132](https://github.com/codeready-toolchain/registration-service/pull/144/files#diff-044c9e6475638d41c551f545f3899df6a42b37489faa3c9f7eb87edfa87b79b0R132) just to be sure that the reactivated UserSignup:
* is not banned
* uses the latest annotations & labels and other fields

Is that correct? I would propose adding some comment to this line to make it clear.

pkg/signup/service/signup_service_test.go
```diff
@@ -98,32 +130,50 @@ func (s *TestSignupServiceSuite) TestCreateUserSignup() {
...
 
-	userSignups := values.(*v1alpha1.UserSignupList)
-	require.NotEmpty(s.T(), userSignups.Items)
-	require.Len(s.T(), userSignups.Items, 1)
+	s.Run("deactivate and reactivate again", func() {
+		// given
+		deactivatedUS := existing
+		deactivatedUS.Spec.Deactivated = true
+		deactivatedUS.Status.Conditions = deactivated()
+		err := s.FakeUserSignupClient.Tracker.Update(gvr, &deactivatedUS, s.Config().GetNamespace())
```
MatousJobanek, line 146: It's probably just a detail in this case, but it's safer to use the deep copy of the object:
```suggestion
		deactivatedUS := existing.DeepCopy()
		deactivatedUS.Spec.Deactivated = true
		deactivatedUS.Status.Conditions = deactivated()
		err := s.FakeUserSignupClient.Tracker.Update(gvr, deactivatedUS, s.Config().GetNamespace())
```

pkg/signup/service/signup_service_test.go
```diff
@@ -98,32 +130,50 @@ func (s *TestSignupServiceSuite) TestCreateUserSignup() {
...
+		require.NoError(s.T(), err)
+		assertUserSignupExists(userSignup, userID.String())
+	})
+
+	s.Run("deactivate and try to reactivate but reactivation fails", func() {
+		// given
+		deactivatedUS := existing
+		deactivatedUS.Spec.Deactivated = true
+		deactivatedUS.Status.Conditions = deactivated()
+		err := s.FakeUserSignupClient.Tracker.Update(gvr, &deactivatedUS, s.Config().GetNamespace())
```
MatousJobanek, line 162: the same here:
```suggestion
		deactivatedUS := existing.DeepCopy()
		deactivatedUS.Spec.Deactivated = true
		deactivatedUS.Status.Conditions = deactivated()
		err := s.FakeUserSignupClient.Tracker.Update(gvr, deactivatedUS, s.Config().GetNamespace())
```

MatousJobanek (commented):
Nice :+1: Very nice and simple solution :1st_place_medal: 
Just waiting for the e2e tests :-)

pkg/signup/service/signup_service.go
alexeykazakov: During re-activation we basically want to override the existing deactivated resource so we have a new fresh one as it was never existed. The main reason is that we don't really know what else can be stored in the resource which can affect the "new" resource. So, I would avoid patching the existing resource and would just replace it (put vs patch) by the "new" one.
Stuff like checking bans, phone verification, etc is a free bonus here.

pkg/signup/service/signup_service_test.go
alexeykazakov: Right. DeepCopy would be safer.

pkg/signup/service/signup_service.go
```diff
@@ -116,9 +124,49 @@ func extractEmailHost(email string) string {
...
+	if err != nil {
+		return nil, err
+	}
+
+	updated, err := s.CRTClient().V1Alpha1().UserSignups().Update(userSignup)
+	if err != nil {
+		return nil, err
+	}
+
+	return updated, nil
```
alexeykazakov, line 164:
```suggestion
	return s.CRTClient().V1Alpha1().UserSignups().Update(userSignup)
```

pkg/signup/service/signup_service.go
alexeykazakov: Added a comment here: https://github.com/codeready-toolchain/registration-service/pull/144/commits/21738ee1e7b205e57a3a4004713486f8134d48ce

alexeykazakov:
Added e2e tests. The PR is ready for the final review.

rajivnathan:
Looks like it failed with yet another flaky test (TestRegistrationService/TestPhoneVerification). I've added it to our list
/retest

MatousJobanek (approved):

PR #146, opened 2020-11-24
Prevent users from signing up if their username has a forbidden prefix

sbryzak:
This PR will prevent users from signing up if their username has a forbidden prefix.  It introduces a new configuration parameter, `username.forbidden.prefixes` which can be used to control which prefixes are forbidden.  By default, it disallows prefixes of `openshift` or `kubernetes` to be used in usernames.

Fixes https://issues.redhat.com/browse/CRT-783

### DIFF
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -143,6 +143,9 @@ const (
 	// varVerificationCodeExpiresInMin is used to set the amount of time (in minutes) that a verification code is active for before expiring
 	varVerificationCodeExpiresInMin     = "verification.code_expires_in_min"
 	DefaultVerificationCodeExpiresInMin = 5
+
+	varForbiddenUsernamePrefixes     = "forbidden.username.prefixes"
+	DefaultForbiddenUsernamePrefixes = "openshift,kubernetes"
 )
 
 type Configuration interface {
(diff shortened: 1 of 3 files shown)


MatousJobanek (commented):
Looks good :+1:
Could we have e2e test for it?

BTW, wouldn't be easier to just modify the compliant user generation https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/usersignup/usersignup_controller.go#L445-L466 and add some prefix when the problematic username is used? The user experience would be maybe better than just forbidding users.

alexeykazakov (commented):
Looks good but yes, we need e2e tests for that.
Also not sure what is better here.. do the username modification as @MatousJobanek is suggesting or return a bad request error as it's done in this PR...

sbryzak:
I'm wondering what prefix we could append to the front of the non-compliant username in order to make it compliant.  Any ideas?

MatousJobanek:
currently, we already add `crt` prefix (or also a suffix) in case it is needed to generate a compliant username - see https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/usersignup/usersignup_controller.go#L445-L466
we could use the same - just add another if statement to the function.
If you wanted to have something closer to `sandbox` then we could replace the `crt-`, and use `sand-` as the prefix and `-box` as a suffix :smile:

MatousJobanek:
@sbryzak I see that you pushed a commit that adds the prefix here c8c7f49 but that should be rather in host-operator where all the logic is, not here.

alexeykazakov (commented):
Yeah. I agree with @MatousJobanek.
If we do the prefix thing that we should drop this PR in reg-service and just add this logic to the usersignup controller: https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/usersignup/usersignup_controller.go#L445-L466

PR #148, opened 2020-11-24
Segment and woopra endpoints

tinakurian:
Jira: https://issues.redhat.com/browse/CRT-853
Related PR: codeready-toolchain/host-operator#323, https://github.com/codeready-toolchain/toolchain-e2e/pull/205
This PR creates two new endpoints; one for retrieving the woopra domain and another for retrieving the segment-write-key.


alexeykazakov (commented):
Looks good. Just waiting for the e2e tests.

deploy/registration-service.yaml
```diff
@@ -171,6 +171,16 @@ objects:
...
+                - name: REGISTRATION_WOOPRA_DOMAIN
+                  valueFrom:
+                    configMapKeyRef:
+                      name: registration-service
+                      key: woopra-domain
+                - name: REGISTRATION_SEGMENT_WRITE_KEY
+                  valueFrom:
+                    configMapKeyRef:
+                      name: registration-service
+                      key: segment-write-key
```
MatousJobanek, line 183: could we please keep some consistency in the key strings? Please, use either snake or dot case format - as is used in previous keys

deploy/registration-service.yaml
```diff
@@ -228,6 +238,8 @@ objects:
       verification.excluded_email_domains: ${VERIFICATION_EXCLUDED_EMAIL_DOMAINS}
       verification.message_template: ${VERIFICATION_MESSAGE_TEMPLATE}
       twilio.from_number: ${TWILIO_FROM_NUMBER}
+      woopra.domain: ${WOOPRA_DOMAIN}
+      segment-write-key: ${SEGMENT_WRITE_KEY}
```
MatousJobanek, line 242: The same here.
In addition, the `woopra.domain` even doesn't match with the one that is used at line [178](https://github.com/codeready-toolchain/registration-service/pull/148/files#diff-d34232b191f95b876219f0dc263f2debac6ca94abf4535dd1433546b01a4a8baR178)

pkg/configuration/configuration.go
```diff
@@ -143,6 +143,12 @@ const (
 	// varVerificationCodeExpiresInMin is used to set the amount of time (in minutes) that a verification code is active for before expiring
 	varVerificationCodeExpiresInMin     = "verification.code_expires_in_min"
 	DefaultVerificationCodeExpiresInMin = 5
+
+	// varWoopraDomain contains the woopra domain
+	varWoopraDomain = "woopra_domain"
+
+	// varSegmentWriteKey contains the segment write key
+	varSegmentWriteKey = "segment_write_key"
```
MatousJobanek, line 151: ok, now you use the snake. btw, wouldn't it be good to use the prefix `woopra.` to express what it is used for?

MatousJobanek: in addition, you are missing tests for these values

pkg/controller/segment.go
```diff
@@ -0,0 +1,30 @@
+package controller
+
+import (
+	"github.com/codeready-toolchain/registration-service/pkg/configuration"
+	"github.com/gin-gonic/gin"
+	"net/http"
+)
+
+type segmentResponse struct {
+	SegmentWriteKey string `json:"segment-write-key"`
```
MatousJobanek, line 10: I'm curious - why do we want to use the kebab case for this json key? I would rather prefer using the camelcase for the sake of consistency.

pkg/controller/woopra.go
```diff
@@ -0,0 +1,30 @@
+package controller
+
+import (
+	"github.com/codeready-toolchain/registration-service/pkg/configuration"
+	"github.com/gin-gonic/gin"
+	"net/http"
+)
+
+type woopraResponse struct {
+	WoopraDomain string `json:"woopra-domain"`
```
MatousJobanek, line 10: the same here

pkg/controller/woopra_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+type TestWoopraSuite struct {
+	test.UnitTestSuite
+}
+
+func TestRunWoopraSuite(t *testing.T) {
+	suite.Run(t, &TestAuthConfigSuite{test.UnitTestSuite{}})
+}
+
+func (s *TestAuthConfigSuite) TestWoopraHandler() {
```
MatousJobanek, line 25: I guess that this is a wrong copy-paste - you should use the `TestWoopraSuite` here, not `TestAuthConfigSuite`

pkg/controller/woopra_test.go
```diff
@@ -0,0 +1,62 @@
...
+	// Create a request to pass to our handler. We don't have any query parameters for now, so we'll
+	// pass 'nil' as the third parameter.
+	req, err := http.NewRequest(http.MethodGet, "/api/v1/woopra", nil)
```
MatousJobanek, line 28: shouldn't be the endpoint `/woopra-domain`?

MatousJobanek (changes requested):

pkg/controller/segment.go
tinakurian: this is consistent with what we have in the other files, see here: https://github.com/codeready-toolchain/registration-service/blob/master/pkg/controller/authconfig.go#L11-L14. but i do see that in signup, when the phone verification stuff was added, it brought some inconsistency i guess.

pkg/controller/woopra.go
tinakurian: see comment above.

pkg/configuration/configuration.go
tinakurian: this is consistent with what we have in this file - everything else uses snake here, see here for example: https://github.com/codeready-toolchain/registration-service/blob/master/pkg/configuration/configuration.go#L65

pkg/controller/woopra_test.go
tinakurian: yeah, good catch, fixed here: fd6538e88bf12f20ad946a8f8d04acebd5acbb02

pkg/controller/woopra_test.go
tinakurian: yep, fixed here: fd6538e88bf12f20ad946a8f8d04acebd5acbb02

deploy/registration-service.yaml
tinakurian: updated here: fd6538e88bf12f20ad946a8f8d04acebd5acbb02

deploy/registration-service.yaml
tinakurian: updated here: fd6538e88bf12f20ad946a8f8d04acebd5acbb02

pkg/configuration/configuration.go
MatousJobanek: cool, but you still forgot to add the test for it.

pkg/controller/segment.go
MatousJobanek: I wasn't aware of this file - I was referring to json key format used in all our CRDs...
OK, let's keep it as it is :+1:

deploy/registration-service.yaml
```diff
@@ -257,3 +269,7 @@ parameters:
     value: ''
   - name: TWILIO_FROM_NUMBER
     value: ''
+  - name: WOOPRA_DOMAIN
+    value: ''
+  - name: SEGMENT_WRITE_KEY
+    value: ''
```
MatousJobanek, line 275: This end of the file doesn't correspond to the one in host-operator repo - could you please fix that? https://github.com/codeready-toolchain/host-operator/pull/326/files#diff-43fc4f9268f9dd2811ffed0c88d92da522f5b2c0cb3e2643cf253243e2025276R275

tinakurian: fixed here: 47b068cff45b7c35ac2543646c015ce4c2bb1bcf

pkg/configuration/configuration.go
tinakurian: sorry! wasn't meaning to ignore the test comment! fixed here: 7850984b18856fa004145dbfb3f9de3afc6fbfdd

alexeykazakov (commented):
Looks good! Just couple of minor comments.

pkg/controller/segment_test.go
```diff
@@ -0,0 +1,63 @@
...
+	segmentCtrl := NewSegment(s.Config())
+	handler := gin.HandlerFunc(segmentCtrl.GetSegmentWriteKey)
+
+	s.Run("valid segment write key json", func() {
+
+		// We create a ResponseRecorder (which satisfies http.ResponseWriter) to record the response.
+		rr := httptest.NewRecorder()
+		ctx, _ := gin.CreateTestContext(rr)
+		ctx.Request = req
+		s.ViperConfig().GetViperInstance().Set("segment.write.key", "testing segment write key")
```
alexeykazakov, line 45:
```suggestion
		s.ViperConfig().GetViperInstance().Set("segment.write_key", "testing segment write key")
```

pkg/configuration/configuration.go
```diff
@@ -143,6 +143,12 @@ const (
 	// varVerificationCodeExpiresInMin is used to set the amount of time (in minutes) that a verification code is active for before expiring
 	varVerificationCodeExpiresInMin     = "verification.code_expires_in_min"
 	DefaultVerificationCodeExpiresInMin = 5
+
+	// varWoopraDomain contains the woopra domain
+	varWoopraDomain = "woopra.domain"
+
+	// varSegmentWriteKey contains the segment write key
+	varSegmentWriteKey = "segment.write.key"
```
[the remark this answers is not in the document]
tinakurian: it makes sense to separate parent/child so i've updated here:  ab173028c923ed60f3fafc5730402651f97364e7

we do it for some twilio keys but not for all of them, we should update it at some point for twilio and mailgun so it's consistent. maybe i'll open a jira so that i remember to do that after this is done!

pkg/controller/segment_test.go
tinakurian: fixed here: 79020e69da3af25d68080ad3b576d3da6377538d

alexeykazakov (approved):

pkg/server/routes.go
```diff
@@ -64,6 +64,8 @@ func (srv *RegistrationServer) SetupRoutes() error {
 		// creating the controllers
 		healthCheckCtrl := controller.NewHealthCheck(srv.Config(), controller.NewHealthChecker(srv.Config()))
 		authConfigCtrl := controller.NewAuthConfig(srv.Config())
+		woopraCtrl := controller.NewWoopra(srv.Config())
+		segmentCtrl := controller.NewSegment(srv.Config())
```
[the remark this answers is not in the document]
tinakurian: @alexeykazakov makes sense - what should we call the controller?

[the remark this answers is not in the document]
alexeykazakov: Naming is hard :) But "woopra" or "segment" would work for me...
It's kinda the same in this context.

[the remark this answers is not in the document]
tinakurian: haha agreed :-) okay, will update that!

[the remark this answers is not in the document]
tinakurian: updated here: 6549aaf795fdc76ad5854b4bc9d00ea37ebecc6f

PR #149, opened 2020-11-25
Fix usersignup polling in UI after phone verification

alexeykazakov:
This PR fixes a bug in the landing page when after successful phone verification the state of the signup is not updated automatically when the signup is approved.


tinakurian (approved):
Looks good Alexey! Nice fix!

PR #164, opened 2021-01-11
fix: set pending approval status when the same is set in UserSignup

MatousJobanek:
set pending approval status when the same is set in UserSignup `approved` condition type - no matter what is set in `complete` condition

### DIFF
--- a/pkg/signup/service/signup_service.go
+++ b/pkg/signup/service/signup_service.go
@@ -183,8 +183,9 @@ func (s *ServiceImpl) GetSignup(userID string) (*signup.Signup, error) {
 	}
 
 	// Check UserSignup status to determine whether user signup is complete
-	signupCondition, found := condition.FindConditionByType(userSignup.Status.Conditions, v1alpha1.UserSignupComplete)
-	if !found {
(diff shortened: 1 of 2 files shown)


pkg/signup/service/signup_service.go
```diff
@@ -183,8 +183,9 @@ func (s *ServiceImpl) GetSignup(userID string) (*signup.Signup, error) {
 	}
 
 	// Check UserSignup status to determine whether user signup is complete
-	signupCondition, found := condition.FindConditionByType(userSignup.Status.Conditions, v1alpha1.UserSignupComplete)
-	if !found {
+	approvedCondition, approvedFound := condition.FindConditionByType(userSignup.Status.Conditions, v1alpha1.UserSignupApproved)
+	signupCondition, completeFound := condition.FindConditionByType(userSignup.Status.Conditions, v1alpha1.UserSignupComplete)
```
xcoulon, line 187: to be consistent with the `approvedCondition` variable above, shouldn't we rename `signupCondition` to `completeCondition`?

MatousJobanek: good point, done: 37d3f15

alexeykazakov (approved):

xcoulon (approved):

PR #165, opened 2021-01-13
Prefix invalid subject values with a valid prefix

sbryzak:
This PR sets the new `UserSignup.Spec.UserID` with the value of the `subject` claim from the user's SSO token.  Further changes forthcoming will perform encoding of the subject value for use as the `UserSignup` resource name.

Fixes https://issues.redhat.com/browse/CRT-915


MatousJobanek (commented):
Correct me if I'm wrong: by adding the prefix containing the hash that is counted using the reverse polynomial we will get the positive id (claim)...? How do we know that this will be accepted by SSO when we put this new ID into the identity in the member cluster?

alexeykazakov (changes requested):
As Matous mentioned we can't just replace the actual User ID to something generated when creating UserSignup.
The UserSignup.Name is used to create User and Identity in member cluster (it's passed through MUR): https://github.com/codeready-toolchain/host-operator/blob/c97e93e04fd641af6e9d38d5ea4e470b2691e441/pkg/controller/usersignup/masteruserrecord.go#L71 
If we just code the name without decoding it before creating an actual user then the user won't be able to login because the User IDs won't match.

Also, have you seen my comments in https://issues.redhat.com/browse/CRT-915 ?
Right now we have an issue with IDs which start with `-`. But we plan to switch to regular sso clients (without sub claim override) and UserIDs will also contain `:` (we should support any sub claim with URI scheme, for example it may contain `/`).

So, I see two options here:
- refactor our UserSignup CRD to use a different approach rather then using UserID as the resource name so we can support URI scheme/names). It should solve that issue with `-` too.
- split https://issues.redhat.com/browse/CRT-915 to two issues:
   1) Add something for now to fix the `-` issue our users currently face
   2) Deep UserSignup refactoring to switch to a new approach to support URI scheme

If we split the issue then for the "quick" fix we could use some prefix in UserSignup.Name which we can easily recognize and decode to get the original User ID.
For the URI scheme issue we need to think more... If we could code the entire User ID to some DNS complaint name and decode it 100% correct for all User IDs then it would solve all our problems. If it's impossible then we will need to stop using User IDs as the resource names and come up with something else.

Also, regardless of what we do now we have to add e2e tests for the PR :)

pkg/signup/service/signup_service.go
```diff
@@ -109,6 +110,17 @@ func extractEmailHost(email string) string {
 	return email[i+1:]
 }
 
+// fixSubjectPrefixIfInvalid examines the first character of the specified subject value, and if invalid (i.e. it
+// doesn't conform to the DNS-1123 specification's requirement of starting with an alphanumeric character)
+// it will prefix the value with a CRC32 checksum of the subject's value, returning the result.
+func fixSubjectPrefixIfInvalid(subject string) string {
+	if strings.HasPrefix(subject, "-") {
+		crc32q := crc32.MakeTable(0xEDB88320)
+		return fmt.Sprintf("%x%s", crc32.Checksum([]byte(subject), crc32q), subject)
```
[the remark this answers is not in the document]
sbryzak: Thanks Matous, I wasn't aware of that shortcut.

sbryzak:
So this PR should now perform a one-way encoding of the user ID into a DNS-1123 compliant value.  With this change a number of further changes are required in the host operator, which are forthcoming.

alexeykazakov (commented):
The encoding part looks good but let's add tests for `encodeUserID()` function with all possible combinations to cover that function 100%.

pkg/signup/service/signup_service.go
```diff
@@ -109,6 +119,34 @@ func extractEmailHost(email string) string {
 	return email[i+1:]
 }
 
+// EncodeUserID examines the first character of the specified subject value, and if invalid (i.e. it
```
alexeykazakov, line 122: It examines all the characters. Not just the first one. Doesn't it?

pkg/signup/service/signup_service_test.go
```diff
@@ -177,6 +179,62 @@ func (s *TestSignupServiceSuite) TestSignup() {
...
+	require.Equal(s.T(), expected, val.Name)
+	require.False(s.T(), strings.HasPrefix(val.Name, "-"))
+}
+*/
+
+func (s *TestSignupServiceSuite) TestEncodeUserID() {
+	s.Run("test valid user ID unchanged", func() {
+		userID := "abcde-12345"
+
+		encoded := service.EncodeUserID(userID)
```
alexeykazakov, line 232: Please add more tests for different use cases. Such as `-something`, `something:something`, etc.

pkg/signup/service/signup_service.go
```diff
@@ -93,6 +102,7 @@ func (s *ServiceImpl) newUserSignup(ctx *gin.Context) (*v1alpha1.UserSignup, err
 		Spec: v1alpha1.UserSignupSpec{
 			TargetCluster:        "",
 			Approved:             false,
+			UserID:               ctx.GetString(context.SubKey),
```
alexeykazakov, line 105: Let's add a test which checks that Spec.UserID is set.

alexeykazakov (commented):
Looks good overall. But please add more tests and fix the description.

sbryzak:
> So this PR should now perform a one-way encoding of the user ID into a DNS-1123 compliant value. With this change a number of further changes are required in the host operator, which are forthcoming.

After further discussion with @alexeykazakov we're going to break up these changes into multiple PRs, as the order in which we do things here are actually quite important.  This PR simply now sets the new UserSignup.Spec.UserID.

alexeykazakov (approved):

PR #167, opened 2021-01-22
Enabled user ID encoding, added tests

sbryzak:
**IMPORTANT** Only to be merged after https://github.com/codeready-toolchain/host-operator/pull/362

This PR performs a one-way encoding of the user's subject claim in order to use the encoded value as the UserSignup resource name.

### DIFF
--- a/pkg/signup/service/signup_service.go
+++ b/pkg/signup/service/signup_service.go
@@ -4,6 +4,8 @@ import (
 	"crypto/md5"
 	"encoding/hex"
 	"fmt"
+	"hash/crc32"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -86,8 +88,7 @@ func (s *ServiceImpl) newUserSignup(ctx *gin.Context) (*v1alpha1.UserSignup, err
 
 	userSignup := &v1alpha1.UserSignup{
 		ObjectMeta: v1.ObjectMeta{
-			//Name:      EncodeUserID(ctx.GetString(context.SubKey)),
-			Name:      ctx.GetString(context.SubKey),
+			Name:      EncodeUserID(ctx.GetString(context.SubKey)),
 			Namespace: s.Config.GetNamespace(),
 			Annotations: map[string]string{
 				v1alpha1.UserSignupUserEmailAnnotationKey:           userEmail,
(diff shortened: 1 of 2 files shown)


pkg/signup/service/signup_service_test.go
```diff
@@ -217,23 +220,33 @@ func (s *TestSignupServiceSuite) TestUserSignupWithInvalidSubjectPrefix() {
...
+	s.Run("test user ID with invalid prefix", func() {
+		userID := "-1234567"
+		encoded := service.EncodeUserID(userID)
+		require.Equal(s.T(), "ca3e1e0f-1234567", encoded)
+	})
+	s.Run("test user ID that exceeds max length", func() {
+		userID := "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-01234567890123456789"
+		encoded := service.EncodeUserID(userID)
+		require.Equal(s.T(), "e3632025-0123456789abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqr", encoded)
+	})
```
alexeykazakov, line 248: Let's also add a test for abc:xyz user id format. This format is used by RH sso.

alexeykazakov (commented):
Looks good but we need to add e2e tests.

pkg/signup/service/signup_service_test.go
sbryzak: Done

MatousJobanek (approved):
Nice solution :+1:

alexeykazakov (approved):

pkg/signup/service/signup_service.go
```diff
@@ -142,14 +143,14 @@ func extractEmailHost(email string) string {
...
+}
 
 // Signup reactivates the deactivated UserSignup resource or creates a new one with the specified username and userID
 // if doesn't exist yet.
 func (s *ServiceImpl) Signup(ctx *gin.Context) (*v1alpha1.UserSignup, error) {
-	userID := ctx.GetString(context.SubKey)
+	encodedUserID := EncodeUserID(ctx.GetString(context.SubKey))
 	// Retrieve UserSignup resource from the host cluster
-	userSignup, err := s.CRTClient().V1Alpha1().UserSignups().Get(userID)
+	userSignup, err := s.CRTClient().V1Alpha1().UserSignups().Get(encodedUserID)
```
alexeykazakov, line 153: It means we can't just change the encoding in the future without a migration step. Let's add a comment to the encoding function to warn about any possible change in the function that would change the user id encoding.

sbryzak: I've added a warning, I'll commit the change after e2e tests have run

PR #168, opened 2021-02-08
Added email address check also when PhoneNumberAlreadyInUse verification

sbryzak:
When doing phone verification and checking if a phone number is already in use, also check whether the email address is the same for the UserSignup which has the duplicate phone number.  If the email address is the same, then don't return an error when doing the check.

Fixes https://issues.redhat.com/browse/CRT-883

### DIFF
--- a/pkg/application/service/services.go
+++ b/pkg/application/service/services.go
@@ -11,7 +11,7 @@ type SignupService interface {
 	GetSignup(userID string) (*signup.Signup, error)
 	GetUserSignup(userID string) (*v1alpha1.UserSignup, error)
 	UpdateUserSignup(userSignup *v1alpha1.UserSignup) (*v1alpha1.UserSignup, error)
-	PhoneNumberAlreadyInUse(userID, e164PhoneNumber, email string) error
+	PhoneNumberAlreadyInUse(userID, value string) error
 }
 
 type VerificationService interface {
--- a/pkg/application/service/services.go
+++ b/pkg/application/service/services.go
@@ -11,7 +11,7 @@ type SignupService interface {
 	GetSignup(userID string) (*signup.Signup, error)
 	GetUserSignup(userID string) (*v1alpha1.UserSignup, error)
 	UpdateUserSignup(userSignup *v1alpha1.UserSignup) (*v1alpha1.UserSignup, error)
-	PhoneNumberAlreadyInUse(userID, e164PhoneNumber string) error
+	PhoneNumberAlreadyInUse(userID, e164PhoneNumber, email string) error
 }
 
 type VerificationService interface {
--- a/pkg/application/service/services.go
+++ b/pkg/application/service/services.go
@@ -11,7 +11,7 @@ type SignupService interface {
 	GetSignup(userID string) (*signup.Signup, error)
 	GetUserSignup(userID string) (*v1alpha1.UserSignup, error)
 	UpdateUserSignup(userSignup *v1alpha1.UserSignup) (*v1alpha1.UserSignup, error)
-	PhoneNumberAlreadyInUse(userID, value string) error
+	PhoneNumberAlreadyInUse(userID, phoneNumberOrHash string) error
 }
 
 type VerificationService interface {
--- a/pkg/controller/signup_test.go
+++ b/pkg/controller/signup_test.go
@@ -290,7 +290,7 @@ func (s *TestSignupSuite) TestInitVerificationHandler() {
 	ctrl := controller.NewSignup(s.Application, s.Config())
 	handler := gin.HandlerFunc(ctrl.InitVerificationHandler)
 
-	assertInitVerificationSuccess := func(phone_number string, expectedCounter int) {
+	assertInitVerificationSuccess := func(phone_number, expectedHash string, expectedCounter int) {
 		gock.New("https://api.twilio.com").
 			Reply(http.StatusNoContent).
 			BodyString("")
@@ -306,22 +306,22 @@ func (s *TestSignupSuite) TestInitVerificationHandler() {
 		require.NotEmpty(s.T(), updatedUserSignup.Annotations[crtapi.UserSignupVerificationInitTimestampAnnotationKey])
 		require.NotEmpty(s.T(), updatedUserSignup.Annotations[crtapi.UserVerificationExpiryAnnotationKey])
 		require.Equal(s.T(), strconv.Itoa(expectedCounter), updatedUserSignup.Annotations[crtapi.UserSignupVerificationCounterAnnotationKey])
-		require.Equal(s.T(), "fd276563a8232d16620da8ec85d0575f", updatedUserSignup.Labels[crtapi.UserSignupUserPhoneHashLabelKey])
+		require.Equal(s.T(), expectedHash, updatedUserSignup.Labels[crtapi.UserSignupUserPhoneHashLabelKey])
 	}
 
 	s.Run("init verification success", func() {
-		assertInitVerificationSuccess("2268213044", 1)
+		assertInitVerificationSuccess("2268213044", "fd276563a8232d16620da8ec85d0575f", 1)
 	})
 
 	s.Run("init verification success phone number with parenthesis and spaces", func() {
-		assertInitVerificationSuccess("(226) 821 3044", 2)
+		assertInitVerificationSuccess("(226) 821 3045", "9691252ac0ea2cb55295ac9b98df1c51", 2)
 	})
 
 	s.Run("init verification success phone number with dashes", func() {
-		assertInitVerificationSuccess("226-821-3044", 3)
+		assertInitVerificationSuccess("226-821-3044", "fd276563a8232d16620da8ec85d0575f", 3)
 	})
 	s.Run("init verification success phone number with spaces", func() {
-		assertInitVerificationSuccess("2 2 6 8 2 1 3 0 4 4", 4)
+		assertInitVerificationSuccess("2 2 6 8 2 1 3 0 4 7", "ce3e697125f35efb76357ed8e3b768b7", 4)
 	})
 	s.Run("init verification fails with invalid country code", func() {
 		gock.New("https://api.twilio.com").
@@ -681,7 +681,7 @@ type FakeSignupService struct {
 	MockSignup                  func(ctx *gin.Context) (*crtapi.UserSignup, error)
 	MockGetUserSignup           func(userID string) (*crtapi.UserSignup, error)
 	MockUpdateUserSignup        func(userSignup *crtapi.UserSignup) (*crtapi.UserSignup, error)
-	MockPhoneNumberAlreadyInUse func(userID, e164phoneNumber string) error
+	MockPhoneNumberAlreadyInUse func(userID, value string) error
 }
 
 func (m *FakeSignupService) GetSignup(userID string) (*signup.Signup, error) {
--- a/pkg/controller/signup_test.go
+++ b/pkg/controller/signup_test.go
@@ -290,12 +290,12 @@ func (s *TestSignupSuite) TestInitVerificationHandler() {
 	ctrl := controller.NewSignup(s.Application, s.Config())
 	handler := gin.HandlerFunc(ctrl.InitVerificationHandler)
 
-	assertInitVerificationSuccess := func(phone_number, expectedHash string, expectedCounter int) {
+	assertInitVerificationSuccess := func(phoneNumber, expectedHash string, expectedCounter int) {
 		gock.New("https://api.twilio.com").
 			Reply(http.StatusNoContent).
 			BodyString("")
 
-		data := []byte(fmt.Sprintf(`{"phone_number": "%s", "country_code": "1"}`, phone_number))
+		data := []byte(fmt.Sprintf(`{"phone_number": "%s", "country_code": "1"}`, phoneNumber))
 		rr := initVerification(s.T(), handler, gin.Param{}, data, userID, http.MethodPut, "/api/v1/signup/verification")
 		require.Equal(s.T(), http.StatusNoContent, rr.Code)
 
--- a/pkg/kubeclient/signup.go
+++ b/pkg/kubeclient/signup.go
@@ -88,10 +88,6 @@ func (c *userSignupClient) Update(obj *crtapi.UserSignup) (*crtapi.UserSignup, e
 // used to query the UserSignups, otherwise if the hash value has been provided, then that value will be used
 // directly for the query.
 func (c *userSignupClient) ListByPhoneNumberOrHash(phoneNumberOrHash string) (*crtapi.UserSignupList, error) {
-	if e164Matcher.Match([]byte(phoneNumberOrHash)) {
-		return c.listByLabelForHashedValue(crtapi.BannedUserPhoneNumberHashLabelKey, phoneNumberOrHash)
-	}
-
 	if md5Matcher.Match([]byte(phoneNumberOrHash)) {
 		return c.listByLabel(crtapi.BannedUserPhoneNumberHashLabelKey, phoneNumberOrHash)
 	}
--- a/pkg/kubeclient/signup.go
+++ b/pkg/kubeclient/signup.go
@@ -118,7 +118,7 @@ func (c *userSignupClient) listByLabel(labelKey, labelValue string) (*crtapi.Use
 
 	r := schema.GroupVersionResource{Group: "toolchain.dev.openshift.com", Version: "v1alpha1", Resource: userSignupResourcePlural}
 	listOptions := v1.ListOptions{
-		LabelSelector: fmt.Sprintf("%s=%s", labelKey, labelValue),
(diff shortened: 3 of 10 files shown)


pkg/signup/service/signup_service.go
```diff
@@ -318,7 +318,7 @@ func (s *ServiceImpl) PhoneNumberAlreadyInUse(userID, e164PhoneNumber string) er
 		return errors3.NewInternalError(err, "failed listing userSignups")
 	}
 	for _, signup := range userSignupList.Items {
-		if signup.Name != userID {
+		if signup.Spec.UserID != userID && signup.Annotations[v1alpha1.UserSignupUserEmailAnnotationKey] != email {
```
alexeykazakov, line 321: So we missed another place with signup.Name? Can we try to add a test for this case? Ideally with an e2e tests too.

alexeykazakov, line 321: Shouldn't we check if the other usersignup (different ID but same email) is deactivated?
We should allow re-using the same phone for different usersignups with the same email if there is no active usersignups with that phone and email.
But if there is already an active usersignups with the same email and phone then we should not allow to re-use the same phone again.
Otherwise it would be possible to register multiple active users with the same email and phone.
Or I'm missing something?

sbryzak: Sure, these are the rules I was unclear about.  Also I think we should check that the existing UserSignup has `VerificationRequired` set to false, to confirm that the phone number has  actually been successfully verified already.  Does this make sense?

sbryzak: Sure, we can add a test here.  It should be for a UserSignup with a non DNS-1123 compliant user ID though.

alexeykazakov: New signup == new verification.

sbryzak: It's a bit challenging to visualise all the possible combinations, so let's spell them out here - let's assume that for all conditions, the phone number is the same:

1. Different user ID, Same email, Not Deactivated - (Block)
2. Different user ID, Same email, Deactivated (Allow)
3. Different user ID, Different email, Not Deactivated (?)
4. Different user ID, Different email, Deactivated (Allow?)

alexeykazakov: 3. Block
4. I would allow if it's easy to implement (not critical though)

sbryzak: So it seems that we don't require an email comparison at all then - just block if the phone number is tied to a non-deactivated UserSignup with the same phone number, otherwise don't block?

alexeykazakov: Yes, looks like that. But we should be careful with that. It should be impossible to register multiple accounts with the same phone first (UserSignups created, verification code generated and sent) and then verify them all.

pkg/application/service/services.go
```diff
@@ -11,7 +11,7 @@ type SignupService interface {
 	GetSignup(userID string) (*signup.Signup, error)
 	GetUserSignup(userID string) (*v1alpha1.UserSignup, error)
 	UpdateUserSignup(userSignup *v1alpha1.UserSignup) (*v1alpha1.UserSignup, error)
-	PhoneNumberAlreadyInUse(userID, e164PhoneNumber, email string) error
+	PhoneNumberAlreadyInUse(userID, value string) error
```
alexeykazakov, line 14: `value` can be both a hashed phone number or a plain phone number string, right? Can we name the param accordingly to avoid confusion? Or at least add a comment?

pkg/kubeclient/signup.go
```diff
@@ -26,7 +32,7 @@ type UserSignupInterface interface {
 	Get(name string) (*crtapi.UserSignup, error)
 	Create(obj *crtapi.UserSignup) (*crtapi.UserSignup, error)
 	Update(obj *crtapi.UserSignup) (*crtapi.UserSignup, error)
-	ListByPhone(phone string) (*crtapi.UserSignupList, error)
+	ListByPhone(value string) (*crtapi.UserSignupList, error)
```
alexeykazakov, line 35: Same here.

pkg/kubeclient/signup.go
```diff
@@ -77,27 +83,42 @@ func (c *userSignupClient) Update(obj *crtapi.UserSignup) (*crtapi.UserSignup, e
 	return result, nil
 }
 
-func (c *userSignupClient) ListByPhone(phone string) (*crtapi.UserSignupList, error) {
-	return c.listByHashedLabel(crtapi.BannedUserPhoneNumberHashLabelKey, phone)
-}
+func (c *userSignupClient) ListByPhone(value string) (*crtapi.UserSignupList, error) {
```
alexeykazakov, line 86: And here.

pkg/signup/service/signup_service.go
```diff
@@ -304,22 +304,23 @@ func (s *ServiceImpl) UpdateUserSignup(userSignup *v1alpha1.UserSignup) (*v1alph
 // PhoneNumberAlreadyInUse checks if the phone number has been banned. If so, return
 // an internal server error. If not, check if a signup with a different userID
 // and email address exists. If so, return an internal server error. Otherwise, return without error.
-func (s *ServiceImpl) PhoneNumberAlreadyInUse(userID, e164PhoneNumber, email string) error {
-	bannedUserList, err := s.CRTClient().V1Alpha1().BannedUsers().ListByPhone(e164PhoneNumber)
+func (s *ServiceImpl) PhoneNumberAlreadyInUse(userID, value string) error {
```
alexeykazakov, line 307: And here.

pkg/kubeclient/signup.go
```diff
@@ -77,27 +83,42 @@ func (c *userSignupClient) Update(obj *crtapi.UserSignup) (*crtapi.UserSignup, e
...
-// ListByHashedLabel returns a UserSignupList containing any UserSignup resources that have a label matching the specified label
-func (c *userSignupClient) listByHashedLabel(labelKey, labelValue string) (*crtapi.UserSignupList, error) {
+	// Default to searching for a hash of the specified value
+	return c.listByHashedLabelValue(crtapi.BannedUserPhoneNumberHashLabelKey, value)
+}
 
-	// Calculate the md5 hash for the phoneNumber
+// listByHashedLabelValue returns a UserSignupList containing any UserSignup resources that have a label matching the
+// md5 hash of the specified value
+func (c *userSignupClient) listByHashedLabelValue(labelKey, value string) (*crtapi.UserSignupList, error) {
```
alexeykazakov, line 101: But `value` can be already hashed, right?

pkg/application/service/services.go
sbryzak: Correct.  Do you have any suggestions for a better param name? I can certainly add a comment.

pkg/kubeclient/signup.go
sbryzak: This function calculates the hash of the provided value and searches for that.  I'm not a fan of the function name but couldn't think of a better one in the moment.

alexeykazakov: OK. I re-read the code and comments and I think it's clear enough. Let's keep it as is.

pkg/application/service/services.go
alexeykazakov: `phoneNumberOrHash string`? And some comments in the implementation function would be helpful too.

sbryzak: I've improved the naming and added more descriptive comments.

pkg/kubeclient/signup.go
```diff
@@ -77,27 +83,46 @@ func (c *userSignupClient) Update(obj *crtapi.UserSignup) (*crtapi.UserSignup, e
...
+// used to query the UserSignups, otherwise if the hash value has been provided, then that value will be used
+// directly for the query.
+func (c *userSignupClient) ListByPhoneNumberOrHash(phoneNumberOrHash string) (*crtapi.UserSignupList, error) {
+	if e164Matcher.Match([]byte(phoneNumberOrHash)) {
+		return c.listByLabelForHashedValue(crtapi.BannedUserPhoneNumberHashLabelKey, phoneNumberOrHash)
+	}
+
+	if md5Matcher.Match([]byte(phoneNumberOrHash)) {
+		return c.listByLabel(crtapi.BannedUserPhoneNumberHashLabelKey, phoneNumberOrHash)
+	}
```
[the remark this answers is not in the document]
MatousJobanek: btw, you should use the `UserSignupUserPhoneHashLabelKey` instead of `BannedUserPhoneNumberHashLabelKey`

pkg/kubeclient/signup.go
```diff
@@ -77,27 +83,46 @@ func (c *userSignupClient) Update(obj *crtapi.UserSignup) (*crtapi.UserSignup, e
...
+
 	intf, err := dynamic.NewForConfig(&c.cfg)
 	if err != nil {
 		return nil, err
 	}
 
 	r := schema.GroupVersionResource{Group: "toolchain.dev.openshift.com", Version: "v1alpha1", Resource: userSignupResourcePlural}
 	listOptions := v1.ListOptions{
-		LabelSelector: fmt.Sprintf("%s=%s", labelKey, hash),
+		LabelSelector: fmt.Sprintf("%s=%s", labelKey, labelValue),
```
MatousJobanek, line 125: you can filter the deactivated UserSignups here
```suggestion
		LabelSelector: fmt.Sprintf("%s!=%s,%s=%s", crtapi.UserSignupStateLabelKey, crtapi.UserSignupStateLabelValueDeactivated,labelKey, labelValue),
```

pkg/signup/service/signup_service.go
```diff
@@ -313,23 +313,26 @@ func (s *ServiceImpl) UpdateUserSignup(userSignup *v1alpha1.UserSignup) (*v1alph
...
-	userSignupList, err := s.CRTClient().V1Alpha1().UserSignups().ListByPhone(e164PhoneNumber)
+	userSignupList, err := s.CRTClient().V1Alpha1().UserSignups().ListByPhoneNumberOrHash(phoneNumberOrHash)
 	if err != nil {
 		return errs.NewInternalError(err, "failed listing userSignups")
 	}
 	for _, signup := range userSignupList.Items {
-		if signup.Name != userID {
-			return errs.NewForbiddenError("cannot re-register with phone number", "phone number already in use")
+
+		if signup.Spec.UserID != userID && !signup.Spec.Deactivated {
```
MatousJobanek, line 333: If you use my proposed way of using the label selector, then you can drop the check from here:
```suggestion
		if signup.Spec.UserID != userID {
```

pkg/controller/signup_test.go
```diff
@@ -290,7 +290,7 @@ func (s *TestSignupSuite) TestInitVerificationHandler() {
 	ctrl := controller.NewSignup(s.Application, s.Config())
 	handler := gin.HandlerFunc(ctrl.InitVerificationHandler)
 
-	assertInitVerificationSuccess := func(phone_number string, expectedCounter int) {
+	assertInitVerificationSuccess := func(phone_number, expectedHash string, expectedCounter int) {
```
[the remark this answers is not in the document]
sbryzak: Sure thing, changed.

pkg/kubeclient/signup.go
```diff
@@ -77,27 +83,42 @@ func (c *userSignupClient) Update(obj *crtapi.UserSignup) (*crtapi.UserSignup, e
...
 	// Ignore the error, as this implementation cannot return one
-	_, _ = md5hash.Write([]byte(labelValue))
+	_, _ = md5hash.Write([]byte(value))
 	hash := hex.EncodeToString(md5hash.Sum(nil))
 
+	return c.listByLabel(labelKey, hash)
+}
+
+// listByLabel returns a UserSignupList containing any UserSignup resources that have a label matching the specified label
+func (c *userSignupClient) listByLabel(labelKey, labelValue string) (*crtapi.UserSignupList, error) {
```
alexeykazakov, line 112: Shouldn't it be something like `listNotDeactivatedSignupsByLabel()` because it look up the signups by the label only among signups with the state != `deactivated`?

pkg/kubeclient/signup.go
```diff
@@ -77,27 +83,42 @@ func (c *userSignupClient) Update(obj *crtapi.UserSignup) (*crtapi.UserSignup, e
...
-// ListByHashedLabel returns a UserSignupList containing any UserSignup resources that have a label matching the specified label
-func (c *userSignupClient) listByHashedLabel(labelKey, labelValue string) (*crtapi.UserSignupList, error) {
+	// Default to searching for a hash of the specified value
+	return c.listByLabelForHashedValue(crtapi.BannedUserPhoneNumberHashLabelKey, phoneNumberOrHash)
+}
 
-	// Calculate the md5 hash for the phoneNumber
+// listByLabelForHashedValue returns a UserSignupList containing any UserSignup resources that have a label matching the
+// md5 hash of the specified value
+func (c *userSignupClient) listByLabelForHashedValue(labelKey, value string) (*crtapi.UserSignupList, error) {
```
alexeykazakov, line 101: Same here. Something like `listNotDeactivatedSignupsByLabelForHashedValue()`

pkg/kubeclient/signup.go
```diff
@@ -77,27 +83,42 @@ func (c *userSignupClient) Update(obj *crtapi.UserSignup) (*crtapi.UserSignup, e
...
 }
 
-func (c *userSignupClient) ListByPhone(phone string) (*crtapi.UserSignupList, error) {
-	return c.listByHashedLabel(crtapi.BannedUserPhoneNumberHashLabelKey, phone)
-}
+// ListByPhoneNumberOrHash will return a list of UserSignups that have a phone number hash label value matching
+// the provided value.  If the value provided is an actual phone number, then the hash will be calculated and then
+// used to query the UserSignups, otherwise if the hash value has been provided, then that value will be used
+// directly for the query.
+func (c *userSignupClient) ListByPhoneNumberOrHash(phoneNumberOrHash string) (*crtapi.UserSignupList, error) {
```
alexeykazakov, line 90: And here (not deactivated only).
And we will need to mention it in the comment.

pkg/signup/service/signup_service.go
```diff
@@ -315,23 +315,26 @@ func (s *ServiceImpl) UpdateUserSignup(userSignup *v1alpha1.UserSignup) (*v1alph
 
 // PhoneNumberAlreadyInUse checks if the phone number has been banned. If so, return
 // an internal server error. If not, check if a signup with a different userID
-// exists. If so, return an internal server error. Otherwise, return without error.
-func (s *ServiceImpl) PhoneNumberAlreadyInUse(userID, e164PhoneNumber string) error {
-	bannedUserList, err := s.CRTClient().V1Alpha1().BannedUsers().ListByPhone(e164PhoneNumber)
+// and email address exists. If so, return an internal server error. Otherwise, return without error.
+// Either the actual phone number, or the md5 hash of the phone number may be provided here.
```
alexeykazakov, line 319: Please mention in the description that deactivated signups are ignored.

pkg/verification/service/verification_service_test.go
```diff
@@ -320,6 +322,71 @@ func (s *TestVerificationServiceSuite) TestInitVerificationFailsDailyCounterExce
 	require.Empty(s.T(), userSignup.Annotations[v1alpha1.UserSignupVerificationCodeAnnotationKey])
 }
 
+func (s *TestVerificationServiceSuite) TestInitVerificationFailsWhenPhoneNumberInUse() {
```
alexeykazakov, line 325: Let's also add a test for the case when there is a deactivated usersignup with the same phone.

alexeykazakov (commented):
Code looks good!
But let's add one more test and improve the function names and descriptions.

pkg/kubeclient/signup.go
sbryzak: I've renamed it to `listActiveSignupsByLabel()`

pkg/kubeclient/signup.go
sbryzak: Renamed to `listActiveSignupsByLabelForHashedValue()`

pkg/kubeclient/signup.go
sbryzak: Renamed to `ListActiveSignupsByPhoneNumberOrHash()`

pkg/signup/service/signup_service.go
sbryzak: Done

pkg/verification/service/verification_service_test.go
sbryzak: Done

sbryzak:
@alexeykazakov comments have been addressed now - one thing to note is this change here:

https://github.com/codeready-toolchain/registration-service/pull/168/files#diff-044c9e6475638d41c551f545f3899df6a42b37489faa3c9f7eb87edfa87b79b0R335

I've also added a check to confirm that `UserSignup.Spec.Deactivated` is false, in the unlikely case that we get a "false positive" returned from the client, i.e. a `UserSignup` with the state label set to deactivated, but not actually deactivated in the spec.

alexeykazakov (approved):

PR #170, opened 2021-02-11
preventing usersignups for usernames that have the crtadmin suffix

tinakurian:
Preventing the registration-service from creating signup resources for users who signup with the crtadmin suffix in their username.
 
Jira: https://issues.redhat.com/browse/CRT-913

Related PRs: https://github.com/codeready-toolchain/toolchain-common/pull/151, https://github.com/codeready-toolchain/toolchain-e2e/pull/237, https://github.com/codeready-toolchain/host-operator/pull/373
**E2E-Tests**: https://github.com/codeready-toolchain/toolchain-e2e/pull/237

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.14
 
 require (
 	github.com/cheekybits/is v0.0.0-20150225183255-68e9c0620927 // indirect
-	github.com/codeready-toolchain/api v0.0.0-20210115184740-77f08aa35c87
+	github.com/codeready-toolchain/api v0.0.0-20210203025716-45f19e752101
 	github.com/codeready-toolchain/toolchain-common v0.0.0-20210209125948-7a9e8c9e4ee6
 	github.com/dgrijalva/jwt-go v3.2.0+incompatible
 	github.com/gin-contrib/cors v1.3.1
@@ -40,8 +40,8 @@ require (
 
 replace (
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.2+incompatible // Required by OLM
+	github.com/codeready-toolchain/toolchain-common v0.0.0-20210209125948-7a9e8c9e4ee6 => github.com/tinakurian/toolchain-common v0.0.0-20210216182308-9a7445b1611e
 	github.com/openshift/api => github.com/openshift/api v0.0.0-20200821140346-b94c46af3f2b // Using 'github.com/openshift/api@release-4.5'
 	k8s.io/client-go => k8s.io/client-go v0.18.3 // Required by prometheus-operator
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20200410145947-61e04a5be9a6 // avoids case-insensitive import collision: "github.com/googleapis/gnostic/openapiv2" and "github.com/googleapis/gnostic/OpenAPIv2"
 )
-
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.14
 require (
 	github.com/cheekybits/is v0.0.0-20150225183255-68e9c0620927 // indirect
 	github.com/codeready-toolchain/api v0.0.0-20210203025716-45f19e752101
-	github.com/codeready-toolchain/toolchain-common v0.0.0-20210209125948-7a9e8c9e4ee6
+	github.com/codeready-toolchain/toolchain-common v0.0.0-20210219141158-8bf475892a90
 	github.com/dgrijalva/jwt-go v3.2.0+incompatible
 	github.com/gin-contrib/cors v1.3.1
 	github.com/gin-contrib/gzip v0.0.1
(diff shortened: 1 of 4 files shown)


alexeykazakov (approved):
Looks good.

pkg/signup/service/signup_service.go
```diff
@@ -56,6 +56,11 @@ func NewSignupService(context servicecontext.ServiceContext, cfg ServiceConfigur
 // newUserSignup generates a new UserSignup resource with the specified username and userID.
 // This resource then can be used to create a new UserSignup in the host cluster or to update the existing one.
 func (s *ServiceImpl) newUserSignup(ctx *gin.Context) (*v1alpha1.UserSignup, error) {
+	username := ctx.GetString(context.UsernameKey)
+	if strings.HasSuffix(username, "crtadmin") {
```
xcoulon, line 60: It's a minor thing, but I think that `-crtadmin` would be a more accurate suffix. 
Still, your code here should do the job and prevent us from signing up with our admin accounts.

xcoulon (approved):

pkg/signup/service/signup_service.go
MatousJobanek: I would rather use the `crtadmin` and check it as a substring - not just as a suffix. If I'm not mistaken, you could create `mjobanek-crtadmin#` that would be then transformed by UserSignup controller into `mjobanek-crtadmin`. In other words, it's still a security issue.

MatousJobanek (changes requested):

xcoulon:
oh, I did not realize that a username like `mjobanek-crtadmin#` would be transformed into `mjobanek-crtadmin` and thus would match an existing CRT admin account.
My apologies @tinakurian, my suggestion to use the `strings.Suffix(...)` func wasn't accurate 😬

MatousJobanek:
My bad, I thought that we remove the invalid chars, but we replace them with `-`, so `mjobanek-crtadmin#` wouldn't cause any issue. However `mjobanek#crtadmin` would...
Anyway, I would still prefer checking the `crtadmin` as a substring.
Or, you could move the transformation logic of the names into toolchain-common: https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/usersignup/usersignup_controller.go#L457-L478 and reuse it here. In that way, you would be able to execute the transformation and be sure that you won't allow the creation of any `crtadmin` users.

tinakurian:
> My bad, I thought that we remove the invalid chars, but we replace them with `-`, so `mjobanek-crtadmin#` wouldn't cause any issue. However `mjobanek#crtadmin` would...
> Anyway, I would still prefer checking the `crtadmin` as a substring.
> Or, you could move the transformation logic of the names into toolchain-common: https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/usersignup/usersignup_controller.go#L457-L478 and reuse it here. In that way, you would be able to execute the transformation and be sure that you won't allow the creation of any `crtadmin` users.

@MatousJobanek Good point! I've updated the code to use contains instead here: 4d14b0e99afeec56880dc86b757fa80c1803ef8a

pkg/signup/service/signup_service.go
```diff
@@ -56,6 +57,13 @@ func NewSignupService(context servicecontext.ServiceContext, cfg ServiceConfigur
 // newUserSignup generates a new UserSignup resource with the specified username and userID.
 // This resource then can be used to create a new UserSignup in the host cluster or to update the existing one.
 func (s *ServiceImpl) newUserSignup(ctx *gin.Context) (*v1alpha1.UserSignup, error) {
+	username := ctx.GetString(context.UsernameKey)
+
+	usersignup.TransformUsername(username)
```
MatousJobanek, line 62: you need to use the output of the transformation: 
```suggestion
	username = usersignup.TransformUsername(username)
```

pkg/signup/service/signup_service.go
```diff
@@ -56,6 +57,13 @@ func NewSignupService(context servicecontext.ServiceContext, cfg ServiceConfigur
...
+	if strings.HasSuffix(username, "crtadmin") {
+		return nil, errors3.NewForbiddenError(fmt.Sprintf("failed to create usersignup for %s", username), "cannot create usersignup for crtadmin")
```
MatousJobanek, line 64: could you plaese fix the name of the imported package `errors3` to something more meaningful?

go.mod
```diff
@@ -40,8 +40,8 @@ require (
 
 replace (
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.2+incompatible // Required by OLM
+	github.com/codeready-toolchain/toolchain-common v0.0.0-20210209125948-7a9e8c9e4ee6 => github.com/tinakurian/toolchain-common v0.0.0-20210217164217-8325ca100257
```
MatousJobanek, line 43: a reminder

MatousJobanek (commented):
Looks good :+1:

pkg/signup/service/signup_service.go
tinakurian: woopsie, looks like i got too excited. fixed here: b49f7bb7673a86099e9ab0fce3101f46f391503f

pkg/signup/service/signup_service.go
tinakurian: @MatousJobanek sure, in another place we use errs, so i updated it to be errs: 6504142457aa5ae8fe870bdc12895a95b6cb0732

pkg/signup/service/signup_service.go
xcoulon: After our discussion on Slack, I believe that a username such as `mjobanek-crtadmin#` will not match any admin account since it will be transformed into `mjobanek-crtadmin-crt`, so I would stand with the `-crtadmin` suffix here. 
@MatousJobanek WDYT?

MatousJobanek (approved):
Nice :+1:

PR #171, opened 2021-02-22
Fix 500 catch all for PostHandler

tinakurian:
Currently, the registration-service signup-service returns other errors such as 403. However, the PostHandler does a catch-all and returns the http code 500 for all errors.  This should be changed to propagate and return the error which the service actually returned.

Related e2e-test PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/241
Jira: https://issues.redhat.com/browse/CRT-966

### DIFF
--- a/pkg/controller/signup.go
+++ b/pkg/controller/signup.go
@@ -40,8 +40,8 @@ func (s *Signup) PostHandler(ctx *gin.Context) {
 	if err, ok := err.(*errors2.StatusError); ok {
 		errors.AbortWithError(ctx, int(err.Status().Code), err, err.Status().Message)
 	}
-	if err != nil {
 
+	if err != nil {
 		log.Error(ctx, err, "error creating UserSignup resource")
 		errors.AbortWithError(ctx, http.StatusInternalServerError, err, "error creating UserSignup resource")
 		return
--- a/pkg/controller/signup.go
+++ b/pkg/controller/signup.go
@@ -39,6 +39,7 @@ func (s *Signup) PostHandler(ctx *gin.Context) {
 	userSignup, err := s.app.SignupService().Signup(ctx)
 	if err, ok := err.(*errors2.StatusError); ok {
 		errors.AbortWithError(ctx, int(err.Status().Code), err, err.Status().Message)
+		return
 	}
 
 	if err != nil {
--- a/pkg/controller/signup.go
+++ b/pkg/controller/signup.go
@@ -10,9 +10,9 @@ import (
 	"github.com/codeready-toolchain/registration-service/pkg/errors"
 	"github.com/codeready-toolchain/registration-service/pkg/log"
 
-	errors2 "k8s.io/apimachinery/pkg/api/errors"
 	"github.com/gin-gonic/gin"
 	"github.com/nyaruka/phonenumbers"
+	errors2 "k8s.io/apimachinery/pkg/api/errors"
 )
 
 // Signup implements the signup endpoint, which is invoked for new user registrations.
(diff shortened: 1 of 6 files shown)


pkg/signup/service/signup_service.go
```diff
@@ -178,7 +180,7 @@ func (s *ServiceImpl) Signup(ctx *gin.Context) (*v1alpha1.UserSignup, error) {
 	}
 
 	username := ctx.GetString(context.UsernameKey)
-	return nil, errors2.Errorf("unable to create UserSignup [id: %s; username: %s] because there is already an active UserSignup with such ID", encodedUserID, username)
+	return nil, errors.NewBadRequest(fmt.Sprintf("unable to create UserSignup [id: %s; username: %s] because there is already an active UserSignup with such ID", encodedUserID, username))
```
alexeykazakov, line 183: Shouldn't it be 409 Conflict instead?

pkg/controller/signup.go
```diff
@@ -36,6 +37,11 @@ func NewSignup(app application.Application, config configuration.Configuration)
 // PostHandler creates a Signup resource
 func (s *Signup) PostHandler(ctx *gin.Context) {
 	userSignup, err := s.app.SignupService().Signup(ctx)
+	if err, ok := err.(*errors2.StatusError); ok {
+		errors.AbortWithError(ctx, int(err.Status().Code), err, "error creating UserSignup resource")
```
alexeykazakov, line 41: Could we maybe create a new function similar to `errors.AbortWithError()` which would accept an error object and either use the status code if it's StatusError (as you do here) or use 500 by default.
Then we could either explicitly set the expected code when we know what we want (eg. `errors.AbortWithError(ctx, code, err, message)`) or do something like `errors.AbortWithStatusError(ctx, err, message)` when we want to use the code from the corresponding StatusError or use 500 by default.

alexeykazakov (commented):
Overall looks good but I have a comment/question.
Also, do we have any tests for the user signup controller which checks the error? It's covered by e2e tests but would be useful to have a unit test for that too if it's not too tricky.

tinakurian:
> Overall looks good but I have a comment/question.
> Also, do we have any tests for the user signup controller which checks the error? It's covered by e2e tests but would be useful to have a unit test for that too if it's not too tricky.

@alexeykazakov we do have a test for the use-case where a generic err is returned. I've added one specifically for the forbidden use-case here: 068a5357fee27e84cfea63d0c81e3b2b3b04dcd3

pkg/signup/service/signup_service.go
tinakurian: changed to conflict here: 66887ddb508818319fd84bfd085321ff902d31bc

pkg/controller/signup.go
tinakurian: @alexeykazakov Yup, I think that's possible and sounds like a good idea. We can maybe add this to the reg-service errors package. There is already an AbortWithError there that we can re-purpose to do as you described above

alexeykazakov: I'm not sure if we can re-purpose `AbortWithError()` because we need to specify the error code explicitly in some cases. It seems that we need both functions. So, I would keep `AbortWithError()` as is but also introduce a new function where we delegate the error code setting. Something like `AbortWithStatusError()`. See an example in my previous comment.

alexeykazakov (approved):
The change is good enough for this particular issue. Feel free to address the refactoring on other errors and introducing a new AbortWithStatusError() function in a separate PR.

xcoulon (approved):

PR #173, opened 2021-02-25
Use k8s StatusError

tinakurian:
The registration-service currently returns either reg-service errors or k8s errors. We should return only k8s errors instead of having a mix. 

Related e2e-tests PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/248
Jira: https://issues.redhat.com/browse/CRT-972

### DIFF
--- a/pkg/controller/signup.go
+++ b/pkg/controller/signup.go
@@ -1,6 +1,7 @@
 package controller
 
 import (
+	"fmt"
 	"net/http"
 	"strconv"
 
@@ -85,14 +86,14 @@ func (s *Signup) InitVerificationHandler(ctx *gin.Context) {
 
 	e164Number := phonenumbers.Format(number, phonenumbers.E164)
 	err = s.app.VerificationService().InitVerification(ctx, userID, e164Number)
+	if err, ok := err.(*errors2.StatusError); ok {
+		errors.AbortWithError(ctx, int(err.Status().Code), err, fmt.Sprintf("Verification for %s could not be sent", userID))
+		return
+	}
+
 	if err != nil {
-		log.Errorf(ctx, nil, "Verification for %s could not be sent", userID)
-		switch t := err.(type) {
-		default:
-			errors.AbortWithError(ctx, http.StatusInternalServerError, err, "error while initiating verification")
-		case *errors.Error:
-			errors.AbortWithError(ctx, int(t.Code), err, t.Message)
-		}
+		log.Error(ctx, err, "error creating UserSignup resource")
+		errors.AbortWithError(ctx, http.StatusInternalServerError, err, fmt.Sprintf("Verification for %s could not be sent", userID))
 		return
 	}
 
@@ -129,17 +130,18 @@ func (s *Signup) VerifyCodeHandler(ctx *gin.Context) {
 	}
 
 	userID := ctx.GetString(context.SubKey)
-
 	err := s.app.VerificationService().VerifyCode(ctx, userID, code)
-	if err != nil {
+	if err, ok := err.(*errors2.StatusError); ok {
 		log.Error(ctx, err, "error validating user verification code")
-		switch t := err.(type) {
-		default:
-			errors.AbortWithError(ctx, http.StatusInternalServerError, err, "unexpected error while verifying code")
-		case *errors.Error:
-			errors.AbortWithError(ctx, int(t.Code), err, "error while verifying code")
-		}
+		errors.AbortWithError(ctx, int(err.Status().Code), err, "error while verifying code")
 		return
 	}
+
+	if err != nil {
+		log.Error(ctx, err, "error validating user verification code")
+		errors.AbortWithError(ctx, http.StatusInternalServerError, err, "unexpected error while verifying code")
+		return
+	}
+
 	ctx.Status(http.StatusOK)
 }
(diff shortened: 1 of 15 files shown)


pkg/errors/error.go
```diff
@@ -24,51 +25,21 @@ func AbortWithError(ctx *gin.Context, code int, err error, details string) {
...
-
-func NewNotFoundError(err error, details string) *Error {
-	return &Error{
-		Status:  http.StatusText(http.StatusNotFound),
-		Code:    http.StatusNotFound,
+	ctx.AbortWithStatusJSON(code, &Error{
+		Status:  http.StatusText(code),
+		Code:    code,
 		Message: err.Error(),
 		Details: details,
```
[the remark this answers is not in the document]
alexeykazakov: See https://projects.engineering.redhat.com/browse/RHDX-1403 for example

pkg/controller/signup.go
```diff
@@ -86,13 +82,8 @@ func (s *Signup) InitVerificationHandler(ctx *gin.Context) {
...
 	if err != nil {
-		log.Errorf(ctx, nil, "Verification for %s could not be sent", userID)
-		switch t := err.(type) {
-		case *errors.Error:
-			errors.AbortWithError(ctx, int(t.Code), err, t.Message)
-		default:
-			errors.AbortWithError(ctx, http.StatusInternalServerError, err, "error while initiating verification")
-		}
+		log.Error(ctx, err, "error creating UserSignup resource")
+		errors.AbortWithStatusError(ctx, err, fmt.Sprintf("Verification for %s could not be sent", userID))
```
alexeykazakov, line 86:
```suggestion
		errors.AbortWithStatusError(ctx, err, "error while initiating verification")
```

pkg/controller/signup.go
```diff
@@ -109,7 +100,7 @@ func (s *Signup) GetHandler(ctx *gin.Context) {
 	signupResource, err := s.app.SignupService().GetSignup(userID)
 	if err != nil {
 		log.Error(ctx, err, "error getting UserSignup resource")
-		errors.AbortWithError(ctx, http.StatusInternalServerError, err, "error getting UserSignup resource")
+		errors.AbortWithStatusError(ctx, err, "error getting UserSignup resource")
```
alexeykazakov, line 103: I guess we always want to return 500 if something goes wrong here.

pkg/verification/service/verification_service.go
```diff
@@ -69,33 +67,29 @@ func NewVerificationService(context servicecontext.ServiceContext, cfg ServiceCo
...
-				return errors.NewForbiddenError("phone number already in use", fmt.Sprintf("cannot register using phone number: %s", e164PhoneNumber))
-			}
-		default:
-			log.Error(ctx, err, "error while looking up users by phone number")
-			return errors.NewInternalError(err, "could not lookup users by phone number")
+		if errors.IsForbidden(err) {
+			log.Errorf(ctx, err, "phone number already in use, cannot register using phone number: %s", e164PhoneNumber)
+			return errors.NewForbidden(schema.GroupResource{}, "", fmt.Errorf("phone number already in use: cannot register using phone number: %s", e164PhoneNumber))
 		}
+		log.Error(ctx, err, "error while looking up users by phone number")
```
alexeykazakov, line 91: I think we still should return 500 if the error is not forbidden.

pkg/errors/error.go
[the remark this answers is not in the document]
tinakurian: @alexeykazakov Yeah.. i noticed that when i was reading the messages that were getting printed. I can reach out to the RHD folks and see if it would be a problem.

pkg/controller/signup.go
tinakurian: Yeah, right now only 500 will be returned. I just thought in case in the future we add some other error message this is already setup to handle it. WDYT?

pkg/controller/signup.go
```diff
@@ -86,13 +82,8 @@ func (s *Signup) InitVerificationHandler(ctx *gin.Context) {
...
 	err = s.app.VerificationService().InitVerification(ctx, userID, e164Number)
 	if err != nil {
-		log.Errorf(ctx, nil, "Verification for %s could not be sent", userID)
-		switch t := err.(type) {
-		case *errors.Error:
-			errors.AbortWithError(ctx, int(t.Code), err, t.Message)
-		default:
-			errors.AbortWithError(ctx, http.StatusInternalServerError, err, "error while initiating verification")
-		}
+		log.Error(ctx, err, "error creating UserSignup resource")
```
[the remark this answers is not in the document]
tinakurian: fixed here: f34b13223c2f5d25a00cc60e9188bd437deeafed

pkg/controller/signup.go
tinakurian: fixed here: f34b13223c2f5d25a00cc60e9188bd437deeafed

pkg/verification/service/verification_service.go
tinakurian: @alexeykazakov agreed, it will be returned as 500 here: https://github.com/codeready-toolchain/registration-service/blob/54c32bb024040def5b051da4d33998e8bf07c015/pkg/errors/error.go#L30

PR #184, opened 2021-05-05
Refactored deprecated VerificationRequired and Deactivation

sbryzak:
This PR refactors the registration service to use the UserSignup `states` property instead of `Spec.VerificationRequired` and `Spec.Deactivation`. 

Fixes: https://issues.redhat.com/browse/CRT-1005

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/290


sbryzak:
@alexeykazakov I've updated registration service to handle both pre and post-migrated UserSignups in regard to their VerificationRequired status.  With this change we should be able to safely  deploy registration first, followed by host operator which will complete the migration.

pkg/signup/service/signup_service.go
```diff
@@ -111,16 +111,16 @@ func (s *ServiceImpl) newUserSignup(ctx *gin.Context) (*v1alpha1.UserSignup, err
...
 		},
 		Spec: v1alpha1.UserSignupSpec{
-			TargetCluster:        "",
-			Approved:             false,
-			UserID:               ctx.GetString(context.SubKey),
-			Username:             ctx.GetString(context.UsernameKey),
-			GivenName:            ctx.GetString(context.GivenNameKey),
-			FamilyName:           ctx.GetString(context.FamilyNameKey),
-			Company:              ctx.GetString(context.CompanyKey),
-			VerificationRequired: verificationRequired,
```
alexeykazakov, line 121: But the host operator still expecting this field to be set: https://github.com/codeready-toolchain/host-operator/blob/7a8b0badfd2329bee20c1d99da4e7fc1d0bbd681/pkg/controller/usersignup/usersignup_controller.go#L358

Shouldn't we instead do something like this:
- Modify Reg-service and set and use both "states" and Spec.VerificationRequired. Deploy it. Now we set the new states thing for new UserSignups but we still support existing resource and the code in host operator still works.
- Modify host operator to 1) migrate existing UserSignups (set states if not set but VerificationRequired is set) **and** 2) start relying on "states" only. Deploy it.
- Update sandbox-sre so it uses "states" only from now on.

Now we can start cleaning up:

- Drop any usage of VerificationRequired in Reg-service. Deploy it.
- Drop any usage of VerificatoinRequired in Host. Remove the migration code. Deploy it.
- Remove VerficationRequired from API. Re-generate all the affected repos. **Update api dependencies in all repos**.

pkg/signup/service/signup_service.go
```diff
@@ -279,7 +280,7 @@ func (s *ServiceImpl) GetSignup(userID string) (*signup.Signup, error) {
 		Ready:                ready,
 		Reason:               murCondition.Reason,
 		Message:              murCondition.Message,
-		VerificationRequired: userSignup.Spec.VerificationRequired,
+		VerificationRequired: states.VerificationRequired(userSignup),
```
alexeykazakov, line 283: Should be 
```suggestion
		VerificationRequired: states.VerificationRequired(userSignup) || userSignup.Spec.VerificationRequired,
```
right?

pkg/signup/service/signup_service.go
```diff
@@ -339,7 +341,7 @@ func (s *ServiceImpl) PhoneNumberAlreadyInUse(userID, phoneNumberOrHash string)
 	}
 	for _, signup := range userSignupList.Items {
 
-		if signup.Spec.UserID != userID && !signup.Spec.Deactivated {
+		if signup.Spec.UserID != userID && !states.Deactivated(&signup) {
```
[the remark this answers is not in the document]
sbryzak: Correct, this one was missed.

alexeykazakov (approved):
Looks good!

PR #189, opened 2021-05-14
chore: URL fix for GH actions badge

rsvoboda:
URL fix for GH actions badge


MatousJobanek (approved):
Good catch, thanks a lot for fixing it @rsvoboda :+1:

alexeykazakov (approved):

PR #192, opened 2021-05-28
Refactor usage of UserSignup.Spec.Approved

sbryzak:
This PR refactors the usage of the `UserSignup.Spec.Approved` property to use states instead.

Fixes https://issues.redhat.com/browse/CRT-1114


alexeykazakov (approved):
Shouldn't we update e2e tests too?

PR #195, opened 2021-06-17
feat: migrate reg-service to operator-sdk 1.8.0

MatousJobanek:
## Migrates registration-service repo to operator-sdk 1.8.0

* cleans README from obsolete parts
* updates dependencies to use k8s v0.20.2
* updated go to 1.16
    * in go.mod
    * in GH Actions workflows  
* as for make targets
    * drops `dev.mk` as it's not needed anymore (or at least for now)
    * significant changes in `manifests.mk` file to match the latest changes in CI/CD setup/logic
    * also updates `test.mk` so it takes into account also the fact that it is executed in GH Action workflow
* as for Dockerfile.tools
    * updated go to 1.16
    * updated oc to the latest 4.7
    * dropped operator-sdk as is not needed in openshift-ci anymore


alexeykazakov (approved):

make/generate.mk
```diff
@@ -1,6 +1,7 @@
 .PHONY: generate
 ## generates the asset bundle to be packaged with the binary
 generate:
+	go get github.com/shurcooL/vfsgen
```
xcoulon, line 4: `github.com/shurcooL/vfsgen` is already declared in `go.mod` and in `pkg/static/assets_generate.go`, so do we need it here as well?

MatousJobanek: no idea why, but the `go mod tidy` in go 1.16 always removes it.

xcoulon: ah I see: maybe it's because `pkg/static/assets_generate.go` is not needed to compile the binary? In that case, we could import this package with a blank identifier in `cmd/main.go`:
```
import (
  ...
  _ "github.com/shurcooL/vfsgen"
)
```

MatousJobanek: I have no clue how this works, but I can give it a try. 
And I'm also curious why do we use this and not the go-bindata as in other repositories. @sbryzak ?

MatousJobanek: yeah, it worked 43e46fa
Thanks a lot for the suggestion @xcoulon :pray:

xcoulon: or since we're moving to go 1.16, we could also take a look at the `//go:embed` directive: https://golang.org/pkg/embed/

xcoulon: but I'm not sure if the `//go:embed` directive will give us all the flexibility we have with go-bindata with regards to switching assets in tests. That's something to look at.

xcoulon (approved):

PR #196, opened 2021-06-21
feat: add publish-operators-for-e2e-tests workflow

MatousJobanek:
similar as is already in host & member operator repos

### DIFF
--- a/.github/workflows/publish-operators-for-e2e-tests.yml
+++ b/.github/workflows/publish-operators-for-e2e-tests.yml
@@ -42,7 +42,7 @@ jobs:
       uses: MatousJobanek/toolchain-cicd/prepare-tools-action@master
 
     - name: Publish current operator bundles for host & member
-      #uses: codeready-toolchain/toolchain-cicd/publish-operator-for-e2e-tests@master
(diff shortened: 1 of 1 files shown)


alexeykazakov (approved):

.github/workflows/publish-operators-for-e2e-tests.yml
```diff
@@ -0,0 +1,49 @@
...
+
+env:
+  GOPATH: /tmp/go
+  GO_VERSION: 1.16.x
+
+jobs:
+  binary:
+    name: Build & push operator bundles for e2e tests
+
+    runs-on: ubuntu-18.04
```
xcoulon, line 13: just curious, what about running on Ubuntu 20.04 (see https://github.com/codeready-toolchain/host-operator/pull/470#discussion_r655521712)

MatousJobanek: I'm not sure what is the current status but at that time, when I had to set the specific version, we weren't able to use the latest nor 20.04 - there was a bug in a specific version of Podman that was automatically installed in Ubuntu via the used channel.

PR #205, opened 2021-08-06
Get configuration from toolchain config

rajivnathan:
Related PRs:
- https://github.com/codeready-toolchain/toolchain-e2e/pull/361
- https://github.com/codeready-toolchain/toolchain-common/pull/204

This PR updates the reg service to get its configuration from ToolchainConfig instead of env vars.

Related Issue: https://issues.redhat.com/browse/CRT-1122

Summary of changes:
- Updates startup of reg service to panic if WATCH_NAMESPACE is not set because it will be used directly for the lookup of the configuration resource
- Upon startup a go routine updates the configuration cache every 10 seconds
- Reg service initializes the configuration at startup and the configuration cache is used directly everywhere else in the code
- HTTP server configurations are now just constants since they're unlikely to require changes and if they do need to be changed we would probably want to change it across all environments


cmd/main.go
```diff
@@ -108,3 +109,18 @@ func gracefulShutdown(hs *http.Server, timeout time.Duration) {
...
+
+func configClient(cfg *rest.Config) (client.Client, error) {
+	scheme := runtime.NewScheme()
+	var AddToSchemes runtime.SchemeBuilder
+	addToSchemes := append(AddToSchemes,
+		corev1.AddToScheme,
+		toolchainv1alpha1.AddToScheme)
+	err := addToSchemes.AddToScheme(scheme)
+	if err != nil {
+		panic(err.Error())
```
alexeykazakov, line 121:
```suggestion
		return client.Client{}, err
```

alexeykazakov (approved):
Nice job!

cmd/main.go
```diff
@@ -79,13 +70,23 @@ func main() {
...
-	gracefulShutdown(srv.HTTPServer(), srv.Config().GetGracefulTimeout())
+	// update cache every 10 seconds
+	go func() {
+		for {
+			if _, err := commonconfig.ForceLoadToolchainConfig(cl); err != nil {
+				log.Error(nil, err, "failed to update the configuration cache")
+			}
+			time.Sleep(10 * time.Second)
+		}
+	}()
```
MatousJobanek, line 87: I usually use the `Until` function provided by `k8s.io/apimachinery/pkg/util/wait` for such use-cases:
```
go wait.Until(func() {
		...
	}, period, stopChannel)
``` 
but your version is fine too - just wanted to share another solution.

MatousJobanek (commented):
Looks good, but as I already said in the common PR - I believe that the registration-specific functions/methods/types should be here and not in toolchain-common repo https://github.com/codeready-toolchain/toolchain-common/pull/204#issuecomment-896020796

MatousJobanek:
One more thing - please don't forget to modify also the ToolchainConfig with the values from https://github.com/codeready-toolchain/host-operator/blob/master/deploy/env/prod.yaml in sandbox-sre

rajivnathan:
@MatousJobanek I updated the ToolchainConfig in sandbox-sre

rajivnathan:
/retest
after rebuilding operator bundles

MatousJobanek (approved):
:+1:

PR #219, opened 2021-10-21
configure logger to output in JSON format

xcoulon:
applies on
- zap logger: configured with custom options, including
  human-readable timestamps)
- Gin logger: configured in production mode with a custom
  formatter to output in JSON


xcoulon:
this will be needed so we can ingest the logs into Loki using the same JSON format as the host and member operators.

PR #223, opened 2021-11-01
WIP: Cluster proxy

alexeykazakov:
To be used by App Studio.

The proxy is built-into the same reg-service image and deployed as part of the reg-service container but it listens to a different port. It exposed as a separate service.

TODO:
- tests


cmd/proxy.go
```diff
@@ -0,0 +1,214 @@
...
+	var transport *http.Transport
+	if !configuration.GetRegistrationServiceConfig().IsProdEnvironment() {
+		transport = &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		}
+	}
+	return &httputil.ReverseProxy{
+		Director:  director,
+		Transport: transport,
+	}
```
MatousJobanek, line 174: so it flushes immediately after each write to the client
```suggestion
	return &httputil.ReverseProxy{
		Director:      director,
		Transport:     transport,
		FlushInterval: -1,
	}
```

alexeykazakov: Thanks! This PR is actually outdated. I've created a new one: https://github.com/codeready-toolchain/registration-service/pull/224
But I added this to the new PR.

alexeykazakov:
Replaced by https://github.com/codeready-toolchain/registration-service/pull/224

PR #224, opened 2021-11-03
Cluster proxy

alexeykazakov:
To be used by App Studio.

The proxy is built-into the same reg-service image and deployed as part of the reg-service container but it listens to a different port. It exposed as a separate service.

This PR is ready for review and getting merged (when approved). This PR is still missing a few things which IMO we can address in the following PRs:

- e2e tests
- add health check and readiness prob
- extra cache layer for the Namespace objects
- add a configuration param to disable proxy
- Make sure the proxy returns the same JSON error responses (when/if applicable) as a kube server would do. For now proxy just returns `err.Error()` string as the response body in case of an error.


pkg/configuration/configuration.go
```diff
@@ -25,6 +25,8 @@ var (
 	BuildTime = "0"
 	// StartTime in ISO 8601 (UTC) format.
 	StartTime = time.Now().UTC().Format("2006-01-02T15:04:05Z")
+
+	prodEnvironment = "prod"
```
xcoulon, line 29: this should be a const, not a var

pkg/proxy/proxy.go
```diff
@@ -0,0 +1,218 @@
...
+		req.URL.Host = target.APIURL.Host
+		req.URL.Path = singleJoiningSlash(target.APIURL.Path, req.URL.Path)
+		log.Info(ctx, fmt.Sprintf("forwarding %s to %s", origin, req.URL.String()))
+		if targetQuery == "" || req.URL.RawQuery == "" {
+			req.URL.RawQuery = targetQuery + req.URL.RawQuery
+		} else {
+			req.URL.RawQuery = targetQuery + "&" + req.URL.RawQuery
+		}
+		if _, ok := req.Header["User-Agent"]; !ok {
+			// explicitly disable User-Agent so it's not set to default value
```
xcoulon, line 162: why do we need to set `User-Agent` to empty value when it's missing?

pkg/proxy/namespace/namespace.go
```diff
@@ -0,0 +1,14 @@
...
+
+// Namespace represents a namespace in a member cluster for the specific sso user-workspace
+type Namespace struct {
+	Username    string
+	ClusterName string
+	APIURL      url.URL
+	Namespace   string
+	Workspace   string
+	// Target cluster token
+	TargetClusterToken string
```
xcoulon, line 13: what does `TargetClusterToken` mean exactly? Or can we rename it to `Token` instead?
Also, the struct is called `Namespace` but it has a `Username` field and a `Namespace` field, so should we rename this struct to `UserNamespace` to make it explicit that it's for a specific User in a specific Namespace?

pkg/application/service/services.go
```diff
@@ -19,7 +20,12 @@ type VerificationService interface {
...
 }
 
+type ToolchainClusterService interface {
+	GetNamespace(ctx *gin.Context, userID string) (*namespace.Namespace, error)
+}
+
 type Services interface {
 	SignupService() SignupService
 	VerificationService() VerificationService
+	ToolchainClusterService() ToolchainClusterService
```
xcoulon, line 30:
```suggestion
	MemberClusterService() MemberClusterService
```
since this service will look-up tokens on the member clusters

pkg/auth/defaultmanager.go
```diff
@@ -12,27 +12,23 @@ type DefaultTokenParserConfiguration interface {
...
 }
 
-var muKM sync.Mutex
-var muTP sync.Mutex
+var (
+	initDefaultTokenParserOnce *sync.Once = &sync.Once{}
 
-var defaultKeyManagerHolder *KeyManager
-var defaultTokenParserHolder *TokenParser
+	defaultKeyManagerHolder  *KeyManager
```
xcoulon, line 18: why do we have those `Holder` suffixes? What do they mean?

pkg/proxy/proxy.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+	return token[1], nil
+}
+
+func (p *Proxy) serveReverseProxy(ctx *gin.Context, target *namespace.Namespace, res http.ResponseWriter, req *http.Request) {
+	proxy := p.newReverseProxy(ctx, target)
+
+	// Note that ServeHttp is non blocking and uses a go routine under the hood
+	proxy.ServeHTTP(res, req)
+}
```
xcoulon, line 146: for the sake of readability, I would suggest that these 2 lines were moved in line 93, inside the `handleRequestAndRedirect` func

pkg/proxy/proxy.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+	return newProxyWithClusterClient(app, config, cl)
+}
+
+func newProxyWithClusterClient(app application.Application, config configuration.RegistrationServiceConfig, cln client.Client) (*Proxy, error) {
+	// Initiate toolchain cluster cache service
+	cacheLog := controllerlog.Log.WithName("registration-service")
+	cluster.NewToolchainClusterService(cln, cacheLog, config.Namespace(), 5*time.Second)
+
+	tokenParserInstance, err := auth.DefaultTokenParser()
```
xcoulon, line 52:
```suggestion
	tokenParser, err := auth.DefaultTokenParser()
```

pkg/proxy/proxy.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+	}
+	return &httputil.ReverseProxy{
+		Director:      director,
+		Transport:     transport,
+		FlushInterval: -1,
+	}
+}
+
+func singleJoiningSlash(a, b string) string {
```
xcoulon, line 181: this would deserve a few unit tests but I wonder if we could use `filepath.Join` instead

pkg/proxy/proxy.go
```diff
@@ -0,0 +1,218 @@
...
+	a := req.Header.Get("Authorization")
+	token := strings.Split(a, "Bearer ")
+	if len(token) < 2 {
+		return "", crterrors.NewUnauthorizedError("no token found", "a Bearer token is expected")
+	}
+	return token[1], nil
+}
+
+func (p *Proxy) serveReverseProxy(ctx *gin.Context, target *namespace.Namespace, res http.ResponseWriter, req *http.Request) {
+	proxy := p.newReverseProxy(ctx, target)
```
xcoulon, line 142: let's avoid calling this var `proxy`, it is confusing because we have a type called `Proxy` which has nothing to do with it

xcoulon: or just `p.newReverseProxy(ctx, target).ServeHTTP(res, req)`

pkg/proxy/cache.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/codeready-toolchain/registration-service/pkg/application"
+	"github.com/codeready-toolchain/registration-service/pkg/proxy/namespace"
+	"github.com/gin-gonic/gin"
+)
+
+type UserNamespaces struct {
+	app application.Application
+}
+
+func NewUserNamespaces(app application.Application) *UserNamespaces {
```
xcoulon, line 13: this is a struct with a single method. Could we refactor this to a func instead?

pkg/proxy/proxy.go
```diff
@@ -0,0 +1,218 @@
...
+	// Initiate toolchain cluster cache service
+	cacheLog := controllerlog.Log.WithName("registration-service")
+	cluster.NewToolchainClusterService(cln, cacheLog, config.Namespace(), 5*time.Second)
+
+	tokenParserInstance, err := auth.DefaultTokenParser()
+	if err != nil {
+		return nil, err
+	}
+	return &Proxy{
+		namespaces:  NewUserNamespaces(app),
```
xcoulon, line 57: instead of using a struct with a single method, could we refactor this `namespaces` field to a func such as:
```
type Proxy struct {
	getNamespaces  func(ctx *gin.Context, userID string) (*namespace.Namespace, error)
	tokenParser *auth.TokenParser
}
```

so we could do something like
```
p.getNamespace(ctx, userID)
```
instead of 
```
p.namespaces.GetNamespace(ctx, userID)
```
in line 116

pkg/proxy/namespace/namespace.go
```diff
@@ -0,0 +1,14 @@
...
+
+import "net/url"
+
+// Namespace represents a namespace in a member cluster for the specific sso user-workspace
+type Namespace struct {
+	Username    string
+	ClusterName string
+	APIURL      url.URL
+	Namespace   string
+	Workspace   string
```
xcoulon, line 11: same here: this field is never used

pkg/application/service/services.go
alexeykazakov: Done in https://github.com/codeready-toolchain/registration-service/pull/224/commits/c44b99a8fc4b8ad059ed484cf3cf58cad94cb06d

pkg/auth/defaultmanager.go
alexeykazakov: I have no idea but I personally don't see any particular reason. That code was not created by me. I'm just slightly re-factoring and re-formatting it so it can be used by the proxy too.

pkg/configuration/configuration.go
alexeykazakov: Done in https://github.com/codeready-toolchain/registration-service/pull/224/commits/f98ba62e3e1b53a95e8a7db698f66ea23cd7af9b

pkg/proxy/cache.go
alexeykazakov: We still need a struct to implement the cache. For now it's just a place holder for the cache we start working on when done with this PR.

pkg/proxy/namespace/namespace.go
```diff
@@ -0,0 +1,14 @@
+package namespace
+
+import "net/url"
+
+// Namespace represents a namespace in a member cluster for the specific sso user-workspace
+type Namespace struct {
+	Username    string
```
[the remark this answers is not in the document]
alexeykazakov: We will need to modify that code when we switch to space-based namespaces from user-based and also when we add the proxy cache. So I just guessed what we would need for that. But I removed those unused fields for now in https://github.com/codeready-toolchain/registration-service/pull/224/commits/5f44dca3ee43d396eae35a0634208d05e0125154
Will add them back when needed.

pkg/proxy/namespace/namespace.go
```diff
@@ -0,0 +1,14 @@
...
+	ClusterName string
+	APIURL      url.URL
+	Namespace   string
```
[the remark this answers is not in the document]
alexeykazakov: https://github.com/codeready-toolchain/registration-service/pull/224#discussion_r745990333

pkg/proxy/namespace/namespace.go
```diff
@@ -0,0 +1,14 @@
+package namespace
+
+import "net/url"
+
+// Namespace represents a namespace in a member cluster for the specific sso user-workspace
+type Namespace struct {
+	Username    string
+	ClusterName string
```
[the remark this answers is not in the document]
alexeykazakov: https://github.com/codeready-toolchain/registration-service/pull/224#discussion_r745990333

pkg/proxy/namespace/namespace.go
alexeykazakov: https://github.com/codeready-toolchain/registration-service/pull/224#discussion_r745990333

pkg/proxy/namespace/namespace.go
alexeykazakov: I added a better comment to the `TargetClusterToken` field.
Regarding the `UserNamespace` vs `Namespace`, this structs actually represents the information needed for accessing specific namespace for specific user in the specific workspace. So, there are three dimensions here: Namespace, User, Workspace. It's so because user can have access to different workspaces.
And while I'm agree that `Namespace` is not perfect I don't thing that UserNamespace will be much better.

pkg/proxy/proxy.go
alexeykazakov: We could but I think it's just matter of personal preferences and not that critical in this case so I would keep it as is for now if you don't mind :)

pkg/proxy/proxy.go
alexeykazakov: Done in https://github.com/codeready-toolchain/registration-service/pull/224/commits/d5534b9647db7e4c182c10ae074c78e1b524ed83

pkg/proxy/proxy.go
alexeykazakov: Done in https://github.com/codeready-toolchain/registration-service/pull/224/commits/d5534b9647db7e4c182c10ae074c78e1b524ed83

pkg/proxy/proxy.go
alexeykazakov: Done in https://github.com/codeready-toolchain/registration-service/pull/224/commits/d5534b9647db7e4c182c10ae074c78e1b524ed83

pkg/proxy/proxy.go
alexeykazakov: As the comment is saying its to avoid using the default value. But I don't know why we would need that. It's something I saw in the reverse proxy examples. Let's keep it here until we find out we don't want or need that :)

rajivnathan (approved):
This is cool! Just have a few minor comments

cmd/main.go
```diff
@@ -74,6 +77,18 @@ func main() {
 		panic(err.Error())
 	}
 
+	_, err = auth.InitializeDefaultTokenParser()
+	if err != nil {
+		panic(errs.Wrapf(err, "failed to init default token parser: %s", err.Error()))
```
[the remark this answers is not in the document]
alexeykazakov: Right. Good catch.

pkg/proxy/proxy.go
```diff
@@ -0,0 +1,212 @@
...
+	cl, err := newClusterClient()
+	if err != nil {
+		return nil, err
+	}
+	return newProxyWithClusterClient(app, config, cl)
+}
+
+func newProxyWithClusterClient(app application.Application, config configuration.RegistrationServiceConfig, cln client.Client) (*Proxy, error) {
+	// Initiate toolchain cluster cache service
+	cacheLog := controllerlog.Log.WithName("registration-service")
```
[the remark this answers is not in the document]
alexeykazakov: Unfortunatly this specific log is used by the kube client which is used by the cluster service from common (the service which pulls the member clusters and caches them). It's not used by the rest of the proxy. We use the same log for the reg-service and for the rest of the proxy. I thought it would be confusing if some small part of the proxy uses one logger and the rest of the proxy uses a different one.

pkg/proxy/cache.go
```diff
@@ -0,0 +1,22 @@
+package proxy
+
+import (
+	"github.com/codeready-toolchain/registration-service/pkg/application"
+	"github.com/codeready-toolchain/registration-service/pkg/proxy/namespace"
+	"github.com/gin-gonic/gin"
+)
+
+type UserNamespaces struct {
```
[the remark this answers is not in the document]
alexeykazakov: Maybe.. but let's keep it as is for now. For now it's just a place holder. We can change it when implementing the cache.

pkg/auth/defaultmanager_test.go
```diff
@@ -164,22 +117,10 @@ func (s *TestDefaultManagerSuite) TestKeyManagerDefaultTokenParser() {
...
-			if thisEntry.TokePrsr == nil && thisEntry.TpErr != nil {
-				fails++
-			}
-			if (thisEntry.TokePrsr == nil && thisEntry.TpErr == nil) || (thisEntry.TokePrsr != nil && thisEntry.TpErr != nil) {
-				require.Fail(s.T(), "unexpected return values when calling InitializeDefaultTokenParser")
-			}
+		// check if only all entries have a TokenParser
+		for _, entry := range holder {
+			require.NoError(s.T(), entry.TpErr)
+			require.NotNil(s.T(), entry.TokePrsr)
```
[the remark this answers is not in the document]
alexeykazakov: Done in https://github.com/codeready-toolchain/registration-service/pull/224/commits/33193acd690ddac479543507beabb2bc5dc83d65

pkg/proxy/service/cluster_service.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/codeready-toolchain/toolchain-common/pkg/cluster"
+
+	errs "github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+type Option func(f *ServiceImpl)
+
+// ServiceImpl represents the implementation of the signup service.
```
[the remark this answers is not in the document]
alexeykazakov: Yep. Thanks. Done in https://github.com/codeready-toolchain/registration-service/pull/224/commits/33193acd690ddac479543507beabb2bc5dc83d65

pkg/proxy/proxy.go
alexeykazakov: I would avoid using `filepath.Join` because it's OC specific.
Added more tests in https://github.com/codeready-toolchain/registration-service/pull/224/commits/d633a95af25cbc705ba48f0ce35afcd03c05bdaa

alexeykazakov:
@xcoulon @rajivnathan thanks for the review!
@xcoulon I've addressed or commented on all your comments so please review.

pkg/proxy/proxy.go
xcoulon: > @xcoulon @rajivnathan thanks for the review! @xcoulon I've addressed or commented on all your comments so please review.

thanks for taking my comments into account @alexeykazakov

xcoulon (approved):

pkg/proxy/proxy_test.go
```diff
@@ -0,0 +1,243 @@
...
+	err        error
+}
+
+func (a *fakeApp) SignupService() service.SignupService {
+	panic("implement me")
+}
+
+func (a *fakeApp) VerificationService() service.VerificationService {
+	panic("implement me")
+}
```
MatousJobanek, line 231: do we really expect that someone would implement it? I would rather say that this shouldn't be called.
```suggestion
func (a *fakeApp) SignupService() service.SignupService {
	panic("SignupService shouldn't be called")
}

func (a *fakeApp) VerificationService() service.VerificationService {
	panic("VerificationService shouldn't be called")
}
```

MatousJobanek (commented):
Functional-wise looks good :+1: :1st_place_medal: 
I left a few suggestions/comments

pkg/proxy/service/cluster_service.go
```diff
@@ -0,0 +1,103 @@
...
+				log.Info(ctx, fmt.Sprintf("Getting secret %v", secretNamespacedName))
+				if err := member.Client.Get(context.TODO(), secretNamespacedName, s); err != nil {
+					return nil, err
+				}
+				if s.Annotations["kubernetes.io/created-by"] == "openshift.io/create-dockercfg-secrets" {
+					// There are two secrets/tokens for the SA and both are valid
+					// but let's always use the non-docker one for the sake of consistency
+					log.Info(nil, fmt.Sprintf("Skipping docker secret with the creted-by label: %v", secretNamespacedName))
+					continue
+				}
```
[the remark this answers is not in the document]
alexeykazakov: Both docker and the regular tokens/secrets have `type: kubernetes.io/service-account-token`. I wanted to use the regular one, not the docker one. Both are valid tokens and can be used to access the namespace but:
1. the regular one has less permissions (can't use for the registry)
2. Once I noticed that the docker secret the token... It was a different secret structure. Not sure why... So it seems that it's safer to rely on the regular secret only and skip the docker one completely.

pkg/proxy/proxy.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+func extractUserToken(req *http.Request) (string, error) {
+	a := req.Header.Get("Authorization")
+	token := strings.Split(a, "Bearer ")
+	if len(token) < 2 {
+		return "", crterrors.NewUnauthorizedError("no token found", "a Bearer token is expected")
+	}
+	return token[1], nil
+}
```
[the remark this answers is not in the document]
alexeykazakov: Yes, it's similar but we can't just re-use it as is. And I don't think extracting the common code from both functions will give us anything. It's just a few lines of a very simple code.

pkg/proxy/namespace/namespace.go
alexeykazakov: > Also the comment of the type - because it doesn't represent the namespace at all. It doesn't have any name of the namespace stored anywhere, but correct me if I'm wrong.

It did have the namespace name. Initially the struct had all the information about the namespace in case we need it in the future. The purpose of this struct was more general then just the SA token. But Xavier asked me to remove it because it's not use yet :)
So, it's really about an access to specific namespace for a specific user for a specific workspace.
So, maybe I will rename it to ClusterAccess but I'm not super happy with this name either...

pkg/proxy/service/cluster_service.go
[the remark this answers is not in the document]
MatousJobanek: I didn't know that - that's good to know. :+1:

pkg/proxy/proxy.go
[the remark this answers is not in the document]
MatousJobanek: fair enough :+1:

pkg/auth/defaultmanager.go
```diff
@@ -12,27 +12,23 @@ type DefaultTokenParserConfiguration interface {
...
+	defaultTokenParserHolder *TokenParser
+)
 
 // InitializeDefaultKeyManager creates the default key manager if it has not created yet.
-// This function must be called in main to make sure the default manager is created during service startup.
-// It will try to create the default manager only once even if called multiple times.
 func initializeDefaultKeyManager() (*KeyManager, error) {
-	muKM.Lock()
-	defer muKM.Unlock()
 	if defaultKeyManagerHolder == nil {
```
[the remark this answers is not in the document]
alexeykazakov: Right. I've cleaned it up even further: https://github.com/codeready-toolchain/registration-service/pull/224/commits/6de5fdf306ee588db4fdb04dfb078b08ce306fac

pkg/errors/error.go
```diff
@@ -37,6 +37,15 @@ func NewForbiddenError(message, details string) *Error {
...
 }
 
+func NewUnauthorizedError(message, details string) *Error {
+	return &Error{
+		Status:  http.StatusText(http.StatusUnauthorized),
+		Code:    http.StatusUnauthorized,
+		Message: message,
+		Details: details,
+	}
+}
```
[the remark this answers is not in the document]
alexeykazakov: But how it would help us? We have to return an http error response (`http.Error()`) which requires three params: the http response writer, the status code (401 in our case) and the error message. We provide all of it.
`Reason: "Unauthorized"` is an implementation detail in the kube status error object which doesn't add anything to the http error we are already returning. Unless I'm missing something.

pkg/proxy/proxy_test.go
```diff
@@ -0,0 +1,243 @@
...
+			require.NoError(s.T(), err)
+			require.NotNil(s.T(), req)
+			_, err = http.DefaultClient.Do(req)
+			if err != nil {
+				time.Sleep(time.Second)
+				continue
+			}
+			// Server is up and running!
+			break
+		}
```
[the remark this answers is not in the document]
alexeykazakov: Agreed. Moved here: https://github.com/codeready-toolchain/registration-service/pull/224/commits/6f81d0ed428e0bccbc23db4cbd0c86f4e147d109

pkg/proxy/namespace/namespace.go
alexeykazakov: I've renamed it to "NamespaceAccess" in https://github.com/codeready-toolchain/registration-service/pull/224/commits/b5e7f4e8e266901c0d91394b184396abf38664da

pkg/proxy/proxy.go
```diff
@@ -0,0 +1,212 @@
...
+	if err := v1.AddToScheme(scheme); err != nil {
+		return nil, err
+	}
+	if err := toolchainv1alpha1.AddToScheme(scheme); err != nil {
+		return nil, err
+	}
+	toolchainv1alpha1.SchemeBuilder.Register(
+		&toolchainv1alpha1.ToolchainCluster{},
+		&toolchainv1alpha1.ToolchainClusterList{},
+	)
```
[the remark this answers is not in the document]
alexeykazakov: Good catch. Removed in https://github.com/codeready-toolchain/registration-service/pull/224/commits/39be94aeaf05bb5995166638648d735e734a8bbc

pkg/auth/defaultmanager.go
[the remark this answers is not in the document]
MatousJobanek: Nice :+1:

pkg/errors/error.go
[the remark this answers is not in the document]
MatousJobanek: The main motivation was to return the same error as K8s would return, so the client doesn't recognize that it's not a real K8s server. If any client would use the `Reason` for anything (as it's always returned by K8s) then I thought that it would be fine to add it there.
But I agree that it's just a detail and that `Status`, `Code`, and `Message` are the only fields that really matter.

pkg/proxy/namespace/namespace.go
MatousJobanek: `NamespaceAccess` sounds good as well. Thanks :+1:

MatousJobanek (approved):
Good job :+1: :1st_place_medal: 
Thanks for addressing my comments

pkg/errors/error.go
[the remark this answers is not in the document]
alexeykazakov: OK. I'm guessing that you are talking about the http response body here. Basically the JSON returned by the kube or proxy.
For now proxy returns `err.Error()` as the error body. It doesn't marshal any JSON. So the underlying error struct doesn't really matter.
We need to improve it and return the same JSON a kube server would return but it requires more work than just changing our internal error structs (or switching to kube error objects).
So it's a valid point but let's do it in a separate PR. I will add it to the TODO list in the PR description.

PR #228, opened 2021-11-19
Changes required for proxy e2e tests

rajivnathan:
Related issue: https://issues.redhat.com/browse/CRT-1274

Related PRs:
- https://github.com/codeready-toolchain/toolchain-e2e/pull/413

Summary of changes:
- update the call to `cluster.NewToolchainClusterService()` to use the current namespace of the reg service because the RegistrationServiceConfig.Namespace() configuration defaulted to "toolchain-host-operator" but in e2e the host operator namespace is dynamic
- remove Namespace() func from RegistrationServiceConfig because it was confusing and not used anywhere else
- update the check in GetNamespace() to also check the name of the member cluster `member.Name == signup.ClusterName` because the api endpoint is the same in e2e test environment so testing multi-member scenarios led to mismatching the member clusters
  - updated the Signup struct to include ClusterName to support this

### DIFF
--- a/deploy/registration-service.yaml
+++ b/deploy/registration-service.yaml
@@ -48,6 +48,7 @@ objects:
         resources:
           - secrets
           - configmaps
+          - serviceaccounts
         verbs:
           - get
           - list
--- a/deploy/registration-service.yaml
+++ b/deploy/registration-service.yaml
@@ -48,7 +48,6 @@ objects:
         resources:
           - secrets
           - configmaps
-          - serviceaccounts
         verbs:
           - get
           - list
--- a/pkg/proxy/service/cluster_service.go
+++ b/pkg/proxy/service/cluster_service.go
@@ -56,6 +56,8 @@ func (s *ServiceImpl) GetNamespace(ctx *gin.Context, userID string) (*namespace.
 		return nil, errs.New("no member clusters found")
 	}
 	for _, member := range members {
+		// also check that the member cluster name matches because the api endpoint is the same for both members
+		// in the e2e tests because a single cluster is used for testing multi-member scenarios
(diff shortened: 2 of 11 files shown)


pkg/proxy/service/cluster_service.go
```diff
@@ -56,7 +56,7 @@ func (s *ServiceImpl) GetNamespace(ctx *gin.Context, userID string) (*namespace.
 		return nil, errs.New("no member clusters found")
 	}
 	for _, member := range members {
-		if member.APIEndpoint == signup.APIEndpoint {
+		if member.APIEndpoint == signup.APIEndpoint && member.Name == signup.ClusterName {
```
alexeykazakov, line 59: I guess you need it because of the e2e test env when we actually have two members in the same physical cluster?
Let's add a comment here. Because it might be unclear why just checking the API Endpoint is not enough.

deploy/registration-service.yaml
```diff
@@ -48,6 +48,7 @@ objects:
         resources:
           - secrets
           - configmaps
+          - serviceaccounts
```
alexeykazakov, line 51: Same question as in the host operator PR. Why do you need it?

pkg/proxy/service/cluster_service.go
rajivnathan: Done: c3fa10f2df514a1f696c726ec0eaf35086046b8a

deploy/registration-service.yaml
rajivnathan: You're right, it's not needed. Removed it

MatousJobanek (approved):
Looks good :+1:

alexeykazakov (approved):

PR #230, opened 2021-11-23
Health/readiness probe for proxy

rajivnathan:
Related Issue: https://issues.redhat.com/browse/CRT-1372

There can only be one health endpoint per container and since we have the reg service and proxy servers both in the same container there needs to be a combined health endpoint.

Summary of changes:
- Adds a health endpoint to the proxy server
- Modified the reg service health checker to also initiate a GET request to the proxy health endpoint

### DIFF
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -80,7 +80,10 @@ func (p *Proxy) StartProxy() *http.Server {
 func (p *Proxy) health(res http.ResponseWriter, req *http.Request) {
 	res.Header().Set("Content-Type", "application/json")
 	res.WriteHeader(http.StatusOK)
-	io.WriteString(res, `{"alive": true}`)
+	_, err := io.WriteString(res, `{"alive": true}`)
+	if err != nil {
+		log.Error(nil, err, "failed to write health response")
+	}
 }
 
 func (p *Proxy) handleRequestAndRedirect(res http.ResponseWriter, req *http.Request) {
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -3,6 +3,7 @@ package proxy
 import (
 	"crypto/tls"
 	"fmt"
+	"io"
 	"net/http"
 	"net/http/httputil"
 	"strings"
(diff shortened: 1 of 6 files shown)


pkg/proxy/proxy.go
```diff
@@ -63,19 +64,28 @@ func (p *Proxy) StartProxy() *http.Server {
 	// start server
 	mux := http.NewServeMux()
 	mux.HandleFunc("/", p.handleRequestAndRedirect)
+	mux.HandleFunc("/health", p.health)
```
alexeykazakov, line 67: There's no overlapping with the kube health endpoints, right? I guess all kube endpoints start with /api ?

alexeykazakov (approved):

pkg/proxy/proxy.go
rajivnathan: Good point, I didn't think about that. It sounds like there is no overlap: https://kubernetes.io/docs/reference/using-api/#api-groups

MatousJobanek: I think that @alexeykazakov didn't mean the k8s api endpoint, but any of the "health" endpoints (there are more of them in k8s https://kubernetes.io/docs/reference/using-api/health-checks/) that k8s cluster provides. If I wanted to check whether my particular member cluster is healthy or not and I wanted to use proxy for it, then I would send a request to the  `/healthz` endpoint.
Looking at the set of the paths that k8s provides, there should be no overlapping. But it could be maybe more transparent if we used for example `/proxyhealth`  endpoint. WDYT?

MatousJobanek: I just read the description of the PR :man_facepalming: yeah, simple `/health` is probably better one :-)

alexeykazakov: Yeah I was thinking about something like /proxyhealth too. Just to make sure it's unique. /health should be ok because it seems there is no overlapping with the kube health endpoints. At least for now. Anyway I leave it to you Rajiv. It's your call :)

rajivnathan: I changed it to `/proxyhealth` since it's less ambiguous and reduces any possibility of overlap in the future. 🙂

PR #244, opened 2022-01-25
chore: update linter config

xcoulon:
disable 'golint', use 'revive' instead

also, apply changes suggested by linters

### DIFF
--- a/pkg/auth/keymanager.go
+++ b/pkg/auth/keymanager.go
@@ -4,7 +4,7 @@ import (
 	"bytes"
 	"crypto/rsa"
 	"encoding/json"
-	stderrors "errors"
+	"errors"
 	"io"
 	"io/ioutil"
 	"net/http"
@@ -77,7 +77,7 @@ func NewKeyManager() (*KeyManager, error) {
 func (km *KeyManager) Key(kid string) (*rsa.PublicKey, error) {
 	key, ok := km.keyMap[kid]
 	if !ok {
-		return nil, stderrors.New("unknown kid")
+		return nil, errors.New("unknown kid")
 	}
 	return key, nil
 }
@@ -113,7 +113,7 @@ func (km *KeyManager) unmarshalKey(jsonData []byte) (*PublicKey, error) {
 	}
 	rsaKey, ok := key.Key.(*rsa.PublicKey)
 	if !ok {
-		return nil, stderrors.New("Key is not an *rsa.PublicKey")
+		return nil, errors.New("Key is not an *rsa.PublicKey")
 	}
 	return &PublicKey{key.KeyID, rsaKey}, nil
 }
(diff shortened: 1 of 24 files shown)


alexeykazakov (approved):

pkg/auth/keymanager.go
```diff
@@ -4,7 +4,7 @@ import (
 	"bytes"
 	"crypto/rsa"
 	"encoding/json"
-	"errors"
+	stderrors "errors"
```
MatousJobanek, line 7: why? I would rather use just `errors`.

xcoulon: ah, that's because I used the same alias in another file, and since the repo has its own `errors` package, I wanted to make sure that the standard `errors` package is aliased everywhere (for the sake of consistency). Plus, there's also the `k8s.io/apimachinery/pkg/api/errors` package which I aliased to `apierrors`.

MatousJobanek: so what about `errs` as is used in other repos?

xcoulon: `errs` is usually `github.com/pkg/errors` (AFAIR)

sbryzak (approved):
Nice cleanup!

xcoulon:
> Nice cleanup!

thanks for your feedback, @sbryzak ! 🙌

pkg/auth/keymanager.go
MatousJobanek: my point is - I'm not aware of any single file where we would use `stderrors` as an alias for `errors` import. we usually create custom aliases for the other error imports not this standard one. I thought that it would be fine to keep it consistent with other files/repos.
I just shared my preference, but I don't wan to block this PR because of the import name (which is really a detail).

xcoulon: fair point. I noticed the usage of "crterrors" in other files of this repo, so I will use that instead, and keep "errors" as-is

xcoulon: see ccbb84a

pkg/kubeclient/banneduser.go
```diff
@@ -2,7 +2,7 @@ package kubeclient
 
 import (
 	"context"
-	"crypto/md5"
+	"crypto/md5" //nolint:gosec
```
MatousJobanek, line 5: just out of curiosity - what is the error we ignore here?

MatousJobanek (approved):

pkg/kubeclient/banneduser.go
xcoulon: md5 is a weak crypto algorithm, but since already use it, we would need to update all our resources if we decide to use something else, such as sha-1 for example.

PR #247, opened 2022-01-28
Allow credentials for proxy CORS requests

alexeykazakov:


MatousJobanek (approved):

MatousJobanek:
LGTM

pkg/proxy/proxy_test.go
```diff
@@ -250,7 +250,10 @@ func (s *TestProxySuite) TestProxy() {
 					"plain http": {
 						ProxyHeaders:                    map[string]string{"Authorization": "Bearer " + s.token(userID)},
 						ExpectedAPIServerRequestHeaders: map[string]string{"Authorization": "Bearer clusterSAToken"},
-						ExpectedProxyResponseHeaders:    map[string]string{"Access-Control-Allow-Origin": "*"},
+						ExpectedProxyResponseHeaders: map[string]string{
+							"Access-Control-Allow-Origin":      "*",
+							"Access-Control-Allow-Credentials": "true",
```
andrewballantyne, line 255: I do not believe these are not compatible. See [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#credentialed_requests_and_wildcards). I think we need to specify the URLs that is allowed. Pete made [this comment](https://coreos.slack.com/archives/C02CTEB3MMF/p1643386835498229?thread_ts=1643303583.137000&cid=C02CTEB3MMF) on the thread with a suggestion.

andrewballantyne (changes requested):
/hold

PR #249, opened 2022-01-31
feat: echo origin for CORS header

MatousJobanek:
Echo the same origin as was received in the request

### DIFF
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -109,6 +109,7 @@ func (p *Proxy) handleRequestAndRedirect(res http.ResponseWriter, req *http.Requ
 	}
 
 	// Note that ServeHttp is non blocking and uses a go routine under the hood
+	addCorsToResponse(res, req)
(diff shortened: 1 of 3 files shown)


karelhala (approved):

xcoulon (approved):

pkg/proxy/proxy.go
```diff
@@ -199,10 +200,9 @@ func (p *Proxy) newReverseProxy(ctx *gin.Context, target *namespace.NamespaceAcc
 		}
 	}
 	return &httputil.ReverseProxy{
-		Director:       director,
-		Transport:      transport,
-		FlushInterval:  -1,
-		ModifyResponse: addCorsToResponse,
```
alexeykazakov, line 205: I'm a bit concerned with modifying the response headers before we redirect the request to the api server instead of using a response modifier. In this case our CORS headers can be overridden by the api server based on the default strict api server settings.

MatousJobanek: Good point - I didn't realize that. :man_facepalming:

PR #251, opened 2022-02-03
Modify registration service to support locating UserSignup by username if userID lookup fails

sbryzak:
This PR updates the registration service to allow it to locate UserSignup resources by the user's username (i.e. the `username` claim in the JWT token) if the user ID (i.e. the `sub` claim) lookup fails.

Fixes https://issues.redhat.com/browse/CRT-1398

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/472

### DIFF
--- a/pkg/controller/signup_test.go
+++ b/pkg/controller/signup_test.go
@@ -448,7 +448,7 @@ func (s *TestSignupSuite) TestInitVerificationHandler() {
 			MockUpdateUserSignup: func(userSignup *crtapi.UserSignup) (userSignup2 *crtapi.UserSignup, e error) {
 				return userSignup, nil
 			},
-			MockPhoneNumberAlreadyInUse: func(userID, e164phoneNumber string) error {
+			MockPhoneNumberAlreadyInUse: func(userID, username, e164phoneNumber string) error {
 				return nil
 			},
 		}
@@ -729,7 +729,7 @@ type FakeSignupService struct {
 	MockSignup                  func(ctx *gin.Context) (*crtapi.UserSignup, error)
 	MockGetUserSignup           func(userID, username string) (*crtapi.UserSignup, error)
 	MockUpdateUserSignup        func(userSignup *crtapi.UserSignup) (*crtapi.UserSignup, error)
-	MockPhoneNumberAlreadyInUse func(userID, value string) error
+	MockPhoneNumberAlreadyInUse func(userID, username, value string) error
 }
 
 func (m *FakeSignupService) GetSignup(userID, username string) (*signup.Signup, error) {
@@ -748,6 +748,6 @@ func (m *FakeSignupService) UpdateUserSignup(userSignup *crtapi.UserSignup) (*cr
 	return m.MockUpdateUserSignup(userSignup)
 }
 
-func (m *FakeSignupService) PhoneNumberAlreadyInUse(userID, e164phoneNumber string) error {
-	return m.MockPhoneNumberAlreadyInUse(userID, e164phoneNumber)
+func (m *FakeSignupService) PhoneNumberAlreadyInUse(userID, username, e164phoneNumber string) error {
+	return m.MockPhoneNumberAlreadyInUse(userID, username, e164phoneNumber)
 }
--- a/pkg/controller/signup_test.go
+++ b/pkg/controller/signup_test.go
@@ -691,7 +691,7 @@ func (s *TestSignupSuite) TestVerifyCodeHandler() {
 			Key:   "code",
 			Value: "999127",
 		}
-		rr := initVerification(s.T(), handler, param, nil, otherUserSignup.Spec.Userid, otherUserSignup.Spec.Username, http.MethodGet, "/api/v1/signup/verification")
+		rr := initVerification(s.T(), handler, param, nil, "", otherUserSignup.Spec.Username, http.MethodGet, "/api/v1/signup/verification")
 
 		// Check the status code is what we expect.
 		require.Equal(s.T(), http.StatusOK, rr.Code)
--- a/pkg/controller/signup_test.go
+++ b/pkg/controller/signup_test.go
@@ -201,7 +201,7 @@ func (s *TestSignupSuite) TestSignupGetHandler() {
 				Reason: "Provisioning",
 			},
 		}
-		svc.MockGetSignup = func(id string) (*signup.Signup, error) {
+		svc.MockGetSignup = func(id, username string) (*signup.Signup, error) {
 			if id == userID {
 				return expected, nil
 			}
@@ -228,7 +228,7 @@ func (s *TestSignupSuite) TestSignupGetHandler() {
 		ctx.Request = req
 		ctx.Set(context.SubKey, userID)
 
-		svc.MockGetSignup = func(id string) (*signup.Signup, error) {
+		svc.MockGetSignup = func(id, username string) (*signup.Signup, error) {
 			return nil, nil
 		}
 
@@ -245,7 +245,7 @@ func (s *TestSignupSuite) TestSignupGetHandler() {
 		ctx.Request = req
 		ctx.Set(context.SubKey, userID)
 
-		svc.MockGetSignup = func(id string) (*signup.Signup, error) {
+		svc.MockGetSignup = func(id, username string) (*signup.Signup, error) {
 			return nil, errors.New("oopsie woopsie")
 		}
 
@@ -429,7 +429,7 @@ func (s *TestSignupSuite) TestInitVerificationHandler() {
 
 		// Create a mock SignupService
 		svc := &FakeSignupService{
-			MockGetUserSignup: func(userID string) (userSignup *crtapi.UserSignup, e error) {
+			MockGetUserSignup: func(userID, username string) (userSignup *crtapi.UserSignup, e error) {
 				us := crtapi.UserSignup{
 					TypeMeta: v1.TypeMeta{},
 					ObjectMeta: v1.ObjectMeta{
@@ -674,23 +674,23 @@ func initVerification(t *testing.T, handler gin.HandlerFunc, params gin.Param, d
 }
 
 type FakeSignupService struct {
-	MockGetSignup               func(userID string) (*signup.Signup, error)
+	MockGetSignup               func(userID, username string) (*signup.Signup, error)
 	MockSignup                  func(ctx *gin.Context) (*crtapi.UserSignup, error)
-	MockGetUserSignup           func(userID string) (*crtapi.UserSignup, error)
+	MockGetUserSignup           func(userID, username string) (*crtapi.UserSignup, error)
 	MockUpdateUserSignup        func(userSignup *crtapi.UserSignup) (*crtapi.UserSignup, error)
 	MockPhoneNumberAlreadyInUse func(userID, value string) error
 }
 
-func (m *FakeSignupService) GetSignup(userID string) (*signup.Signup, error) {
-	return m.MockGetSignup(userID)
+func (m *FakeSignupService) GetSignup(userID, username string) (*signup.Signup, error) {
+	return m.MockGetSignup(userID, username)
 }
 
 func (m *FakeSignupService) Signup(ctx *gin.Context) (*crtapi.UserSignup, error) {
 	return m.MockSignup(ctx)
 }
 
-func (m *FakeSignupService) GetUserSignup(userID string) (*crtapi.UserSignup, error) {
-	return m.MockGetUserSignup(userID)
+func (m *FakeSignupService) GetUserSignup(userID, username string) (*crtapi.UserSignup, error) {
+	return m.MockGetUserSignup(userID, username)
 }
 
 func (m *FakeSignupService) UpdateUserSignup(userSignup *crtapi.UserSignup) (*crtapi.UserSignup, error) {
--- a/pkg/controller/signup_test.go
+++ b/pkg/controller/signup_test.go
@@ -300,7 +300,7 @@ func (s *TestSignupSuite) TestInitVerificationHandler() {
 			BodyString("")
 
 		data := []byte(fmt.Sprintf(`{"phone_number": "%s", "country_code": "1"}`, phoneNumber))
-		rr := initVerification(s.T(), handler, gin.Param{}, data, userID, http.MethodPut, "/api/v1/signup/verification")
+		rr := initVerification(s.T(), handler, gin.Param{}, data, userID, "", http.MethodPut, "/api/v1/signup/verification")
 		require.Equal(s.T(), http.StatusNoContent, rr.Code)
 
 		updatedUserSignup, err := s.FakeUserSignupClient.Get(userSignup.Name)
@@ -333,7 +333,7 @@ func (s *TestSignupSuite) TestInitVerificationHandler() {
 			BodyString("")
 
 		data := []byte(`{"phone_number": "2268213044", "country_code": "(1)"}`)
-		rr := initVerification(s.T(), handler, gin.Param{}, data, userID, http.MethodPut, "/api/v1/signup/verification")
+		rr := initVerification(s.T(), handler, gin.Param{}, data, userID, "", http.MethodPut, "/api/v1/signup/verification")
 		require.Equal(s.T(), http.StatusBadRequest, rr.Code)
 
 		bodyParams := make(map[string]interface{})
@@ -347,7 +347,7 @@ func (s *TestSignupSuite) TestInitVerificationHandler() {
 	})
 	s.Run("init verification request body could not be read", func() {
 		data := []byte(`{"test_number": "2268213044", "test_code": "1"}`)
-		rr := initVerification(s.T(), handler, gin.Param{}, data, userID, http.MethodPut, "/api/v1/signup/verification")
+		rr := initVerification(s.T(), handler, gin.Param{}, data, userID, "", http.MethodPut, "/api/v1/signup/verification")
 
 		// Check the status code is what we expect.
 		assert.Equal(s.T(), http.StatusBadRequest, rr.Code)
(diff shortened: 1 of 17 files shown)


pkg/signup/service/signup_service.go
```diff
@@ -164,16 +164,27 @@ func EncodeUserID(subject string) string {
...
+			// The UserSignup could not be located by its encoded UserID, attempt to load it using its encoded Username instead
+			encodedUsername := EncodeUserIdentifier(ctx.GetString(context.UsernameKey))
+			userSignup, err = s.CRTClient().V1Alpha1().UserSignups().Get(encodedUsername)
+			if err != nil {
+				if apierrors.IsNotFound(err) {
+					// New Signup
+					log.WithValues(map[string]interface{}{"encoded_user_id": encodedUserID}).Info(ctx, "user not found, creating a new one")
+					return s.createUserSignup(ctx)
+				}
+				return nil, err
```
alexeykazakov, line 183: Can we add a test for this case? When the client return a non 404 error when trying to get the usersignup.
There are actually two sub-cases: an error when trying to obtain the usersignup by userID and by username. Let's cover at least the username case since this is what we are going to keep after the migration is done.

pkg/signup/service/signup_service.go
```diff
@@ -223,15 +234,28 @@ func (s *ServiceImpl) reactivateUserSignup(ctx *gin.Context, existing *toolchain
...
 		if apierrors.IsNotFound(err) {
-			return nil, nil
+			userSignup, err = s.CRTClient().V1Alpha1().UserSignups().Get(EncodeUserIdentifier(username))
+
+			// Otherwise if there was any other error, return it instead
+			if err != nil {
+				if apierrors.IsNotFound(err) {
+					return nil, nil
+				}
+				return nil, err
```
alexeykazakov, line 254: Same here. Let's add a test for this error.

pkg/signup/service/signup_service.go
```diff
@@ -223,15 +234,28 @@ func (s *ServiceImpl) reactivateUserSignup(ctx *gin.Context, existing *toolchain
...
-func (s *ServiceImpl) GetSignup(userID string) (*signup.Signup, error) {
+func (s *ServiceImpl) GetSignup(userID, username string) (*signup.Signup, error) {
+	// Declare variables up front
+	var userSignup *toolchainv1alpha1.UserSignup
+	var err error
 
-	// Retrieve UserSignup resource from the host cluster
-	userSignup, err := s.CRTClient().V1Alpha1().UserSignups().Get(EncodeUserID(userID))
+	// Retrieve UserSignup resource from the host cluster, using the specified UserID
+	userSignup, err = s.CRTClient().V1Alpha1().UserSignups().Get(EncodeUserIdentifier(userID))
```
alexeykazakov, line 243: I wonder if we can just reuse ` GetUserSignup()` here instead? This code which obtains the UserSignup resource looks a lot like a duplication of ` GetUserSignup()`.

alexeykazakov (commented):
Overall looks good! I have a few comments though.

MatousJobanek:
So, do we really want to go with this option and not consider using a label with the encoded username?
My main concern is that if you want to rely on the UserID, then you would need to either keep getting the override claim with the current UserID from SSO forever or migrate all existing UserSignups (which can be risky). And by migrating I mean replacing existing UserSignup with a new one that would use the encoded username as a name. As I mentioned in the other threads - I would love to avoid that!

The label solution doesn't require any of the steps mentioned above. It's much simpler and not risky at all. I don't really understand why you don't want to go with that solution :man_shrugging: Or do you see any potential problem with that?

pkg/signup/service/signup_service.go
sbryzak: Done

pkg/signup/service/signup_service.go
sbryzak: Done

pkg/signup/service/signup_service.go
sbryzak: Done

pkg/signup/service/signup_service_test.go
```diff
@@ -173,6 +175,57 @@ func (s *TestSignupServiceSuite) TestSignup() {
...
+	ctx.Set(context.OriginalSubKey, "original-sub-value")
+	ctx.Set(context.EmailKey, "lsmith@gmail.com")
+	ctx.Set(context.GivenNameKey, "lisa")
+	ctx.Set(context.FamilyNameKey, "smith")
+	ctx.Set(context.CompanyKey, "red hat")
+
+	callCount := 0
+
+	s.FakeUserSignupClient.MockGet = func(id string) (*toolchainv1alpha1.UserSignup, error) {
+		if callCount == 0 {
```
alexeykazakov, line 197: Or you could check the `id` param to return 404 or 500. But the count works too. Just not that obvious.

pkg/signup/service/signup_service_test.go
```diff
@@ -173,6 +175,57 @@ func (s *TestSignupServiceSuite) TestSignup() {
...
+	require.Equal(s.T(), "something bad happened:something very bad happened", err.Error())
+}
+
+func (s *TestSignupServiceSuite) TestGetSignupFailsWithNotFoundThenOtherError() {
+
+	// given
+	callCount := 0
+
+	s.FakeUserSignupClient.MockGet = func(id string) (*toolchainv1alpha1.UserSignup, error) {
+		if callCount == 0 {
```
alexeykazakov, line 216: Same here.

pkg/signup/service/signup_service.go
```diff
@@ -164,16 +164,27 @@ func EncodeUserID(subject string) string {
...
+			if err != nil {
+				if apierrors.IsNotFound(err) {
+					// New Signup
+					log.WithValues(map[string]interface{}{"encoded_user_id": encodedUserID}).Info(ctx, "user not found, creating a new one")
+					return s.createUserSignup(ctx)
+				}
+				return nil, err
+			}
+		} else {
+			return nil, err
```
alexeykazakov, line 186: This error is still not covered by any test.

alexeykazakov (approved):
Looks good!
But there is still one test missing and I also have two very minor comments regarding existing tests. But those are cosmetic. Feel free to ignore them. But please do add the missing test before merging.

pkg/signup/service/signup_service_test.go
sbryzak: Changed here also.

pkg/signup/service/signup_service.go
```diff
@@ -223,12 +234,16 @@ func (s *ServiceImpl) reactivateUserSignup(ctx *gin.Context, existing *toolchain
...
-	userSignup, err := s.CRTClient().V1Alpha1().UserSignups().Get(EncodeUserID(userID))
-	if err != nil {
-		if apierrors.IsNotFound(err) {
+func (s *ServiceImpl) GetSignup(userID, username string) (*signup.Signup, error) {
+	// Declare variables up front
+	var userSignup *toolchainv1alpha1.UserSignup
+	var err error
+
+	// Retrieve UserSignup resource from the host cluster, using the specified UserID and username
+	userSignup, err = s.GetUserSignup(userID, username)
```
MatousJobanek, line 243: The same question here - why do we need to declare them up front?

pkg/signup/service/signup_service.go
```diff
@@ -304,10 +319,26 @@ func (s *ServiceImpl) GetSignup(userID string) (*signup.Signup, error) {
 }
 
 // GetUserSignup is used to return the actual UserSignup resource instance, rather than the Signup DTO
-func (s *ServiceImpl) GetUserSignup(userID string) (*toolchainv1alpha1.UserSignup, error) {
+func (s *ServiceImpl) GetUserSignup(userID, username string) (*toolchainv1alpha1.UserSignup, error) {
+	// Declare variables up front
+	var userSignup *toolchainv1alpha1.UserSignup
+	var err error
+	var err2 error
```
MatousJobanek, line 326: The same

pkg/verification/service/verification_service.go
```diff
@@ -164,7 +164,7 @@ func (s *ServiceImpl) InitVerification(ctx *gin.Context, userID string, e164Phon
 	}
 
 	doUpdate := func() error {
-		signup, err := s.Services().SignupService().GetUserSignup(userID)
+		signup, err := s.Services().SignupService().GetUserSignup(userID, "")
```
MatousJobanek, line 167: why is the username not passed here? 
```suggestion
		signup, err := s.Services().SignupService().GetUserSignup(userID, username)
```

MatousJobanek (changes requested):
I have a few comments/questions.
Apart from that, I noticed that `verification_service_test.go` doesn't have any test for the use-case when the UserSignup is created with the encoded username.

pkg/signup/service/signup_service.go
```diff
@@ -164,16 +164,27 @@ func EncodeUserID(subject string) string {
 // Signup reactivates the deactivated UserSignup resource or creates a new one with the specified username and userID
 // if doesn't exist yet.
 func (s *ServiceImpl) Signup(ctx *gin.Context) (*toolchainv1alpha1.UserSignup, error) {
-	encodedUserID := EncodeUserID(ctx.GetString(context.SubKey))
+	encodedUserID := EncodeUserIdentifier(ctx.GetString(context.SubKey))
+	var userSignup *toolchainv1alpha1.UserSignup
+	var err error
 	// Retrieve UserSignup resource from the host cluster
-	userSignup, err := s.CRTClient().V1Alpha1().UserSignups().Get(encodedUserID)
+	userSignup, err = s.CRTClient().V1Alpha1().UserSignups().Get(encodedUserID)
```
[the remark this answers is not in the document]
sbryzak: You're right, this was an artifact left over from a previous iteration of this code.  Fixed.

pkg/proxy/service/cluster_service.go
```diff
@@ -42,7 +42,7 @@ func NewMemberClusterService(context servicecontext.ServiceContext, options ...O
 
 func (s *ServiceImpl) GetNamespace(ctx *gin.Context, userID string) (*namespace.NamespaceAccess, error) {
 	// Get Signup
-	signup, err := s.ServiceContext.Services().SignupService().GetSignup(userID)
+	signup, err := s.ServiceContext.Services().SignupService().GetSignup(userID, "")
```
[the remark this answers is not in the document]
sbryzak: I just looked through this code and I think you're right.. I'll make the changes required for this, it will mean quite a bit of work to fix the tests here as they're geared towards using just the UserID.

pkg/signup/service/signup_service.go
sbryzak: Fixed

pkg/signup/service/signup_service.go
sbryzak: Fixed

pkg/verification/service/verification_service.go
sbryzak: Good catch, fixed this.

pkg/proxy/service/cluster_service.go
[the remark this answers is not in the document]
sbryzak: This is done now.

pkg/signup/service/signup_service.go
```diff
@@ -223,12 +233,12 @@ func (s *ServiceImpl) reactivateUserSignup(ctx *gin.Context, existing *toolchain
...
-	userSignup, err := s.CRTClient().V1Alpha1().UserSignups().Get(EncodeUserID(userID))
-	if err != nil {
-		if apierrors.IsNotFound(err) {
+func (s *ServiceImpl) GetSignup(userID, username string) (*signup.Signup, error) {
+	// Retrieve UserSignup resource from the host cluster, using the specified UserID and username
+	userSignup, err := s.GetUserSignup(userID, username)
+	// If an error was returned, *or* the returned userSignup is nil, then return here
+	if err != nil || userSignup == nil {
+		if err != nil && apierrors.IsNotFound(err) {
 			return nil, nil
```
MatousJobanek, line 242: When could happen that the `userSignup` is `nil` and the `err` is not set (is nil as well)? This code is pretty confusing and error-prone.

pkg/signup/service/signup_service.go
```diff
@@ -342,7 +363,7 @@ func (s *ServiceImpl) PhoneNumberAlreadyInUse(userID, phoneNumberOrHash string)
 		return errors.NewInternalError(err, "failed listing userSignups")
 	}
 	for _, signup := range userSignupList.Items {
-		if signup.Spec.Userid != userID && !states.Deactivated(&signup) { // nolint:gosec
+		if signup.Spec.Userid != userID && signup.Spec.Username != username && !states.Deactivated(&signup) { // nolint:gosec
```
MatousJobanek, line 366: just out of curiosity - why do we need to verify also the `username`? wouldn't be the `userID` enough?

MatousJobanek (commented):
Looks good :+1: Thanks a lot for addressing my comments :pray: 
I have one more request - could you please add a unit test that would verify the usage of the username also in the proxy test [pkg/proxy/service/cluster_service_test.go](https://github.com/codeready-toolchain/registration-service/pull/251/files#diff-3fa75d73df72f66f7cbd972ef50b2755a053324b7105cd1b506381567b57c2a0)?

pkg/signup/service/signup_service.go
sbryzak: Rewritten to be less confusing.

pkg/signup/service/signup_service.go
sbryzak: What if the userID is empty though?  I think this is just safer to check both.

pkg/controller/signup_test.go
```diff
@@ -651,21 +651,72 @@ func (s *TestSignupSuite) TestVerifyCodeHandler() {
...
+
+		updatedUserSignup, err := s.FakeUserSignupClient.Get(otherUserSignup.Name)
+		require.NoError(s.T(), err)
+
+		// Check that the correct UserSignup is passed into the FakeSignupService for update
+		require.False(s.T(), states.VerificationRequired(updatedUserSignup))
+		require.Empty(s.T(), updatedUserSignup.Annotations[crtapi.UserVerificationAttemptsAnnotationKey])
+		require.Empty(s.T(), updatedUserSignup.Annotations[crtapi.UserSignupVerificationCodeAnnotationKey])
+		require.Empty(s.T(), updatedUserSignup.Annotations[crtapi.UserVerificationExpiryAnnotationKey])
+
```
alexeykazakov, line 707:
```suggestion
```

pkg/controller/signup_test.go
```diff
@@ -651,21 +651,71 @@ func (s *TestSignupSuite) TestVerifyCodeHandler() {
...
+		// Create Signup controller instance.
+		ctrl := controller.NewSignup(s.Application)
+		handler := gin.HandlerFunc(ctrl.VerifyCodeHandler)
+
+		param := gin.Param{
+			Key:   "code",
+			Value: "999127",
+		}
+		rr := initVerification(s.T(), handler, param, nil, otherUserSignup.Spec.Userid, otherUserSignup.Spec.Username, http.MethodGet, "/api/v1/signup/verification")
```
xcoulon, line 694: AFAICS, this test uses both valid `UserID` and `Username` args, so how do we know that the resource look-up was based on the latter and not the former?

pkg/signup/service/signup_service.go
```diff
@@ -223,17 +233,22 @@ func (s *ServiceImpl) reactivateUserSignup(ctx *gin.Context, existing *toolchain
...
-func (s *ServiceImpl) GetSignup(userID string) (*signup.Signup, error) {
-
-	// Retrieve UserSignup resource from the host cluster
-	userSignup, err := s.CRTClient().V1Alpha1().UserSignups().Get(EncodeUserID(userID))
+func (s *ServiceImpl) GetSignup(userID, username string) (*signup.Signup, error) {
+	// Retrieve UserSignup resource from the host cluster, using the specified UserID and username
+	userSignup, err := s.GetUserSignup(userID, username)
+	// If an error was returned, then return here
 	if err != nil {
 		if apierrors.IsNotFound(err) {
```
xcoulon, line 241: if the `UserSignup` is not found when looking-up by `userID`, then we shouldn't return, but instead we should look-up by `username`, shouldn't we?

pkg/verification/service/verification_service_test.go
```diff
@@ -159,6 +186,40 @@ func (s *TestVerificationServiceSuite) TestInitVerification() {
...
+	require.NoError(s.T(), err)
+
+	userSignup2, err = s.FakeUserSignupClient.Get(userSignup2.Name)
+	require.NoError(s.T(), err)
+
+	// Ensure the verification code is set
+	require.NotEmpty(s.T(), userSignup2.Annotations[toolchainv1alpha1.UserSignupVerificationCodeAnnotationKey])
+
+	buf = new(bytes.Buffer)
+	_, err = buf.ReadFrom(reqBody)
```
xcoulon, line 212: don't you need to call `reqBody.Close()` in a deferred func?

MatousJobanek (approved):
Looks good :+1: Thanks for addressing my comments.
Please, take a look at @xcoulon's review https://github.com/codeready-toolchain/registration-service/pull/251#pullrequestreview-901505284 as well

pkg/controller/signup_test.go
sbryzak: Good point, modified to just pass the username.

pkg/verification/service/verification_service_test.go
sbryzak: Not sure if it matters in a unit test, but I've added deferred `Close()` calls anyway.

pkg/signup/service/signup_service.go
sbryzak: This is all handled in the `GetUserSignup()` function - if the lookup by `userID` fails, then it attempts to find the `UserSignup` again using the `username` instead.

xcoulon: ah my bad, I missread the parameters in the `GetUserSignup` method 🤦‍♂️

PR #254, opened 2022-03-10
Create new UserSignups using the encoded username instead of userID

sbryzak:
This PR modifies registration service to create new UserSignup instances using the user's username instead of userID.

Fixes https://issues.redhat.com/browse/CRT-1401

WIP

### DIFF
--- a/pkg/signup/service/signup_service_test.go
+++ b/pkg/signup/service/signup_service_test.go
@@ -60,7 +60,7 @@ func (s *TestSignupServiceSuite) ServiceConfiguration(namespace string, verifica
 func (s *TestSignupServiceSuite) TestSignup() {
 	s.ServiceConfiguration(TestNamespace, true, "", 5)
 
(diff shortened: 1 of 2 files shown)


pkg/signup/service/signup_service_test.go
```diff
@@ -76,9 +76,8 @@ func (s *TestSignupServiceSuite) TestSignup() {
 
 		val := userSignups.Items[0]
 		require.Equal(s.T(), configuration.Namespace(), val.Namespace)
-		require.Equal(s.T(), userID, val.Name)
-		require.Equal(s.T(), userID, val.Spec.Userid)
```
MatousJobanek, line 80: we should verify that the `Spec.UserID` is correctly set as well

MatousJobanek (commented):
Looks good, waiting for the e2e tests

pkg/signup/service/signup_service_test.go
sbryzak: No problem, done.

alexeykazakov (approved):

PR #264, opened 2022-05-09
refactor(err): insert space between message and details

xcoulon:
also, no need for a `:` suffix character when error has no details

### DIFF
--- a/pkg/signup/service/signup_service_test.go
+++ b/pkg/signup/service/signup_service_test.go
@@ -198,8 +198,7 @@ func (s *TestSignupServiceSuite) TestSignupFailsWhenClientReturnsError() {
 
 	// when
 	_, err = s.Application.SignupService().Signup(ctx)
-	require.Error(s.T(), err)
-	require.Equal(s.T(), "an internal error:an internal error happened", err.Error())
+	require.EqualError(s.T(), err, "an internal error: an internal error happened", err.Error())
 }
 
 func (s *TestSignupServiceSuite) TestSignupFailsWithNotFoundThenOtherError() {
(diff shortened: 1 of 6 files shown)


pkg/signup/service/signup_service_test.go
```diff
@@ -198,8 +198,7 @@ func (s *TestSignupServiceSuite) TestSignupFailsWhenClientReturnsError() {
 
 	// when
 	_, err = s.Application.SignupService().Signup(ctx)
-	require.Error(s.T(), err)
-	require.Equal(s.T(), "an internal error:an internal error happened", err.Error())
+	require.EqualError(s.T(), err, "an internal error: an internal error happened", err.Error())
```
rajivnathan, line 201: `require.Error(s.T(), err)` was there to handle when `err` was `nil`?

rajivnathan (commented):
Just 1 question about the removal of `require.Error(s.T(), err)`

pkg/signup/service/signup_service_test.go
MatousJobanek: good point - when the error is nil, then it would panic in this case. It should have been just:
```suggestion
	require.EqualError(s.T(), err, "an internal error: an internal error happened")
```

xcoulon: ah right, I made those changes using a search/replace but I missed the last `err.Error()` argument in some case. I'll fix that

MatousJobanek (approved):
Looks good, just drop the last parameter from:
```go
require.EqualError(s.T(), err, "...", err.Error())
```
so it doesn't panic when the err is nil.

pkg/signup/service/signup_service_test.go
xcoulon: done in 90d0d39

PR #267, opened 2022-06-14
feat: activation code endpoint and service

xcoulon:
checks that the activation code is valid, ie:
- there is a `SocialEvent` resource named after the code (same name)
- there are enough seats available
- the start and end dates are valid

if one of the condition above is not met, an error is returned
and the user's attempts counter is incremented, until a max
is reached, after which no other attempt will be permitted.

If all good, then also label the UserSignup resource with the name
of the SocialEvent (ie, the activation code)

Note: updating the status of the `SocialEvent` resource when a user passed
the verification (ie, incrementing the `status.ActivationCount` is the responsibility
of the Host Operator and will be provided in a subsequent PR 
(because pairing multiple PRs with e2e tests is not supported)

See also:
- https://github.com/codeready-toolchain/api/pull/315
- https://github.com/codeready-toolchain/toolchain-common/pull/261
- https://github.com/codeready-toolchain/toolchain-e2e/pull/575

Fixes https://issues.redhat.com/browse/CRT-1386

### DIFF
--- a/pkg/verification/service/verification_service.go
+++ b/pkg/verification/service/verification_service.go
@@ -395,16 +395,17 @@ func (s *ServiceImpl) VerifyActivationCode(ctx *gin.Context, userID, username, c
 	now := metav1.NewTime(time.Now())
 	log.Infof(ctx, "verifying activation code '%s': event.Status.ActivationCount=%d, event.Spec.MaxAttendees=%d, event.Spec.StartTime=%s, event.Spec.EndTime=%s", code, strconv.Itoa(event.Status.ActivationCount), strconv.Itoa(event.Spec.MaxAttendees), event.Spec.StartTime.Format("2006-01-02:03:04:05"), event.Spec.EndTime.Format("2006-01-02:03:04:05"))
 
-	if event.Status.ActivationCount < event.Spec.MaxAttendees &&
-		event.Spec.StartTime.Before(&now) &&
-		event.Spec.EndTime.After(now.Time) {
-		log.Infof(ctx, "approving user signup request with activation code '%s'", code)
-		// If the activation code is acceptable then set `VerificationRequired` state to false and reset other verification annotations
-		unsetVerificationRequired = true
-		annotationsToDelete = append(annotationsToDelete, toolchainv1alpha1.UserVerificationAttemptsAnnotationKey)
-		return nil
+	if event.Status.ActivationCount >= event.Spec.MaxAttendees {
+		return crterrors.NewForbiddenError("invalid code", "the event is full")
+	} else if event.Spec.StartTime.After(now.Time) || event.Spec.EndTime.Before(&now) {
+		log.Infof(ctx, "the event with code '%s' has not started yet or is already past", code)
+		return crterrors.NewForbiddenError("invalid code", "the provided code is invalid")
 	}
-	return crterrors.NewForbiddenError("invalid code", "the provided code is invalid")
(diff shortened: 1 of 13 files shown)


pkg/verification/service/verification_service.go
```diff
@@ -322,6 +323,109 @@ func (s *ServiceImpl) VerifyPhoneCode(ctx *gin.Context, userID, username, code s
...
+			// The code doesn't match
+			return crterrors.NewForbiddenError("invalid code", "the provided code is invalid")
+		}
+		return crterrors.NewInternalError(err, fmt.Sprintf("error retrieving event '%s'", code))
+	}
+	// if there is room for the user and if the "time window" to signup is valid
+	now := metav1.NewTime(time.Now())
+	log.Infof(ctx, "verifying activation code '%s': event.Status.ActivationCount=%d, event.Spec.MaxAttendees=%d, event.Spec.StartTime=%s, event.Spec.EndTime=%s", code, strconv.Itoa(event.Status.ActivationCount), strconv.Itoa(event.Spec.MaxAttendees), event.Spec.StartTime.Format("2006-01-02:03:04:05"), event.Spec.EndTime.Format("2006-01-02:03:04:05"))
+
+	if event.Status.ActivationCount < event.Spec.MaxAttendees &&
```
rajivnathan, line 398: What updates the `ActivationCount`? Is it the host operator?

pkg/verification/service/verification_service.go
```diff
@@ -322,6 +323,109 @@ func (s *ServiceImpl) VerifyPhoneCode(ctx *gin.Context, userID, username, code s
...
+		event.Spec.StartTime.Before(&now) &&
+		event.Spec.EndTime.After(now.Time) {
+		log.Infof(ctx, "approving user signup request with activation code '%s'", code)
+		// If the activation code is acceptable then set `VerificationRequired` state to false and reset other verification annotations
+		unsetVerificationRequired = true
+		annotationsToDelete = append(annotationsToDelete, toolchainv1alpha1.UserVerificationAttemptsAnnotationKey)
+		return nil
+	}
+	return crterrors.NewForbiddenError("invalid code", "the provided code is invalid")
```
rajivnathan, line 407: Should we maybe return a different error if the event is full? I think that would help event runners understand why the activation is failing in that case?

rajivnathan (commented):
Looks good! 👍 Just some minor comments but also a question about how we can better help event organizers understand why a user's activation might be failing

pkg/verification/service/verification_service.go
xcoulon: yes, that's the `SocialEventController` which takes care of updating the status. This will be addressed in a subsequent PR (because a couple of PRs for both registration-service and host-operator at the same time would not work on OpenShift-CI)

pkg/verification/service/verification_service.go
xcoulon: yeah, that would make sense, indeed.

xcoulon: but I would keep the same default error message (ie, `the provided code is invalid`) in case the current time does not match with the event range

xcoulon: done in b9101fd. The error message is still `the provided code is invalid` if the time is invalid, but we'll have a descriptive message in the logs, in case we need to provide support to users.

rajivnathan (approved):
This will be really nice for workshops! 👍

PR #275, opened 2022-07-21
Proxy cache

alexeykazakov:
This PR introduces a cache for the proxy.

Without such a cache the proxy generates 6 requests to the api servers for every client request. Five of them is for retrieving user information to determine where the user request should be forwarded to and what token to use. With this cache the proxy will generate only two requests in most cases. A quick one to verify the cached token using a special kube authentication endpoint (via TokenReview pseudo resource) and the second one is the original forwarded user request.
The proxy does a full round only if nothing is found in cache yet or if the token is not valid anymore.

JIRA: https://issues.redhat.com/browse/CRT-1373

Paired e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/583


pkg/proxy/cache.go
```diff
@@ -1,22 +1,78 @@
...
+func (c *UserNamespaces) namespaceFromCache(ctx *gin.Context, username string) (*namespace.NamespaceAccess, error) {
+	c.RLock()
+	defer c.RUnlock()
+	na, ok := c.namespaces[username]
+	if ok {
+		valid, err := na.Validate()
+		if err != nil {
+			return nil, err
+		}
+		if valid {
```
xcoulon, line 63: if `na` not valid, shouldn't you remove it from the cache?

pkg/proxy/namespace/namespace.go
```diff
@@ -1,11 +1,53 @@
...
+
+	authenticationv1 "k8s.io/api/authentication/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
 
 // NamespaceAccess holds an information needed to access the namespace in a member cluster for the specific user
 type NamespaceAccess struct { // nolint:revive
 	// APIURL is the Cluster API Endpoint for the namespace
-	APIURL url.URL
+	apiURL url.URL
```
xcoulon, line 17: what's the benefit of using private field along with a getter? do you need to protect the write-access to these fields?

xcoulon (approved):
Looks good, just a few minor comments

pkg/proxy/namespace/namespace.go
alexeykazakov: Yes. Wanted to protect the write-access.

pkg/proxy/cache.go
alexeykazakov: Well... we could but if the `na` is not valid than we get a new one from the cluster and put it in the cache. Removing it from cache, retrieving a new one and then putting back could actually decrease performance in case of multiple parallel requests from the same user. One go routine may be removing the invalid na which was just updated by another. It wouldn't cause any errors or something like this. Eventually the cache would be properly updated but my point is that it's probably doesn't give us any performance benefits if we remove it before updating.

xcoulon: sure, the invalid entry will be replaced by a valid one shortly after. Except maybe when users disconnect from the platform, in which case the cache will retain the entries indefinitely?

pkg/proxy/namespace/namespace.go
xcoulon: ok

pkg/proxy/cache.go
xcoulon: that being said, clearing outdated entries in the cache could be addressed later

alexeykazakov: Not indefinitely. Until the next pod restart :)

xcoulon: actually, it's a separate use-case 🤦‍♂️

alexeykazakov: And it wouldn't actually help much with outdated cache entries for disconnected users anyway. We check if the na is valid only if there is a request from a client/user. It's unlikely that a user which was removed would keep using the proxy.

xcoulon: yes, which is why I wrote it's a separate use-case: the code above is only going to be called by active users! ;)

PR #278, opened 2022-09-21
Added code to support AWS SNS

sbryzak:
This PR adds support for using AWS SNS to send verification notification messages instead of Twilio.

Fixes: https://issues.redhat.com/browse/CRT-1588


pkg/verification/service/verification_service.go
```diff
@@ -149,15 +153,10 @@ func (s *ServiceImpl) InitVerification(ctx *gin.Context, userID, username, e164P
...
-		msg, err := client.Messages.SendMessage(fromNumber, e164PhoneNumber, content, nil)
+
+		err = s.NotificationService.SendNotification(ctx, content, e164PhoneNumber)
 		if err != nil {
-			if msg != nil {
-				log.Error(ctx, err, fmt.Sprintf("error while sending, code: %d message: %s", msg.ErrorCode, msg.ErrorMessage))
-			} else {
-				log.Error(ctx, err, "unknown error while sending")
-			}
+			log.Error(ctx, err, "unknown error while sending")
```
rajivnathan, line 159:
```suggestion
			log.Error(ctx, err, "error while sending notification")
```

pkg/verification/sender/amazon_sns_sender.go
```diff
@@ -0,0 +1,62 @@
...
+			"AWS.SNS.SMS.SenderID": senderId,
+			"AWS.SNS.SMS.SMSType":  smsType,
+		},
+	})
+
+	if err != nil {
+		return err
+	}
+
+	log.Info(ctx, fmt.Sprintf("Notification Message Sent.  Message ID: [%s] Phone Number: [%s]",
```
rajivnathan, line 58: Not sure if this was just for testing purposes but I believe we cannot log the phone number?

pkg/verification/sender/twilio_sender.go
```diff
@@ -0,0 +1,42 @@
...
+	msg, err := client.Messages.SendMessage(fromNumber, phoneNumber, content, nil)
+	if err != nil {
+		if msg != nil {
+			log.Error(ctx, err, fmt.Sprintf("error while sending, code: %d message: %s", msg.ErrorCode, msg.ErrorMessage))
+		} else {
+			log.Error(ctx, err, "unknown error while sending")
+		}
+
+		// If we get an error here then just die, don't bother updating the UserSignup
+		return crterrors.NewInternalError(err, "error while sending verification code")
```
rajivnathan, line 38: We can just wrap the error and return it since it'll be logged by the caller.

pkg/verification/service/verification_service.go
```diff
@@ -149,15 +153,10 @@ func (s *ServiceImpl) InitVerification(ctx *gin.Context, userID, username, e164P
...
 
 			// If we get an error here then just die, don't bother updating the UserSignup
 			return crterrors.NewInternalError(err, "error while sending verification code")
```
rajivnathan, line 162: If we get past this point should we have an INFO level message to indicate sending the verification code was sent successfully?

rajivnathan (commented):
Looks good, just a couple suggestions

pkg/verification/service/verification_service.go
sbryzak: Done

pkg/verification/sender/amazon_sns_sender.go
sbryzak: Good catch, removed.

pkg/verification/sender/twilio_sender.go
sbryzak: Fixed.

pkg/verification/service/verification_service.go
sbryzak: I don't think we need it, we seemingly didn't have any logging like this before.

rajivnathan (commented):
Looks good! Think we're just missing some unit tests for the configuration.go, verification_service.go and sender.go changes

sbryzak:
@rajivnathan there are no unit tests for the AWS notification sender, however I can confirm that the new code has been manually tested and works.

pkg/verification/service/verification_service_test.go
```diff
@@ -12,6 +12,8 @@ import (
 	"testing"
 	"time"
 
+	sender2 "github.com/codeready-toolchain/registration-service/pkg/verification/sender"
```
rajivnathan, line 15: nit: think it'd be better to rename this to vsender or something

pkg/configuration/configuration_test.go
```diff
@@ -78,12 +78,21 @@ func TestRegistrationService(t *testing.T) {
 			Verification().MessageTemplate("Developer Sandbox verification code: %s").
 			Verification().ExcludedEmailDomains("redhat.com,ibm.com").
 			Verification().CodeExpiresInMin(151).
-			Verification().Secret().Ref("verification-secrets").TwilioAccountSID("twilio.sid").TwilioAuthToken("twilio.token").TwilioFromNumber("twilio.fromnumber"))
+			Verification().AWSRegion("us-west-2").
+			Verification().AWSSenderID("sandbox").
```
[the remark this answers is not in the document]
sbryzak: I've added this too.

pkg/verification/service/verification_service_test.go
sbryzak: renamed to senderpkg

rajivnathan (approved):
Thanks for addressing all the comments

PR #289, opened 2022-11-30
Update proxy cluster access model

rajivnathan:
Proxy will no longer use service account lookup, instead it will use impersonation.

Related Issue: https://issues.redhat.com/browse/CRT-1461
Related PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/628


pkg/proxy/cache.go
```diff
@@ -6,73 +6,62 @@ import (
...
-// Get tries to retrieve the namespace access from the cache first. If found then it checks if the cached access is still valid.
-// If not found or invalid then retrieves a new namespace access from the member cluster service and stores it in the cache.
-func (c *UserNamespaces) Get(ctx *gin.Context, userID, username string) (*namespace.NamespaceAccess, error) {
-	na, err := c.namespaceFromCache(ctx, username)
-	if err != nil {
-		return nil, err
-	}
-	if na != nil {
-		return na, nil
+// Get tries to retrieve the cluster access from the cache first. If found then it checks if the cached access is still valid.
```
alexeykazakov, line 29: It doesn't validate the access anymore. Let's fix the comment.

pkg/proxy/cache.go
```diff
@@ -6,73 +6,62 @@ import (
...
+	c.add(username, ca)
 
-	return na, nil
+	return ca, nil
 }
 
-// namespaceFromCache tries to retrieve a namespace access from the cache and validates it.
-// Returns nil, nil if no namespace access found or if it's invalid.
-func (c *UserNamespaces) namespaceFromCache(ctx *gin.Context, username string) (*namespace.NamespaceAccess, error) {
+// clusterAccessFromCache tries to retrieve a cluster access from the cache and validates it.
```
alexeykazakov, line 49: Same here. No validation anymore.

pkg/proxy/cache.go
```diff
@@ -6,73 +6,62 @@ import (
...
+// Returns nil if no cluster access found.
+func (c *UserAccess) clusterAccessFromCache(ctx *gin.Context, username string) *access.ClusterAccess {
 	c.RLock()
 	defer c.RUnlock()
-	na, ok := c.namespaces[username]
-	if ok {
-		valid, err := na.Validate()
```
alexeykazakov, line 59: This is what you are drooping in terms of validation ^^^

pkg/proxy/cache.go
```diff
@@ -6,73 +6,62 @@ import (
...
-		log.Info(ctx, fmt.Sprintf("Namespace Access found in cache for user '%s' but it's not valid anymore", username))
-		return nil, nil
+	ca, ok := c.clusterAccesses[username]
+	if !ok {
+		log.Info(ctx, fmt.Sprintf("Cluster Access NOT found in cache for user : %s", username))
+		return nil
 	}
-	log.Info(ctx, fmt.Sprintf("Namespace Access NOT found in cache for user : %s", username))
-	return nil, nil
+	log.Info(ctx, fmt.Sprintf("A valid Cluster Access found in cache for user '%s'", username))
```
alexeykazakov, line 59:
```suggestion
	log.Info(ctx, fmt.Sprintf("A Cluster Access found in cache for user '%s'", username))
```

pkg/proxy/access/clusteraccess.go
```diff
@@ -0,0 +1,40 @@
...
+	"net/url"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ClusterAccess holds an information needed to access the namespace in a member cluster for the specific user
+type ClusterAccess struct { // nolint:revive
+	// APIURL is the Cluster API Endpoint for the namespace
+	apiURL url.URL
+	// SAToken is a token of the Service Account which represents the user in the namespace
```
alexeykazakov, line 13: It's not a SA token representing the user anymore. It's a member client SA now.
Maybe let's rename it from `saToken` to `impersonatorToken`?

pkg/proxy/access/clusteraccess.go
```diff
@@ -0,0 +1,40 @@
+package access
+
+import (
+	"net/url"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ClusterAccess holds an information needed to access the namespace in a member cluster for the specific user
```
alexeykazakov, line 9: This CA is designed to be used for impersonation only. Let's mention it in the comment.

alexeykazakov (approved):
Looks good! I left a few minor comments to address though.

pkg/proxy/cache.go
rajivnathan: Good catch, done in d36401dd79ba7c2e9467cd07278381699783fd21

pkg/proxy/cache.go
rajivnathan: Good catch, done in d36401dd79ba7c2e9467cd07278381699783fd21

pkg/proxy/access/clusteraccess.go
rajivnathan: Updated the comment in 967fdab265e038a8df86bf12da73dce210e478bb. Let me know if you would like to add/change anything else.

alexeykazakov: Looks good! Thanks.

alexeykazakov (approved):
Thank you for addressing my comments.
Great work!

pkg/proxy/access/clusteraccess.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+// ClusterAccess holds information needed to access user namespaces in a member cluster for the specific user via impersonation
+type ClusterAccess struct { // nolint:revive
+	// APIURL is the Cluster API Endpoint for the namespace
+	apiURL url.URL
+	// impersonatorToken is a token of the impersonator's Service Account, usually the impersonator SA belongs to a member toolchaincluster
+	impersonatorToken string
+	// client is the kube client for the cluster.
+	client client.Client
```
MatousJobanek, line 16: The client doesn't seem to be used

pkg/proxy/access/clusteraccess.go
```diff
@@ -0,0 +1,40 @@
...
+	return a.apiURL
+}
+
+func (a *ClusterAccess) ImpersonatorToken() string {
+	return a.impersonatorToken
+}
+
+func (a *ClusterAccess) Username() string {
+	return a.username
+}
```
MatousJobanek, line 40: you could easily export the parameters and drop these "getter" functions.

pkg/proxy/service/cluster_service.go
```diff
@@ -43,7 +33,7 @@ func NewMemberClusterService(context servicecontext.ServiceContext, options ...O
 	return si
 }
 
-func (s *ServiceImpl) GetNamespace(ctx *gin.Context, userID, username string) (*namespace.NamespaceAccess, error) {
+func (s *ServiceImpl) GetClusterAccess(ctx *gin.Context, userID, username string) (*access.ClusterAccess, error) {
 	// Get Signup
 	signup, err := s.ServiceContext.Services().SignupService().GetSignup(userID, username)
```
MatousJobanek, line 38: it's probably fine for now, but keep in mind that the `GetSignup` function sets the cluster-name based on the value that is in MUR, but we should ideally take it from the Space CR.

MatousJobanek (approved):
just minor comments

pkg/proxy/access/clusteraccess.go
rajivnathan: We could but then we'd also open them up to being modifiable outside of this package which I'm not sure we need/want? But we could change it if we want later

pkg/proxy/service/cluster_service.go
rajivnathan: I guess we'll need to change the cluster lookup logic anyway so I think we can change it then?

pkg/proxy/access/clusteraccess.go
rajivnathan: Good point, removed in ce1d76fe6af282dc57a9caaed5a237f7e32b54df

PR #293, opened 2022-12-15
Fix username

rajivnathan:
CompliantUsername should be used for impersonation since the username in the RoleBinding in the user namespace is the CompliantUsername.

```
subjects:
  - kind: User
    apiGroup: rbac.authorization.k8s.io
    name: compliantusername
```

### DIFF
--- a/pkg/proxy/proxy_test.go
+++ b/pkg/proxy/proxy_test.go
@@ -12,7 +12,7 @@ import (
 	"testing"
 	"time"
 
-	service2 "github.com/codeready-toolchain/registration-service/pkg/application/service"
+	appservice "github.com/codeready-toolchain/registration-service/pkg/application/service"
 	"github.com/codeready-toolchain/registration-service/pkg/auth"
 	"github.com/codeready-toolchain/registration-service/pkg/proxy/access"
 	"github.com/codeready-toolchain/registration-service/pkg/proxy/service"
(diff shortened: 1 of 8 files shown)


alexeykazakov (commented):
Looks good. Just need to add some tests.

pkg/proxy/proxy_test.go
```diff
@@ -9,22 +9,25 @@ import (
 	"log"
 	"net/http"
 	"net/http/httptest"
-	"net/url"
 	"testing"
 	"time"
 
-	"github.com/codeready-toolchain/registration-service/pkg/application/service"
+	service2 "github.com/codeready-toolchain/registration-service/pkg/application/service"
```
xcoulon, line 15: `appservice` and `proxyservice` would be better aliases, but it's not a blocker

xcoulon (approved):

pkg/proxy/proxy_test.go
MatousJobanek: I'll update it in following PR - to not lose the 20 minutes of running the e2e tests :innocent:

MatousJobanek:
> Looks good. Just need to add some tests.

added

pkg/proxy/proxy_test.go
MatousJobanek: anyway, I figured that I missed a header in two test cases, and since the review meeting already started then I updated the alias right away:  1149fa4

alexeykazakov (approved):

rajivnathan:
/retest
infra issue

PR #297, opened 2023-01-12
Informers only

rajivnathan:
Related Issue: https://issues.redhat.com/browse/CRT-1619

Related PR: https://github.com/codeready-toolchain/host-operator/pull/726

**High level summary of changes**
- Removes the old proxy cache mechanism in favour of using [Shared Informers](https://github.com/kubernetes/client-go/blob/master/tools/cache/shared_informer.go). Cache invalidation is no longer necessary since the informers will update to the latest versions of resources.
- Wraps the informers in an Informer Service that is accessible application-wide so it can be easily reused for quick retrieval of resources
- Signup Service includes a`GetSignupFromInformer` function that uses informers to get the Signup DTO. The original GetSignup function still uses CRTClient to get resources since we haven't used informers directly in this way I figured it would be best to have a transition period where regular registration service functionality uses the CRTClient way and the proxy uses the informer way. If the informer way proves to be stable then we can remove the old way and use only the Informer way.


alexeykazakov (approved):
Nice!

pkg/informers/service/informer_service_test.go
```diff
@@ -0,0 +1,245 @@
...
+								{
+									"targetCluster": "member1",
+								},
+							},
+						},
+					},
+				},
+			},
+		}
+
```
MatousJobanek, line 56: you could also add another MUR with a different name as an additional noise

pkg/informers/service/informer_service_test.go
```diff
@@ -0,0 +1,245 @@
...
+						},
+					},
+				},
+			}
+
+			val, err := svc.GetMasterUserRecord("johnMur")
+			require.NotNil(s.T(), val)
+			require.NoError(s.T(), err)
+			assert.Equal(s.T(), val, expected)
+		})
```
MatousJobanek, line 89: could you please add `// given` `// when` `// then` comments to all tests? it helps with reading them
```suggestion
		s.Run("found", func() {
			// given
			expected := &toolchainv1alpha1.MasterUserRecord{
				Spec: toolchainv1alpha1.MasterUserRecordSpec{
					TierName: "deactivate30",
					UserID:   "john-id",
					UserAccounts: []toolchainv1alpha1.UserAccountEmbedded{
						{
							TargetCluster: "member1",
						},
					},
				},
			}

			// when
			val, err := svc.GetMasterUserRecord("johnMur")

			// then
			require.NotNil(s.T(), val)
			require.NoError(s.T(), err)
			assert.Equal(s.T(), val, expected)
		})
```

pkg/kubeclient/toolchainstatus.go
```diff
@@ -6,6 +6,8 @@ import (
 	crtapi "github.com/codeready-toolchain/api/api/v1alpha1"
 )
 
+const ToolchainStatusName = "toolchain-status"
+
 const (
 	toolchainStatusResourcePlural = "toolchainstatuses"
 )
```
MatousJobanek, line 13: this can be removed & replaced by the `ToolchainStatusPlural` constant from the new `resources.go`

MatousJobanek (approved):
Looks great :+1: 
I just have a feeling that if we would get rid of the service abstraction then we could make the code much simpler cc @sbryzak 

BTW, correct me if I'm wrong - we actually don't use any part of the informer implementation in the proxy test, right? We only use the mock.

pkg/kubeclient/toolchainstatus.go
rajivnathan: good catch! done in 67e4d1e7f601312db32f160ea3ba93cc35250a64

pkg/informers/service/informer_service_test.go
rajivnathan: done in d28349e0f8b60e6f3ffef74e239af1f56df15a3e

pkg/informers/service/informer_service_test.go
rajivnathan: done in d28349e0f8b60e6f3ffef74e239af1f56df15a3e

rajivnathan:
@MatousJobanek 
>BTW, correct me if I'm wrong - we actually don't use any part of the informer implementation in the proxy test, right? We only use the mock.
If you're talking about proxy_test.go then that's correct. Initially when I started writing the tests I created a fake informer, but then I realized the proxy logic doesn't really care about that. It just uses the cluster service. Even cluster service shouldn't really care about how the resources are retrieved specifically, it just calls a function with a userID and username and expects the `signup.Signup` object that contains the required information it needs.

That said, this made me realize that there's a test gap in the signup service because it's only testing the `GetSignup` function, not the `GetSignupFromInformer`. I should update the signup_service_test.go so that it tests with a fake informer service.

rajivnathan:
@MatousJobanek added unit tests for GetSignupFromInformer using a fake informer service in f739474a8f80883b80d49b9cd822d61c7eeb7c0b

PR #301, opened 2023-02-07
Space lister

rajivnathan:
Implements the proxy space lister API as described in https://issues.redhat.com/browse/CRT-1641

Also replaces the previous proxy http router with https://github.com/labstack/echo

Related PRs:
https://github.com/codeready-toolchain/api/pull/337
https://github.com/codeready-toolchain/member-operator/pull/408
https://github.com/codeready-toolchain/toolchain-e2e/pull/661
https://github.com/codeready-toolchain/toolchain-common/pull/290
https://github.com/codeready-toolchain/host-operator/pull/742


pkg/proxy/handlers/spacelister.go
```diff
@@ -0,0 +1,157 @@
...
+	ctx.Response().Writer.Header().Set("Content-Type", "application/json")
+	ctx.Response().Writer.WriteHeader(http.StatusOK)
+	return json.NewEncoder(ctx.Response().Writer).Encode(workspaceList)
+}
+
+func (s *SpaceLister) workspacesFromSpaceBindings(spaceBindings []*toolchainv1alpha1.SpaceBinding) ([]toolchainv1alpha1.Workspace, error) {
+	workspaces := []toolchainv1alpha1.Workspace{}
+	for _, spaceBinding := range spaceBindings {
+
+		if spaceBinding.Labels == nil {
```
alexeykazakov, line 120: Labels can be nil? In case it's not fully initialized yet? Worth adding a comment and maybe log it..

pkg/proxy/handlers/spacelister_test.go
```diff
@@ -0,0 +1,102 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
+	"github.com/labstack/echo/v4"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Work in progress
```
alexeykazakov, line 24: These tests do look like WIP :)
Let's add more tests to cover:
- cases when there are multiple space bindings (shared spaces) including some noise (spaces and space bindings belonging to other users)
- error handling

alexeykazakov (commented):
Looks good! But need more tests :)

pkg/proxy/handlers/spacelister.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+
+		workspace := commonproxy.NewWorkspace(spaceName,
+			commonproxy.WithNamespaces(namespaces),
+			commonproxy.WithOwner(ownerName),
+			commonproxy.WithRole(spaceBinding.Spec.SpaceRole),
+		)
+		workspaces = append(workspaces, *workspace)
+	}
+	return workspaces, nil
```
mfrancisc, line 156: Just curious - should this function list also sub-spaces? 
since those will not have their own spacebindings , but they inherit them from the parentSpace.

In case sub-space retrieval is needed it can be done using following label:
`toolchainv1alpha1.ParentSpaceLabelKey: spaceName`

rajivnathan: It might eventually but it was discussed in the design doc linked in https://issues.redhat.com/browse/CRT-1641 and the decision was to not include sub-spaces (at least at this point in time).

pkg/proxy/handlers/spacelister.go
rajivnathan: Not sure, figured I'd just cover it just in case (eg. if a SpaceBinding is created manually or something).

But actually now that I think of it it's redundant because `spaceName` would just be empty a few lines below and we would log the error when we try to find the space so it's covered. I removed this check in https://github.com/codeready-toolchain/registration-service/pull/301/commits/a1734586fc0770f0c714eb1a264d06ff9b3b349f
Thanks!

pkg/proxy/handlers/spacelister.go
```diff
@@ -0,0 +1,155 @@
...
+func (s *SpaceLister) HandleSpaceListRequest(ctx echo.Context) error {
+	userID, _ := ctx.Get(context.SubKey).(string)
+	username, _ := ctx.Get(context.UsernameKey).(string)
+	workspaceName := ctx.Param("workspace")
+	doGetWorkspace := len(workspaceName) > 0
+
+	signup, err := s.GetSignupFunc(userID, username)
+	if err != nil {
+		log.Error(nil, err, "error retrieving signup")
+		return errorResponse(ctx, apierrors.NewInternalError(fmt.Errorf("user account lookup failed")), http.StatusInternalServerError)
```
alexeykazakov, line 45: I might be missing something but this is not currently covered by the tests, right?

pkg/proxy/handlers/spacelister.go
```diff
@@ -0,0 +1,155 @@
...
+
+func (s *SpaceLister) workspacesFromSpaceBindings(spaceBindings []*toolchainv1alpha1.SpaceBinding) []toolchainv1alpha1.Workspace {
+	workspaces := []toolchainv1alpha1.Workspace{}
+	for _, spaceBinding := range spaceBindings {
+
+		var ownerName string
+		spaceName := spaceBinding.Labels[toolchainv1alpha1.SpaceBindingSpaceLabelKey]
+		if spaceName == "" { // space may not be initialized
+			// log error and continue so that the api behaves in a best effort manner
+			log.Errorf(nil, fmt.Errorf("spacebinding has no '%s' label", toolchainv1alpha1.SpaceBindingSpaceLabelKey), "unable to get space", "space", spaceName)
```
alexeykazakov, line 125: And this is not covered by the test either, right?

alexeykazakov (approved):
Thanks for adding more tests!
Looks good!
I'm approving the PR but it looks like there are still two error cases which are not covered by the tests. However if it's really tricky to cover them then we can skip it.

alexeykazakov:
/retest

after merging https://github.com/codeready-toolchain/member-operator/pull/408

PR #304, opened 2023-02-13
Add workspace type and object meta

rajivnathan:
For https://issues.redhat.com/browse/ASC-286

Adds `type` for each Workspace returned from the space lister. Type can currently be "home" or "" (empty). Only one can be the "home" Workspace for a user. This is the Workspace that HAC and the proxy will use when no Workspace context is provided.

Also, Workspaces returned will inherit the following object metadata properties from the corresponding Space resources:
- `ResourceVersion`
- `UID`
- `Generation`
- `CreationTimestamp`

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/666

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,8 +4,8 @@ go 1.18
 
 require (
 	github.com/aws/aws-sdk-go v1.44.100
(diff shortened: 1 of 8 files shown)


alexeykazakov (approved):
Looks good!

PR #305, opened 2023-02-14
Extract user_id and account_id claims from token and set annotations

sbryzak:
Set userId and accountId annotations in UserSignup

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.18
 
 require (
 	github.com/aws/aws-sdk-go v1.44.100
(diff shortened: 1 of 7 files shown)


alexeykazakov (approved):

xcoulon:
just curious: why do we need these new annotations?

xcoulon:
> just curious: why do we need these new annotations?

never mind, just found it here: https://github.com/codeready-toolchain/api/pull/342 
(linking in the description here would have helped in understanding the context, though)

PR #307, opened 2023-02-17
Add more detailed logging of proxy requests

rajivnathan:
Adds workspace and user information to the request logger middleware.


alexeykazakov (approved):
Thanks!

PR #309, opened 2023-02-28
Proxy request validation

rajivnathan:
Adds validation to safeguard proxy requests. See https://issues.redhat.com/browse/ASC-293

Also cleans up some of the error handling and messages.

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/674

### DIFF
--- a/pkg/proxy/handlers/spacelister.go
+++ b/pkg/proxy/handlers/spacelister.go
@@ -14,6 +14,7 @@ import (
 	commonproxy "github.com/codeready-toolchain/toolchain-common/pkg/proxy"
 
 	"github.com/labstack/echo/v4"
+	errs "github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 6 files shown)


alexeykazakov (approved):

pkg/proxy/handlers/spacelister.go
```diff
@@ -34,25 +35,59 @@ func NewSpaceLister(app application.Application) *SpaceLister {
...
+	if doGetWorkspace { // specific workspace requested
+		if len(workspaces) != 1 {
+			// not found
+			r := schema.GroupResource{Group: "toolchain.dev.openshift.com", Resource: "workspaces"}
+			return errorResponse(ctx, apierrors.NewNotFound(r, workspaceName))
+		}
+		return getWorkspaceResponse(ctx, workspaces[0])
+	}
+
+	return listWorkspaceResponse(ctx, workspaces)
```
mfrancisc, line 56: just cosmetics - using the `switch`  statement could be another option, something similar to:

```
    switch {
    case len(workspaces) != 1:
        // invalid request with multiple workspaces
        r := schema.GroupResource{Group: "toolchain.dev.openshift.com", Resource: "workspaces"}
		return errorResponse(ctx, apierrors.NewNotFound(r, workspaceName))
    case len(workspaces) == 1:
        // valid request for single workspace
        return getWorkspaceResponse(ctx, workspaces[0])
    default:
        // valid request to list all available workspaces 
        return listWorkspaceResponse(ctx, workspaces)
    }
```
just an alternative idea.

pkg/proxy/handlers/spacelister.go
```diff
@@ -34,25 +35,59 @@ func NewSpaceLister(app application.Application) *SpaceLister {
...
+}
+
+func (s *SpaceLister) ListUserWorkspaces(ctx echo.Context) ([]toolchainv1alpha1.Workspace, error) {
+	userID, _ := ctx.Get(context.SubKey).(string)
+	username, _ := ctx.Get(context.UsernameKey).(string)
```
mfrancisc, line 61: are those inputs from the "user"? is there any validation required or already in place for those?

pkg/proxy/handlers/spacelister.go
```diff
@@ -61,57 +96,16 @@ func (s *SpaceLister) HandleSpaceListRequest(ctx echo.Context) error {
...
-			APIVersion: "toolchain.dev.openshift.com/v1alpha1",
-		},
-		Items: workspaces,
+		return nil, err
 	}
-
-	ctx.Response().Writer.Header().Set("Content-Type", "application/json")
-	ctx.Response().Writer.WriteHeader(http.StatusOK)
-	return json.NewEncoder(ctx.Response().Writer).Encode(workspaceList)
+	return spaceBindings, nil
```
mfrancisc, line 108: minor detail - another option could be:

```
return s.GetInformerServiceFunc().ListSpaceBindings(requirements...)
```

pkg/proxy/handlers/spacelister.go
mfrancisc, line 61: are those inputs from the "users" ? is there any validation required or already in place ?

pkg/proxy/handlers/spacelister.go
mfrancisc, line 108: minor detail, another option could be:

```
return s.GetInformerServiceFunc().ListSpaceBindings(requirements...)
```

pkg/proxy/proxy.go
```diff
@@ -373,3 +389,48 @@ func replaceTokenInWebsocketRequest(req *http.Request, newToken string) {
...
+	// check workspace access
+	isHomeWSRequested := requestedWorkspace == ""
+
+	allowedWorkspace := -1
+	for i, w := range workspaces {
+		if w.Name == requestedWorkspace || (isHomeWSRequested && w.Status.Type == "home") {
+			allowedWorkspace = i
+			break
+		}
+	}
```
mfrancisc, line 403: can there be more workspaces of `.Status.Type: home` in the `workspaces` list ? 
In that case it will provide access to the first one it finds if I'm not wrong.

pkg/proxy/handlers/spacelister.go
rajivnathan: It's based on the token we get from SSO and we have a middleware function in the proxy that is guaranteed to set them or return an error before we ever get here. See https://github.com/codeready-toolchain/registration-service/blob/4b359be48a41f69fb832c82b333b630210cc5d6e/pkg/proxy/proxy.go#L192

pkg/proxy/handlers/spacelister.go
rajivnathan: True! I think I do that way too often haha. Done in f675b9238999e4d587e5cf17cb28f52c890cf8ce

pkg/proxy/proxy.go
rajivnathan: There should be only 1 home workspace for a user. Only the space created for the user when they first sign up should be their home workspace and remain their home workspace. We have some plans to ensure there can only be 1 home workspace. See https://github.com/codeready-toolchain/registration-service/blob/4b359be48a41f69fb832c82b333b630210cc5d6e/pkg/proxy/handlers/spacelister.go#L154

pkg/proxy/handlers/spacelister.go
mfrancisc: Ah I see now, my bad! I got confused between `len(workspaceName)` and `len(workspaces)`. So basically if a workspace was requested but you have more then one in the list or zero workspaces you return a `not found` error. 

If not specific workspace was asked you provide the list , got it! 
Maybe if there is more then 1 workspaces in the list, the error message could be different one, not sure.

pkg/proxy/handlers/spacelister.go
mfrancisc: I see! There is a `extractUserID` function. 

Thanks for the details!

pkg/proxy/proxy.go
mfrancisc: Great! Thanks for the additional details!

PR #311, opened 2023-03-07
PLNSRVCE-1166: add proxy plugin to routes hosted on member clusters

gabemontero:
e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/679

### DIFF
--- a/pkg/application/service/services.go
+++ b/pkg/application/service/services.go
@@ -37,7 +37,7 @@ type VerificationService interface {
 }
 
 type MemberClusterService interface {
-	GetClusterAccess(userID, username, workspace, proxyName string) (*access.ClusterAccess, error)
+	GetClusterAccess(userID, username, workspace, proxyPluginName string) (*access.ClusterAccess, error)
 }
 
 type Services interface {
--- a/pkg/application/service/services.go
+++ b/pkg/application/service/services.go
@@ -14,6 +14,7 @@ type InformerService interface {
 	GetToolchainStatus() (*toolchainv1alpha1.ToolchainStatus, error)
 	GetUserSignup(name string) (*toolchainv1alpha1.UserSignup, error)
 	ListSpaceBindings(reqs ...labels.Requirement) ([]*toolchainv1alpha1.SpaceBinding, error)
+	GetProxyPluginConfig(name string) (*toolchainv1alpha1.ProxyPlugin, error)
 }
 
 type SignupService interface {
@@ -36,7 +37,7 @@ type VerificationService interface {
 }
 
 type MemberClusterService interface {
-	GetClusterAccess(userID, username, workspace string) (*access.ClusterAccess, error)
+	GetClusterAccess(userID, username, workspace, proxyName string) (*access.ClusterAccess, error)
 }
 
 type Services interface {
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -194,16 +194,8 @@ func getWorkspaceContext(req *http.Request) (string, string, error) {
 				return "", "", fmt.Errorf("path %q not a proxied route request", path)
 			}
 		}
-		prefixToTrim := ""
-		// NOTE: just in case the Split behavior is golang version dependent, let's make sure the first entry is empty
-		if len(strings.TrimSpace(segments[0])) == 0 {
-			prefixToTrim = fmt.Sprintf("/%s/%s", segments[1], segments[2])
-			proxyPluginName = segments[2]
-		} else {
-			prefixToTrim = fmt.Sprintf("/%s/%s", segments[0], segments[1])
-			proxyPluginName = segments[1]
-		}
-		req.URL.Path = strings.TrimPrefix(path, prefixToTrim)
+		proxyPluginName = segments[2]
+		req.URL.Path = strings.TrimPrefix(path, fmt.Sprintf("/%s/%s", segments[1], segments[2]))
 		path = req.URL.Path
 	}
 	var workspace string
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -150,13 +150,13 @@ func (p *Proxy) handleRequestAndRedirect(ctx echo.Context) error {
 	userID, _ := ctx.Get(context.SubKey).(string)
 	username, _ := ctx.Get(context.UsernameKey).(string)
 
-	proxyName, workspace, err := getWorkspaceContext(ctx.Request())
+	proxyPluginName, workspace, err := getWorkspaceContext(ctx.Request())
 	if err != nil {
 		return crterrors.NewBadRequest("unable to get workspace context", err.Error())
 	}
 	ctx.Set(workspaceCtxKey, workspace) // set workspace context for logging
 
-	cluster, err := p.app.MemberClusterService().GetClusterAccess(userID, username, workspace, proxyName)
+	cluster, err := p.app.MemberClusterService().GetClusterAccess(userID, username, workspace, proxyPluginName)
 	if err != nil {
 		return crterrors.NewInternalError(errs.New("unable to get target cluster"), err.Error())
 	}
@@ -179,7 +179,7 @@ func (p *Proxy) handleRequestAndRedirect(ctx echo.Context) error {
 
 func getWorkspaceContext(req *http.Request) (string, string, error) {
 	path := req.URL.Path
-	proxyName := ""
+	proxyPluginName := ""
 	// first string off any preceding proxy plugin url segment
 	if strings.HasPrefix(path, pluginsEndpoint) {
 		segments := strings.Split(path, "/")
@@ -191,10 +191,10 @@ func getWorkspaceContext(req *http.Request) (string, string, error) {
 		// NOTE: just in case the Split behavior is golang version dependent, let's make sure the first entry is empty
 		if len(segments[0]) == 0 {
 			prefixToTrim = fmt.Sprintf("/%s/%s", segments[1], segments[2])
-			proxyName = segments[2]
+			proxyPluginName = segments[2]
 		} else {
 			prefixToTrim = fmt.Sprintf("/%s/%s", segments[0], segments[1])
-			proxyName = segments[1]
+			proxyPluginName = segments[1]
 		}
 		req.URL.Path = strings.TrimPrefix(path, prefixToTrim)
 		path = req.URL.Path
@@ -206,7 +206,7 @@ func getWorkspaceContext(req *http.Request) (string, string, error) {
 		// there should be at least 4 segments eg. /workspaces/mycoolworkspace/api/clusterroles counts as 4
 		if len(segments) < 4 {
 			// NOTE, still return valid proxy name in case that helps with debug down the line
-			return proxyName, "", fmt.Errorf("workspace request path has too few segments '%s'; expected path format: /workspaces/<workspace_name>/api/...", path) // nolint:revive
+			return proxyPluginName, "", fmt.Errorf("workspace request path has too few segments '%s'; expected path format: /workspaces/<workspace_name>/api/...", path) // nolint:revive
 		}
 		// get the workspace segment eg. mycoolworkspace
 		workspace = segments[2]
@@ -214,7 +214,7 @@ func getWorkspaceContext(req *http.Request) (string, string, error) {
 		req.URL.Path = strings.TrimPrefix(req.URL.Path, "/workspaces/"+workspace)
 	}
 
-	return proxyName, workspace, nil
+	return proxyPluginName, workspace, nil
 }
 
 func customHTTPErrorHandler(cause error, ctx echo.Context) {
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -173,7 +173,7 @@ func (p *Proxy) handleRequestAndRedirect(ctx echo.Context) error {
 	}
 
 	// Note that ServeHttp is non blocking and uses a go routine under the hood
-	p.newReverseProxy(ctx, cluster).ServeHTTP(ctx.Response().Writer, ctx.Request())
+	p.newReverseProxy(ctx, cluster, len(proxyPluginName) > 0).ServeHTTP(ctx.Response().Writer, ctx.Request())
 	return nil
 }
 
@@ -297,7 +297,7 @@ func extractUserToken(req *http.Request) (string, error) {
 	return token[1], nil
 }
 
-func (p *Proxy) newReverseProxy(ctx echo.Context, target *access.ClusterAccess) *httputil.ReverseProxy {
+func (p *Proxy) newReverseProxy(ctx echo.Context, target *access.ClusterAccess, isPlugin bool) *httputil.ReverseProxy {
 	req := ctx.Request()
 	targetQuery := target.APIURL().RawQuery
 	director := func(req *http.Request) {
@@ -305,6 +305,12 @@ func (p *Proxy) newReverseProxy(ctx echo.Context, target *access.ClusterAccess) 
 		req.URL.Scheme = target.APIURL().Scheme
 		req.URL.Host = target.APIURL().Host
 		req.URL.Path = singleJoiningSlash(target.APIURL().Path, req.URL.Path)
+		if isPlugin {
+			// for non k8s clients testing, like vanilla http clients accessing plugin proxy flows, testing has proven that the request
+			// host needs to be updated in addition to the URL in order to have the reverse proxy contact the openshift
+			// route on the member cluster
+			req.Host = target.APIURL().Host
+		}
 		ctx.Logger().Info(fmt.Sprintf("forwarding %s to %s", origin, req.URL.String()))
 		if targetQuery == "" || req.URL.RawQuery == "" {
 			req.URL.RawQuery = targetQuery + req.URL.RawQuery
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -196,7 +196,7 @@ func getWorkspaceContext(req *http.Request) (string, string, error) {
 		}
 		prefixToTrim := ""
 		// NOTE: just in case the Split behavior is golang version dependent, let's make sure the first entry is empty
-		if len(segments[0]) == 0 {
+		if len(strings.TrimSpace(segments[0])) == 0 {
 			prefixToTrim = fmt.Sprintf("/%s/%s", segments[1], segments[2])
 			proxyPluginName = segments[2]
 		} else {
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -43,6 +43,8 @@ const (
 
 	proxyHealthEndpoint = "/proxyhealth"
 
+	pluginsEndpoint = "/plugins"
+
 	workspaceCtxKey = "workspace"
 )
 
@@ -148,13 +150,13 @@ func (p *Proxy) handleRequestAndRedirect(ctx echo.Context) error {
 	userID, _ := ctx.Get(context.SubKey).(string)
 	username, _ := ctx.Get(context.UsernameKey).(string)
 
-	workspace, err := getWorkspaceContext(ctx.Request())
+	proxyName, workspace, err := getWorkspaceContext(ctx.Request())
 	if err != nil {
 		return crterrors.NewBadRequest("unable to get workspace context", err.Error())
 	}
 	ctx.Set(workspaceCtxKey, workspace) // set workspace context for logging
 
-	cluster, err := p.app.MemberClusterService().GetClusterAccess(userID, username, workspace)
+	cluster, err := p.app.MemberClusterService().GetClusterAccess(userID, username, workspace, proxyName)
 	if err != nil {
 		return crterrors.NewInternalError(errs.New("unable to get target cluster"), err.Error())
 	}
@@ -175,22 +177,44 @@ func (p *Proxy) handleRequestAndRedirect(ctx echo.Context) error {
 	return nil
 }
 
-func getWorkspaceContext(req *http.Request) (string, error) {
+func getWorkspaceContext(req *http.Request) (string, string, error) {
 	path := req.URL.Path
+	proxyName := ""
+	// first string off any preceding proxy plugin url segment
+	if strings.HasPrefix(path, pluginsEndpoint) {
+		segments := strings.Split(path, "/")
+		// NOTE: a split on "/plugins" results in an array with 2 items.  One is the empty string, two is "plugins"
+		if len(segments) < 3 {
+			return "", "", fmt.Errorf("path %q not a proxied route request", path)
+		}
+		prefixToTrim := ""
+		// NOTE: just in case the Split behavior is golang version dependent, let's make sure the first entry is empty
+		if len(segments[0]) == 0 {
+			prefixToTrim = fmt.Sprintf("/%s/%s", segments[1], segments[2])
+			proxyName = segments[2]
+		} else {
+			prefixToTrim = fmt.Sprintf("/%s/%s", segments[0], segments[1])
+			proxyName = segments[1]
+		}
+		req.URL.Path = strings.TrimPrefix(path, prefixToTrim)
+		path = req.URL.Path
+	}
 	var workspace string
 	// handle specific workspace request eg. /workspaces/mycoolworkspace/api/clusterroles
 	if strings.HasPrefix(path, "/workspaces/") {
 		segments := strings.Split(path, "/")
 		// there should be at least 4 segments eg. /workspaces/mycoolworkspace/api/clusterroles counts as 4
 		if len(segments) < 4 {
-			return "", fmt.Errorf("workspace request path has too few segments '%s'; expected path format: /workspaces/<workspace_name>/api/...", path) // nolint:revive
+			// NOTE, still return valid proxy name in case that helps with debug down the line
+			return proxyName, "", fmt.Errorf("workspace request path has too few segments '%s'; expected path format: /workspaces/<workspace_name>/api/...", path) // nolint:revive
 		}
 		// get the workspace segment eg. mycoolworkspace
 		workspace = segments[2]
 		// remove workspaces/mycoolworkspace from the request path before forwarding the request
 		req.URL.Path = strings.TrimPrefix(req.URL.Path, "/workspaces/"+workspace)
 	}
-	return workspace, nil
+
+	return proxyName, workspace, nil
 }
 
 func customHTTPErrorHandler(cause error, ctx echo.Context) {
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -43,7 +43,7 @@ const (
 
 	proxyHealthEndpoint = "/proxyhealth"
 
-	pluginsEndpoint = "/plugins"
+	pluginsEndpoint = "/plugins/"
 
 	workspaceCtxKey = "workspace"
 )
@@ -183,9 +183,16 @@ func getWorkspaceContext(req *http.Request) (string, string, error) {
 	// first string off any preceding proxy plugin url segment
 	if strings.HasPrefix(path, pluginsEndpoint) {
 		segments := strings.Split(path, "/")
-		// NOTE: a split on "/plugins" results in an array with 2 items.  One is the empty string, two is "plugins"
+		// NOTE: a split on "/plugins/" results in an array with 2 items.  One is the empty string, two is "plugins", 3 is empty string
+		// behavior is not unique to "/";  "," and ",plugins," works the same way
 		if len(segments) < 3 {
 			return "", "", fmt.Errorf("path %q not a proxied route request", path)
+		}
+		if len(segments) == 3 {
+			// need to distinguish between the third entry being "" vs "<plugin-name>"
+			if len(strings.TrimSpace(segments[2])) == 0 {
+				return "", "", fmt.Errorf("path %q not a proxied route request", path)
+			}
 		}
 		prefixToTrim := ""
 		// NOTE: just in case the Split behavior is golang version dependent, let's make sure the first entry is empty
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -211,9 +211,18 @@ func getWorkspaceContext(req *http.Request) (string, string, error) {
 	if strings.HasPrefix(path, "/workspaces/") {
 		segments := strings.Split(path, "/")
 		// there should be at least 4 segments eg. /workspaces/mycoolworkspace/api/clusterroles counts as 4
-		if len(segments) < 4 {
-			// NOTE, still return valid proxy name in case that helps with debug down the line
-			return proxyPluginName, "", fmt.Errorf("workspace request path has too few segments '%s'; expected path format: /workspaces/<workspace_name>/api/...", path) // nolint:revive
+		if len(segments) < 4 && len(proxyPluginName) == 0 {
+			return "", "", fmt.Errorf("workspace request path has too few segments '%s'; expected path format: /workspaces/<workspace_name>/api/...", path) // nolint:revive
+		}
+		// with proxy plugins, the route host is sufficient, and hence do not need api/...
+		if len(segments) < 3 {
+			return "", "", fmt.Errorf("workspace request path has too few segments '%s'; expected path format: /workspaces/<workspace_name>/<optional path>", path) // nolint:revive
+		}
+		if len(segments) == 3 {
+			// need to distinguish between the third entry being "" vs "<plugin-name>"
+			if len(strings.TrimSpace(segments[2])) == 0 {
+				return "", "", fmt.Errorf("workspace request path has too few segments '%s'; expected path format: /workspaces/<workspace_name>/<optional path>", path) // nolint:revive
+			}
 		}
 		// get the workspace segment eg. mycoolworkspace
 		workspace = segments[2]
--- a/pkg/proxy/proxy_test.go
+++ b/pkg/proxy/proxy_test.go
@@ -580,6 +580,7 @@ func (s *TestProxySuite) TestGetWorkspaceContext() {
 		expectedWorkspace string
 		expectedPath      string
 		expectedErr       string
+		expectedProxy     string
 	}{
 		"valid workspace context": {
 			path:              "/workspaces/myworkspace/api",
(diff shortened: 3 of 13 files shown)


gabemontero:
/assign @alexeykazakov 
/assign @rajivnathan 

minimally to add the ok-to-test

and  then review and/or instructions on what I missed wrt y'all's process (like do I need to create the sandbox-sre instance first??)

thanks

pkg/application/service/services.go
```diff
@@ -36,7 +37,7 @@ type VerificationService interface {
 }
 
 type MemberClusterService interface {
-	GetClusterAccess(userID, username, workspace string) (*access.ClusterAccess, error)
+	GetClusterAccess(userID, username, workspace, proxyName string) (*access.ClusterAccess, error)
```
alexeykazakov, line 40: Can we call this `pluginName` or `proxyPluginName` instead? `proxyName` could be confusing. There is only one proxy. Just multiple plugins.

pkg/proxy/proxy.go
```diff
@@ -148,13 +150,13 @@ func (p *Proxy) handleRequestAndRedirect(ctx echo.Context) error {
 	userID, _ := ctx.Get(context.SubKey).(string)
 	username, _ := ctx.Get(context.UsernameKey).(string)
 
-	workspace, err := getWorkspaceContext(ctx.Request())
+	proxyName, workspace, err := getWorkspaceContext(ctx.Request())
```
alexeykazakov, line 153: Same here. `pluginName` or `proxyPluginName`.

pkg/proxy/proxy.go
```diff
@@ -175,22 +177,44 @@ func (p *Proxy) handleRequestAndRedirect(ctx echo.Context) error {
 	return nil
 }
 
-func getWorkspaceContext(req *http.Request) (string, error) {
+func getWorkspaceContext(req *http.Request) (string, string, error) {
 	path := req.URL.Path
+	proxyName := ""
```
alexeykazakov, line 182: And here.

pkg/proxy/proxy_test.go
```diff
@@ -605,6 +606,50 @@ func (s *TestProxySuite) TestGetWorkspaceContext() {
...
+			expectedErr:       "",
+			expectedProxy:     "tekton-results",
+		},
+		"workspace instead of workspaces with route": {
+			path:              "/plugins/tekton-results/workspace/myworkspace/api",
+			expectedWorkspace: "",
+			expectedPath:      "/workspace/myworkspace/api",
+			expectedErr:       "",
+			expectedProxy:     "tekton-results",
+		},
```
alexeykazakov, line 652: Let's add two more test cases which would test that the plugin name is expected in the beginning of the path only:
- `/plugins/tekton-results/workspaces/myworkspace/api/plugins/something` should work. In this case the actual kube api contains "plugins" which has nothing to do with our proxy plugins.
- `/api/plugins/something`. There is no plugins or workspace context. But the kube api still includes some "plugins". Still should work.

pkg/proxy/service/cluster_service.go
```diff
@@ -36,7 +40,7 @@ func NewMemberClusterService(context servicecontext.ServiceContext, options ...O
 	return si
 }
 
-func (s *ServiceImpl) GetClusterAccess(userID, username, workspace string) (*access.ClusterAccess, error) {
+func (s *ServiceImpl) GetClusterAccess(userID, username, workspace, proxyName string) (*access.ClusterAccess, error) {
```
alexeykazakov, line 43: And here and in all other places. `proxyName` again.

alexeykazakov (commented):
Looks good! Nice work. I have a few minor comments though.

And we will need to get it covered by some e2e tests too: https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/e2e/proxy_test.go

Please create a paired e2e test PR there (the branch should have the same name as the branch for this PR).

alexeykazakov:
>  (like do I need to create the sandbox-sre instance first??)

Not really. Let's get the plugin mechanism merged first. Then we will add some plugin manifests to sandbox-sre. Btw let's make sure that we have a test which verifies that the main proxy flow still works if there is no plugins available (and the proxy URL doesn't have any plugins in the path) in case we miss this test case.

gabemontero:
thanks @alexeykazakov 

I've pushed the variable name change and the additional unit tests in a separate commit to narrow your review scope.  I will squash if you like if you are happy with them.

Also looked at the existing e2e failure.  The registration service is not coming up because of permissions for the new CRD.

I believe we need https://github.com/codeready-toolchain/toolchain-cicd/pull/85 to be merged to address that.

I'll start working on the new, in parallel e2e PR after lunch my time

alexeykazakov (commented):
Looks great. Thank you!

> I've pushed the variable name change and the additional unit tests in a separate commit to narrow your review scope. I will squash if you like if you are happy with them.

Please do not squash commits in the PRs. As you mentioned it's much easier to do the followup reviews when there are new commits instead of push with the force. Everything will be squashed by GitHub automatically when merging to master anyway.

> Also looked at the existing e2e failure. The registration service is not coming up because of permissions for the new CRD.
I believe we need https://github.com/codeready-toolchain/toolchain-cicd/pull/85 to be merged to address that.

Right. You can still make your e2e-test working by replacing the cicd github reference to your branch here: https://github.com/codeready-toolchain/toolchain-e2e/blob/707edb7a0f4698ac2982b0d708941dcdc614da7a/make/run-cicd-script.mk#L2
When the cicd PR is merged you revert this change in your e2e-tests PR.

> I'll start working on the new, in parallel e2e PR after lunch my time

+1

gabemontero:
interesting - 2 of the 3 registration service pods came up OK, but one of them still had RBAC issues with the new CRD

check out https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/codeready-toolchain_registration-service/311/pull-ci-codeready-toolchain-registration-service-master-e2e/1633546514296999936/artifacts/e2e/gather-extra/artifacts/pods/

maybe a timing issue between the merging of the toolchain-cicd PR and the retest?  Or is it a leader election thing?

let's try one more time, then level set

/retest

gabemontero:
> Looks great. Thank you!
> 
> > I've pushed the variable name change and the additional unit tests in a separate commit to narrow your review scope. I will squash if you like if you are happy with them.
> 
> Please do not squash commits in the PRs. As you mentioned it's much easier to do the followup reviews when there are new commits instead of push with the force. Everything will be squashed by GitHub automatically when merging to master anyway.
> 
> > Also looked at the existing e2e failure. The registration service is not coming up because of permissions for the new CRD.
> > I believe we need [codeready-toolchain/toolchain-cicd#85](https://github.com/codeready-toolchain/toolchain-cicd/pull/85) to be merged to address that.
> 
> Right. You can still make your e2e-test working by replacing the cicd github reference to your branch here: https://github.com/codeready-toolchain/toolchain-e2e/blob/707edb7a0f4698ac2982b0d708941dcdc614da7a/make/run-cicd-script.mk#L2 When the cicd PR is merged you revert this change in your e2e-tests PR.
> 
> **UPDATE**: Never mind. I've merged [codeready-toolchain/toolchain-cicd#85](https://github.com/codeready-toolchain/toolchain-cicd/pull/85)

yeah @alexeykazakov looks like simply merging  [codeready-toolchain/toolchain-cicd#85](https://github.com/codeready-toolchain/toolchain-cicd/pull/85) is not sufficient .... still get the rbac error in one (presumably the leader) of the registration-service pods

I'll update make/run-cici-script.mk per ^^ in my upcoming e2e PR

> 
> > I'll start working on the new, in parallel e2e PR after lunch my time
> 
> +1

gabemontero:
Yeah I've reproduced the situation deploying locally on my own cluster.  The `registration-service` role is not impacted by my toolchain-cicd pr https://github.com/codeready-toolchain/toolchain-cicd/pull/85:

```
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  annotations:
    toolchain.dev.openshift.com/last-applied-configuration: '{"apiVersion":"rbac.authorization.k8s.io/v1","kind":"Role","metadata":{"labels":{"provider":"codeready-toolchain"},"name":"registration-service","namespace":"toolchain-host-operator"},"rules":[{"apiGroups":["toolchain.dev.openshift.com"],"resources":["usersignups"],"verbs":["create","get","update","list","watch"]},{"apiGroups":["toolchain.dev.openshift.com"],"resources":["masteruserrecords","socialevents","spacebindings","spaces","toolchainconfigs","toolchainstatuses"],"verbs":["get","list","watch"]},{"apiGroups":["toolchain.dev.openshift.com"],"resources":["bannedusers","toolchainclusters"],"verbs":["get","list"]},{"apiGroups":[""],"resources":["secrets","configmaps"],"verbs":["get","list"]}]}'
  creationTimestamp: "2023-03-09T16:27:04Z"
  labels:
    provider: codeready-toolchain
  name: registration-service
  namespace: toolchain-host-operator
  ownerReferences:
  - apiVersion: toolchain.dev.openshift.com/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: ToolchainConfig
    name: config
    uid: 219a7c80-1121-4d6c-80c2-894a9b7350d7
  resourceVersion: "140201"
  uid: 5a64e855-ecba-40eb-82da-cf34f86422e2
rules:
- apiGroups:
  - toolchain.dev.openshift.com
  resources:
  - usersignups
  verbs:
  - create
  - get
  - update
  - list
  - watch
- apiGroups:
  - toolchain.dev.openshift.com
  resources:
  - masteruserrecords
  - socialevents
  - spacebindings
  - spaces
  - toolchainconfigs
  - toolchainstatuses
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - toolchain.dev.openshift.com
  resources:
  - bannedusers
  - toolchainclusters
  verbs:
  - get
  - list
- apiGroups:
  - ""
  resources:
  - secrets
  - configmaps
  verbs:
  - get
  - list
```

I'll see if I can sort it out on my own, but if  this turns on any lightbulbs @alexeykazakov @rajivnathan please do chime in.

gabemontero:
found it @alexeykazakov @rajivnathan - https://github.com/codeready-toolchain/host-operator/blob/master/deploy/registration-service/registration-service.yaml#L33-L39

gabemontero:
let's see if the PR pairing needs the ok-to-test in https://github.com/codeready-toolchain/host-operator/pull/762

/retest

gabemontero:
updating https://github.com/codeready-toolchain/host-operator/blob/master/deploy/registration-service/registration-service.yaml#L33-L39 fixed my local repro of the registration-service startup issue

gabemontero:
e2e update: 
- https://github.com/codeready-toolchain/host-operator/pull/763 addresses the rbac issued cited earlier and seen in runs in this PR so far
- for the new e2e test, my local testing uncovered the need for https://github.com/codeready-toolchain/toolchain-common/pull/303.  Once that merges, I'll vendor in the toolchain-common change into the branch for this PR, and continue with my local testing of the new e2e.  Once I see that passing locally, I'll push the paired / same branch name toolchain-e2e PR and we can retest this PR with a corresponding new test

pkg/proxy/service/cluster_service_test.go
```diff
@@ -224,48 +278,102 @@ func (s *TestClusterServiceSuite) TestGetClusterAccess() {
...
+					route.Name = key.Name
+					route.Status.Ingress = []routev1.RouteIngress{
 						{
-							Config: &commoncluster.Config{
-								Name:        "member-3",
-								APIEndpoint: "https://api.endpoint.member-3.com:6443",
-								Type:        commoncluster.Member,
-								RestConfig:  &rest.Config{},
-							},
+							Host: "https://api.endpoint.member-2.com:6443",
```
rajivnathan, line 294: This is the same URL as when the route is not set right? I think it'd be better to use a different URL for the route ingress for testing purposes. eg. `https://myservice.endpoint.member-2.com`

pkg/proxy/proxy_test.go
```diff
@@ -580,6 +580,7 @@ func (s *TestProxySuite) TestGetWorkspaceContext() {
 		expectedWorkspace string
 		expectedPath      string
 		expectedErr       string
+		expectedProxy     string
```
rajivnathan, line 583: Let's rename this to `expectedPlugin` for consistency

pkg/proxy/proxy_test.go
```diff
@@ -599,12 +600,69 @@ func (s *TestProxySuite) TestGetWorkspaceContext() {
...
+			expectedPath:      "/api/plugins/something",
+			expectedErr:       "",
+		},
 		"workspace instead of workspaces": {
 			path:              "/workspace/myworkspace/api",
 			expectedWorkspace: "",
 			expectedPath:      "/workspace/myworkspace/api",
 			expectedErr:       "",
 		},
+		"valid workspace context with route": {
```
rajivnathan, line 615:
```suggestion
		"valid workspace context with plugin": {
```

pkg/proxy/proxy_test.go
```diff
@@ -599,12 +600,69 @@ func (s *TestProxySuite) TestGetWorkspaceContext() {
...
+			path:              "/plugins/tekton-results/workspaces/myworkspace/api",
+			expectedWorkspace: "myworkspace",
+			expectedPath:      "/api",
+			expectedErr:       "",
+			expectedProxy:     "tekton-results",
+		},
+		"valid workspace context with route plus plugin in kube api portion": {
```
rajivnathan, line 622:
```suggestion
		"valid workspace context with plugin plus another plugin in kube api portion": {
```

pkg/proxy/proxy_test.go
```diff
@@ -599,12 +600,69 @@ func (s *TestProxySuite) TestGetWorkspaceContext() {
...
+			path:              "/plugins/tekton-results/workspaces/myworkspace/api/plugins/something",
+			expectedWorkspace: "myworkspace",
+			expectedPath:      "/api/plugins/something",
+			expectedErr:       "",
+			expectedProxy:     "tekton-results",
+		},
+		// the pre-plugin testcases all assumed /workspace would exists or /api/... exists; should we assume /plugins/<plugin name>/workspace or /plugins/<plugin name/api/... exists?
```
rajivnathan, line 629: Not sure I understand what you mean here. If you mean that it doesn't check anything before continuing with the request then yes, it just processes the request according to the context. The request would fail later in the flow if there's something wrong. Is that what you meant?

rajivnathan (commented):
Nice! Looks great!

pkg/proxy/proxy_test.go
gabemontero: yes that is what I was getting at .... thanks for confirming

I can remove the comments with my next update to the PR

pkg/proxy/proxy.go
```diff
@@ -175,22 +177,44 @@ func (p *Proxy) handleRequestAndRedirect(ctx echo.Context) error {
 	return nil
 }
 
-func getWorkspaceContext(req *http.Request) (string, error) {
+func getWorkspaceContext(req *http.Request) (string, string, error) {
 	path := req.URL.Path
+	proxyPluginName := ""
+	// first string off any preceding proxy plugin url segment
+	if strings.HasPrefix(path, pluginsEndpoint) {
```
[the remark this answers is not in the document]
gabemontero: part of next push

the implications of being more precise and the "interesting" behavior of `strings.Split(path, "/")` had some ramifications to the processing of `getWorkspaceContext`.  Be sure to take a look when I push the update.

alexeykazakov (commented):
Looks good! Just waiting for additional e2e tests which would test the plugin mechanism before approving it :)

gabemontero:
> Looks good! Just waiting for additional e2e tests which would test the plugin mechanism before approving it :)

thanks

so on the e2e's as I am about to wrap up this Friday
- I've been working on a test that uses the openshift console's route as the destination
- I've been able to verify that the reverse proxy correctly forwards to the url for my test cluster's console, like `https://console-openshift-console.apps.gmontero412.devcluster.openshift.com/`
- but the best I can do on the response is a 401 unauthenticated
- ultimately, the user token generated by how proxy_test.go in the e2e repo currently sets it up is failing in the `extractUserID` method in the registration service's proxy.go, with the error message `invalid bearer token: unable to extract userID from token: unknown kid` ... so ultimately that forward from the reverse proxy does not have the token set, and so sure, 401
- I started hunting in some of the other e2e's and found `commonauth.GenerateSignedE2ETestToken` but that did not work (it got a rc 500) and there was  an error log of `unable to get target cluster: user is not provisioned (yet)` in the registration service (that log is from cluster_service.go)

Even getting the 401, along with the forward log, proves the basic flow is working, but of course, and rc 200 would be ideal.  Maybe a fall back plan based on answers to my questions below.

Before I get too crazy and starting pulling tokens from service accounts, or dive into / learn more of the sandbox / toolchain details, I thought it was time to get a sanity check @alexeykazakov @rajivnathan 

- any pointers on how I should set up a token from https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/e2e/proxy_test.go so I can access the openshift console for a the member cluster in the e2e?
- is the basic premise of using the console route for the e2e "wrong"?  If so, alternative suggestions please

If it helps to push my e2e branch as is to fill in the blanks for all of ^^ do let me know

I am about to push a small additional commit for some minor tweaks to this PR that facilitated the "use openshift console route" test.

thanks

gabemontero:
>     * I started hunting in some of the other e2e's and found `commonauth.GenerateSignedE2ETestToken` but that did not work (it got a rc 500) and there was  an error log of `unable to get target cluster: user is not provisioned (yet)` in the registration service (that log is from cluster_service.go)
> 

is resolving the user not being provisioned as simple as creating a user signup per https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/e2e/parallel/registration_service_test.go#L732-L753 ?

rajivnathan:
@gabemontero 
>any pointers on how I should set up a token from https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/e2e/proxy_test.go so I can access the openshift console for a the member cluster in the e2e?

There's a convenient function in proxy_test.go for signing up a user: https://github.com/codeready-toolchain/toolchain-e2e/blob/55e71420c4fecffd6dd7d1e14bb065b8eab53f97/test/e2e/proxy_test.go#L560

You can see it used here: https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/e2e/proxy_test.go#L98-L118

It *should* work if that's how you're setting up the user in the test. If it's not working then maybe something broke in the proxy logic and we'll need to take a closer look. If you share your e2e branch and I can take a closer look too

>is the basic premise of using the console route for the e2e "wrong"? 

When I tried curling to my dev cluster's console it doesn't actually need a token to get a 200 response so I'm guessing the 401 is coming from the proxy itself and is failing before forwarding the request to the console. So the basic premise should be fine for testing.

gabemontero:
> @gabemontero
> 
> > any pointers on how I should set up a token from https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/e2e/proxy_test.go so I can access the openshift console for a the member cluster in the e2e?
> 
> There's a convenient function in proxy_test.go for signing up a user: https://github.com/codeready-toolchain/toolchain-e2e/blob/55e71420c4fecffd6dd7d1e14bb065b8eab53f97/test/e2e/proxy_test.go#L560
> 
> You can see it used here: https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/e2e/proxy_test.go#L98-L118
> 
> It _should_ work if that's how you're setting up the user in the test. If it's not working then maybe something broke in the proxy logic and we'll need to take a closer look. If you share your e2e branch and I can take a closer look too

yeah my insertion of the tests worked off of the existing calls to `createAppStudioUser` that were at https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/e2e/proxy_test.go#L98-L118 so unless I'm missing something those are the users (with their tokens) being employed

I'm booting up a cluster now, and will first try my idea at https://github.com/codeready-toolchain/registration-service/pull/311#issuecomment-1465334582, just cause I'm curious, but unless that works and is a useful point of comparison, will revert back to the form that produced the 401 and get an e2e PR up so you can take that close look

thanks @rajivnathan 

> 
> > is the basic premise of using the console route for the e2e "wrong"?
> 
> When I tried curling to my dev cluster's console it doesn't actually need a token to get a 200 response so I'm guessing the 401 is coming from the proxy itself and is failing before forwarding the request to the console. So the basic premise should be fine for testing.

interesting ... but yeah, I was assuming the oauth flow immediately came into play, but in hindsight, sure, that 200 probably stems from providing the login prompt

still, a good clue that the 401 stemmed from a different step in the flow (i.e. from the proxy itself as you said) than I was thinking

gabemontero:
ok @rajivnathan the e2e run https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/codeready-toolchain_registration-service/311/pull-ci-codeready-toolchain-registration-service-master-e2e/1635619361240649728 exhibits the same failure I see locally with my new test case defined in https://github.com/codeready-toolchain/toolchain-e2e/pull/679

gabemontero:
got some more time today @rajivnathan and added some debug on the registration side around the bearer token processing, as well as simulated that bearer token processing on the e2e side

the token on the e2e side passes the failing check on the registration service side, and the token being sent on the e2e side is definitely different than the one processed on the registration service side

something is mutating it (perhaps expected?)

I'll post the debug commits after dinner, see if I see the same results in the e2e run

@alexeykazakov fyi

thanks

gabemontero:
ok debug pushed

rajivnathan:
@rajivnathan yeah, I saw the behaviour you were talking about when I tried it this afternoon but not sure what's causing it. Will take a closer look tomorrow morning. Will take a look at the debug stuff you added as well. Thanks

gabemontero:
thank for the confirmation @rajivnathan 

debug from the last e2e ^^

on the e2e test side:

```
    proxy_test.go:835: GGM kid header d5693c31-7016-46a4-bbe4-867e6d6a3b3a
    proxy_test.go:850: GGM len key map 1
    proxy_test.go:857: GGMGGM found public key for d5693c31-7016-46a4-bbe4-867e6d6a3b3a
    proxy_test.go:274: GGM from string for user token len 1160 returned error <nil> token eyJhbGciOiJSUzI1NiIsImtpZCI6ImQ1NjkzYzMxLTcwMTYtNDZhNC1iYmU0LTg2N2U2ZDZhM2IzYSIsInR5cCI6IkpXVCJ9.eyJleHAiOjE2ODE0MzE4MDQsImp0aSI6IjZlN2RjMDc0LTEyNWUtNDIyNi05YjU2LWU2ZmFmMGY0MGYxZCIsImlhdCI6MTY3ODgzOTgwNCwiaXNzIjoiY29kZXJlYWR5LXRvb2xjaGFpbiIsInN1YiI6IjAyYjFlZWM4LWYxZWQtNDc2ZS1hZTkxLWM4NGI0MDljYmIxYyIsInV1aWQiOiIwMmIxZWVjOC1mMWVkLTQ3NmUtYWU5MS1jODRiNDA5Y2JiMWMiLCJwcmVmZXJyZWRfdXNlcm5hbWUiOiJwcm94eW1lbWJlcjEiLCJzZXNzaW9uX3N0YXRlIjoiYTU5YTUxZTItYTVhNS00ZjA2LWI5OWUtMTM2YTNjYmMwYTJkIiwidHlwIjoiQmVhcmVyIiwiYXBwcm92ZWQiOnRydWUsIm5hbWUiOiJUZXN0IFVzZXIiLCJjb21wYW55IjoiQ29tcGFueSBJbmMuIiwiZ2l2ZW5fbmFtZSI6IlRlc3QiLCJmYW1pbHlfbmFtZSI6IlVzZXIiLCJlbWFpbCI6InRlc3R1c2VyLTc2ODQwZGU0LTdhNjctNGQyMC05MWI2LTRlN2I5MWZhYjY5YkB0ZXN0LmNvbSIsImVtYWlsX3ZlcmlmaWVkIjp0cnVlLCJvcmlnaW5hbF9zdWIiOiIiLCJ1c2VyX2lkIjoiIiwiYWNjb3VudF9pZCI6IiJ9.dTL005uVBeZqOllcBewVG3bzYWr474qaUt9P7JU4Kmfh5nqBr-2na7w34h0S4JOHGY_flfjhAvjof_PDpPVTeiZolEAUtZtK4RzHXt1XHb94nVJkNURiz2fOLa3DE8jB5rY0Aen0GMwlcvK0-xXcO5WLDK97V4mhmqgaZFqQz4i6eGIw78KvrBpH2S4eynjEokUfXR7o9sdjV8j5d0_ZSvQxfLHQ8X__qQ8PI9DrKV6cRqBlj17SW1Xb3X3pMkoXkkiqJK24Ek99WjFsbWCV7lOw8BrSKN_cyCzlDrA9QmS6VwCsrrQW8xPqFO1NRCu8wXRLPGe9ZM71Bhf3as7aXg
    proxy_test.go:284: unexpected http return code of 401 with body text invalid bearer token: unable to extract userID from token: unknown kid
```

within the registration service:

```
{"level":"info","ts":"2023-03-15T00:24:00.279Z","logger":"registration-service","msg":"GGM key map len 1 kid LRt5xBNjzn63cEnTgy1hbFZdcyr_eAhPk_To6b-59hw keymap map[string]*rsa.PublicKey{\"d5693c31-7016-46a4-bbe4-867e6d6a3b3a\":(*rsa.PublicKey)(0xc0008a6c00)}","timestamp":"Wed, 15 Mar 2023 00:24:00 +0000","commit":"0"}
{"level":"info","ts":"2023-03-15T00:24:00.279Z","logger":"registration-service","msg":"GGM key err on token string of len 1353","timestamp":"Wed, 15 Mar 2023 00:24:00 +0000","commit":"0"}
{"time":"2023-03-15T00:24:00.279307254Z","level":"ERROR","prefix":"echo","file":"proxy.go","line":"242","message":"invalid bearer token: unable to extract userID from token: unknown kid"}
{"level":"info","ts":"2023-03-15T00:24:00.279Z","logger":"registration-service","msg":"GGM for user token len 1353 got extract user ID error unknown kid token eyJhbGciOiJSUzI1NiIsImtpZCI6IkxSdDV4Qk5qem42M2NFblRneTFoYkZaZGN5cl9lQWhQa19UbzZiLTU5aHcifQ.eyJpc3MiOiJrdWJlcm5ldGVzL3NlcnZpY2VhY2NvdW50Iiwia3ViZXJuZXRlcy5pby9zZXJ2aWNlYWNjb3VudC9uYW1lc3BhY2UiOiJ0b29sY2hhaW4tbWVtYmVyLTE1MDAwMzUzIiwia3ViZXJuZXRlcy5pby9zZXJ2aWNlYWNjb3VudC9zZWNyZXQubmFtZSI6InRvb2xjaGFpbmNsdXN0ZXItbWVtYmVyLXRva2VuLTVrd2t3Iiwia3ViZXJuZXRlcy5pby9zZXJ2aWNlYWNjb3VudC9zZXJ2aWNlLWFjY291bnQubmFtZSI6InRvb2xjaGFpbmNsdXN0ZXItbWVtYmVyIiwia3ViZXJuZXRlcy5pby9zZXJ2aWNlYWNjb3VudC9zZXJ2aWNlLWFjY291bnQudWlkIjoiNjg5YjY3YzMtYTdiYS00MGQzLWExODAtYzUyNTM5MzZiMDU5Iiwic3ViIjoic3lzdGVtOnNlcnZpY2VhY2NvdW50OnRvb2xjaGFpbi1tZW1iZXItMTUwMDAzNTM6dG9vbGNoYWluY2x1c3Rlci1tZW1iZXIifQ.a5Ns5OodM-mzkBrtGZoR6YBrv4APU8rtMmVP8zMAtkAS3mWw276ctPNFXko4AtQFuPSJGiMEOSNkdguk11ArBXJq9L6h3XQbEa3aMZxfLGXqPX5jK4SUYb4O_hZgmt0jfqWGWn6VYsrEmwMk6gJ4aSwLhlpdeHxrY8JDv2sHP57PZSVQBTsm_wTTB0aenb_UvNwiNJOC0Ib1a2jt0JfGuWPmj3m149tmly2Nr4LXvvY_VqPKY8yAfEN9w7LX3sie0SsrqU1LlpHLCNOkbXXR7UNynxcnpsdX88IQYhWolTu9B5e_SI-z1Kq_7bAPqWsFx-TjWXj7IFOUKvUsBxy0s4sKaXJ5bLmdy-EjvUa46PU9fZLJZjdh5LpGbvgvFa8lliKGfZi3W4qxxkK_AyivLn88-8sHRCn80nECnv8I37VqS7T53RlfP8io9ZYzSZpaiWzHvDHewBzPpaRKDtSTPYri5mONhquEK4TWCTUN5l7VJp_mW2koZuNwrAuNfqKBTSVyKQkb0b0EssvHXXGzjBAcgaal6TJJZCoxcDNiSVnjSG5KmzyqD_6gJMSy0nER7dvqM9_dhq0sQB3IG5f87N7O3WQjcJMgLgugB3w4kk6tBntBZs5Wj3F2GsZ5zI1H8nPJjAxiIeNrp0rxReXq_WJVUFDoH9-lmGZDK5GL3es","timestamp":"Wed, 15 Mar 2023 00:24:00 +0000","commit":"0"}
```

definitely different kid headers, no longer matching the predefined e2e key, thus leading to the unknown kid error, and the tokens differ in size by almost 200 bytes

alexeykazakov:
Hm.. It's some mystery :)
See my comment here: https://github.com/codeready-toolchain/toolchain-e2e/pull/679#discussion_r1137521733

rajivnathan:
I don't think it's the Add vs Set thing because I paused the test just after the request is created and I tried using `curl` against the proxy endpoint and it resulted in the same error:
```
curl -k -H "Authorization: Bearer <token>" https://api-toolchain-host-operator.apps.<cluster>/plugins/openshift-console/workspaces/proxymember1

invalid bearer token: unable to extract userID from token: unknown kid
```

But the same token but with a workspace context request to get configmaps from the user's workspace worked just fine!
```
curl -k -H "Authorization: Bearer <token>" https://api-toolchain-host-operator.apps.<cluster>/workspaces/proxymember1/api/v1/namespaces/proxymember1-tenant/configmaps
```

So it appears to only be a problem with the plugin requests flow.

gabemontero:
> I don't think it's the Add vs Set thing because I paused the test just after the request is created and I tried using `curl` against the proxy endpoint and it resulted in the same error:
> 
> ```
> curl -k -H "Authorization: Bearer <token>" https://api-toolchain-host-operator.apps.<cluster>/plugins/openshift-console/workspaces/proxymember1
> 
> invalid bearer token: unable to extract userID from token: unknown kid
> ```
> 
> But the same token but with a workspace context request to get configmaps from the user's workspace worked just fine!
> 
> ```
> curl -k -H "Authorization: Bearer <token>" https://api-toolchain-host-operator.apps.<cluster>/workspaces/proxymember1/api/v1/namespaces/proxymember1-tenant/configmaps
> ```
> 
> So it appears to only be a problem with the plugin requests flow.

interesting @rajivnathan 

this got me thinking, along with @alexeykazakov 's observation about my debug revealing that the member operator service account token is getting referenced

Any chance that my reuse of `handleRequestAndRedirect` for the proxy flow, with user initialization calls like https://github.com/codeready-toolchain/registration-service/blob/5b5815ea9e78fdf5f5526b0a51a4f65e89707ac4/pkg/proxy/proxy.go#L149-L151 is leading to this somehow?

And if so, I need to change the initialization at https://github.com/codeready-toolchain/registration-service/blob/5b5815ea9e78fdf5f5526b0a51a4f65e89707ac4/pkg/proxy/proxy.go#L124-L125 to call a different a new method that is called when `/plugins` is the start of the URI, where the necessary changes are in place to get the bearer token from the request?

gabemontero:
with that, I believe my unit tests in https://github.com/codeready-toolchain/registration-service/pull/311/files#diff-41ac6f84c8e979d741aa5693c53da94f155d83ef93c5761d3899609971bfe375 don't go through the `StartProxy` path that sets up the router to call `handleRequestAndRedirect`

I'm going to try and add a proxy unit test to `TestProxy` and see if that confirms this theory at all.  But certainly let me know any yay's or nay's wrt plausibility @rajivnathan @alexeykazakov if I've hit on something.

gabemontero:
> > Any chance that my reuse of handleRequestAndRedirect for the proxy flow is leading to this somehow?
> 
> I don't think that should be an issue so I wouldn't change that just yet.
> 
> > I believe my unit tests don't go through the StartProxy path that sets up the router to call handleRequestAndRedirect
> 
> Good point about unit tests, that could give us a clue as to what's going on

I could not reproduce the wrong token / invalid kid error with unit tests today in the `TestProxy` test today.  Pending another idea that you've got @rajivnathan I think we need to figure out how to trace the flow ... does the registration service have access logs somewhere?  Are there labstack traces we can turn on?

gabemontero:
I've pushed a commit with my functional `TestProxy` unit test addition for the proxy plugin @rajivnathan 

The closest I could get to reproducing the e2e failure was an invalid token not enough segments, but not the unknown kid error.

Aside from waiting to see if you have any feedback from my earlier questions in https://github.com/codeready-toolchain/registration-service/pull/311#issuecomment-1470978574, I'm going to look into adding tracing at the start of `handleRequestAndRedirect` to confirm what is in the request headers as soon as `labstack/echo` calls us.

gabemontero:
good find @rajivnathan 

would adding a check for the `/plugins` path in `addUserContext` like that one that exists for the `/proxhealth` path over at https://github.com/codeready-toolchain/registration-service/blob/b60a175bd177d77949f8077c2c9e653a1d860767/pkg/proxy/proxy.go#L213-L215 be the means for bypassing this second call?

as to your question, I have no idea what about the request would make it redirect back to the proxy

that said, I'm not certain whether that is because it should not happen, or I'm too new to this stuff and I'm missing something :-) 

Pending any further finds from you, I may just try the `addUserContext` change I just mentioned ^^ when I can circle back to this later today.

rajivnathan:
@gabemontero 
>would adding a check for the /plugins path in addUserContext like that one that exists for the /proxhealth path be the means for bypassing this second call?

No, we still need to get the userID/username, so we can't just exclude the /plugins path in addUserContext. As I mentioned it's fine the first time, that's expected. What's unexpected is the 2nd call to it.

gabemontero:
> @gabemontero
> 
> > would adding a check for the /plugins path in addUserContext like that one that exists for the /proxhealth path be the means for bypassing this second call?
> 
> No, we still need to get the userID/username, so we can't just exclude the /plugins path in addUserContext. As I mentioned it's fine the first time, that's expected. What's unexpected is the 2nd call to it.

Right.  I've confirmed that with my initial experiment, and now see what you described with the first call being correct wrt the token.   I'm now playing around with identifying the 2nd call, see if dealing with it is possible.  If I get something that has a positive effect I'll post an update.  Likewise let me know if your attempts at removing addUserContext from the middleware leads to progress.

thanks

gabemontero:
see https://redhat-internal.slack.com/archives/CHK0J6HT6/p1678995853888069 for tl;dr but got the e2e working in my local dev / cluster for the day

let's see how it looks in CI

gabemontero:
passing e2e @alexeykazakov @rajivnathan

gabemontero:
minimally there is a debug commit to drop

gabemontero:
removed debug commit

rajivnathan (approved):
Looks like we finally made it! 🎉  Great work on the investigation and discovering the issue with the request and fixing it!

rajivnathan:
We'll get this merged first thing on Monday

gabemontero:
thanks @rajivnathan / good news!

MatousJobanek (approved):
Looks good, thanks a lot for your work :+1:

pkg/proxy/proxy.go
```diff
@@ -171,26 +173,64 @@ func (p *Proxy) handleRequestAndRedirect(ctx echo.Context) error {
...
+		}
+		prefixToTrim := ""
+		// NOTE: just in case the Split behavior is golang version dependent, let's make sure the first entry is empty
+		if len(segments[0]) == 0 {
+			prefixToTrim = fmt.Sprintf("/%s/%s", segments[1], segments[2])
+			proxyPluginName = segments[2]
+		} else {
+			prefixToTrim = fmt.Sprintf("/%s/%s", segments[0], segments[1])
+			proxyPluginName = segments[1]
+		}
```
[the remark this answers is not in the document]
gabemontero: it does not happen yet

I just find it odd that golang's function here works the way it does, with the first entry in the segment array being ""

so this is a defensive coding sort of thing, which I tried to convey in the comments above, if behavior ever changes

that said, I saw the same behavior in existing use of the golang call

so I can just remove at this point if you like ... just let me know

gabemontero:
updates pushed for both comments @MatousJobanek

pkg/proxy/proxy.go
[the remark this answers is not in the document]
MatousJobanek: cool thanks :sweat_smile: 
let's keep it simplified (as you just did) to not confuse other folks :smile: 
if the behavior changes, then the unit test should fail anyway

MatousJobanek (approved):

gabemontero:
the `Test with Coverage` failed with

`[2023-03-20T13:11:14.971Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}`

gabemontero:
ocp api server failure in https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/codeready-toolchain_registration-service/311/pull-ci-codeready-toolchain-registration-service-master-e2e/1637802334807396352 broke deployment of toolchains

/retest

gabemontero:
more ocp api svr issues https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/codeready-toolchain_registration-service/311/pull-ci-codeready-toolchain-registration-service-master-e2e/1637840745333264384

/retest

gabemontero:
more of the same infra pain:  ` from server for: "STDIN": Get "https://api.ci-op-sfp6tlpk-93ede.origin-ci-int-aws.dev.rhcloud.com:6443/apis/rbac.authorization.k8s.io/v1/clusterrolebindings/toolchain-member2-20172319-toolchain-member2-20172319": dial tcp: lookup api.ci-op-sfp6tlpk-93ede.origin-ci-int-aws.dev.rhcloud.com on 172.30.0.10:53: no such host `

I'm going to wait a bit before another retest

rajivnathan:
/retest
looks like more infra issues

alexeykazakov:
Infra issue again.

/retest

MatousJobanek:
/retest
infra

gabemontero:
some indication on ocp announce-testplatform that things may be back up

/retest

gabemontero:
past the ocp ci infra issues, all green e2e's @alexeykazakov @rajivnathan @MatousJobanek 

anything else you all want me to do before this gets lgtm and merged?

thanks

PR #312, opened 2023-03-30
Fix logger bug

rajivnathan:
I ran into an issue when trying to process a request with some contents in its body. The body was always empty but I could see in the browser the body had contents when it was sent. This was quite puzzling.

It turned out there's a bug in the logger where the request body in `ctx` is lost if logger.Info is called because when we read the body contents the `io.ReadCloser` is emptied. So later when the request is being processed `req.Body` is empty. The fix is to restore its contents after reading it in the logger.

Note: I tried making a copy of the request at the entrypoint of the logger.Info function but that didn't work because we'd still need to read from req.Body to copy it so it was effectively the same thing but a little more messy.

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"os"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/log/log.go
```diff
@@ -257,6 +258,9 @@ func addRequestInfo(req *http.Request) []interface{} {
 		}
 		fields = append(fields, "req_payload")
 		fields = append(fields, newStr)
+		// Once req.Body is read, it is empty. Restore its contents by
+		// assigning a new reader with the same contents.
```
xcoulon, line 262: 🙌 for putting on comment here, it's good to remember why we have this in the future 😅

PR #314, opened 2023-04-18
Add captcha to the signup flow

rajivnathan:
Related Issue: https://issues.redhat.com/browse/CRT-1767

Adds captcha functionality and configuration.

**Summary of changes**
- the main gate is the `IsPhoneVerificationRequired` func in the signup service, if there's any issue getting the captcha score or the score is below the configured threshold then it will return false and that will revert to the phone verification flow as it was (verification service remains unchanged)
- internal landing page updated to use captcha (it makes a call to get a captcha token to use in the signup request, this should work even if captcha is disabled)
- captcha is disabled by default, should only be enabled once configured correctly

**Unrelated changes**
- added `proxyplugins` to registration-service.yaml since it's in the host operator registration-service.yaml but was missing here


pkg/configuration/configuration_test.go
```diff
@@ -81,19 +86,25 @@ func TestRegistrationService(t *testing.T) {
 			Verification().AWSRegion("us-west-2").
 			Verification().AWSSenderID("sandbox").
 			Verification().AWSSMSType("Transactional").
+			Verification().CaptchaEnabled(true).
+			Verification().CaptchaProjectID("test-project").
+			Verification().CaptchaSiteKey("site-key").
+			Verification().CaptchaScoreThreshold("0.7").
```
alexeykazakov, line 92: Let's also test the default threshold: when it's not set then we use the default "0.9".

pkg/assets/static/index.html
```diff
@@ -5,6 +5,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1" />
   <link rel="stylesheet" href="https://unpkg.com/@patternfly/patternfly@4.6.3/patternfly.css" crossorigin="anonymous">
   <link rel="stylesheet" href="landingpage.css">
+  <script src="https://www.google.com/recaptcha/enterprise.js?render=6LcaESslAAAAAO1aPnbNiP1fg6z9VC5Y33Rq-Sl4"></script>
```
alexeykazakov, line 8: I'm curious what is that `6LcaESslAAAAAO1aPnbNiP1fg6z9VC5Y33Rq-Sl4`? is it some kind of our account/client ID or something?

pkg/assets/static/landingpage.js
```diff
@@ -213,15 +220,15 @@ function updateSignupState() {
 }
 
 function stopPolling() {
-  console.log('start polling..');
+  console.log('stop polling..');
```
alexeykazakov, line 223: lol

pkg/assets/static/landingpage.js
```diff
@@ -213,15 +220,15 @@ function updateSignupState() {
...
   if (intervalRef) {
     clearInterval(intervalRef);
     intervalRef = undefined;
   }
 }
 
 function startPolling() {
-  console.log('stop polling..');
+  console.log('start polling..');
```
alexeykazakov, line 231: rofl

pkg/signup/service/signup_service.go
```diff
@@ -83,11 +94,10 @@ func (s *ServiceImpl) newUserSignup(ctx *gin.Context) (*toolchainv1alpha1.UserSi
...
 	}
 
-	cfg := configuration.GetRegistrationServiceConfig()
-
-	verificationRequired := cfg.Verification().Enabled()
+	verificationRequired, captchaScore := IsPhoneVerificationRequired(s.CaptchaChecker, ctx)
 
 	// Check if the user's email address is in the list of domains excluded for phone verification
+	cfg := configuration.GetRegistrationServiceConfig()
 	emailHost := extractEmailHost(userEmail)
```
alexeykazakov, line 101: Do we have to do the captcha call if the email domain is excluded from verification?
What do you think about setting the verificatioRequired to false if the email domain is excluded and do not call `IsPhoneVerificationRequired()` at all in this case? The captcha annotation wouldn't be set in this case but does it matter if the verification is not required anyway?

But if it's tricky to organize the code this way or can potentially break anything I'm fine with leaving it as is. It's a minor optimization anyway.

pkg/signup/service/signup_service.go
```diff
@@ -120,6 +130,11 @@ func (s *ServiceImpl) newUserSignup(ctx *gin.Context) (*toolchainv1alpha1.UserSi
 			OriginalSub:   ctx.GetString(context.OriginalSubKey),
 		},
 	}
+
+	if captchaScore > -1.0 {
```
alexeykazakov, line 134: Why `-1.0`? Is `-1.0` some magic captcha? What does captcha =< -1.0 mean?

alexeykazakov: Now I see how it works below. But let's add the comment to the `IsPhoneVerificationRequired()` stating that the function either returns the actual captcha score or -1 with explanation what -1 means.

pkg/signup/service/signup_service.go
```diff
@@ -131,6 +146,53 @@ func (s *ServiceImpl) newUserSignup(ctx *gin.Context) (*toolchainv1alpha1.UserSi
...
+	if !cfg.Verification().Enabled() {
+		return false, -1
+	}
+
+	if !cfg.Verification().CaptchaEnabled() {
+		return true, -1
+	}
+
+	if ctx == nil || ctx.Request == nil {
+		log.Error(ctx, nil, "no request in context")
```
alexeykazakov, line 172: This should never happen, right? I mean if the context is nil or there is no request then it's a bug in our code. If so then I would return an error and fail rather then log and ignore.

pkg/signup/service/signup_service.go
```diff
@@ -131,6 +146,53 @@ func (s *ServiceImpl) newUserSignup(ctx *gin.Context) (*toolchainv1alpha1.UserSi
...
+	}
+
+	score, err := captchaChecker.CompleteAssessment(ctx, cfg, captchaToken[0])
+	if err != nil {
+		log.Error(ctx, err, "signup assessment failed")
+		return true, -1
+	}
+
+	if score < cfg.Verification().CaptchaScoreThreshold() {
+		log.Error(ctx, fmt.Errorf("the risk analysis score did not meet the expected threshold"), "signup assessment failed")
```
alexeykazakov, line 189: It's not an error. Should be `log.Info()` instead if we want to log it. Plus the actual score and the threshold could be useful in the log message.

pkg/verification/captcha/captcha.go
```diff
@@ -0,0 +1,86 @@
...
+
+/*
+*
+* Creates an assessment to analyze the risk of a signup.
+*
+* @param ctx: The request context.
+* @param cfg: The Registration Service Configuration object.
+* @param token: The token obtained from the client on passing the reCAPTCHA Site Key.
+
+returns an error if the assessment failed due to error or the assessment score was below the threshold.
```
alexeykazakov, line 32: You don't actually check the threshold, do you? You just return the score as is and check it outside of this function, right?

alexeykazakov (commented):
Looks good overall! I have a few comments/questions though.

cmd/main.go
```diff
@@ -70,6 +70,18 @@ func main() {
...
 
+	if crtConfig.Verification().CaptchaEnabled() {
+		if err := createCaptchaFileFromSecret(crtConfig, cl); err != nil {
+			log.Error(nil, err, "failed to create captcha file")
+			panic("failed to create captcha file")
+		}
+
+		// set application credentials env var required for recaptcha client
+		if err := os.Setenv("GOOGLE_APPLICATION_CREDENTIALS", configuration.CaptchaFilePath); err != nil {
+			panic("cannot set captcha credentials")
```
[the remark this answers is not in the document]
rajivnathan: oops, thanks for catching that. fixed in bc4bdb31bc7da5a3328d9ce2e1d4c977186e7071

pkg/assets/static/index.html
rajivnathan: It's the recaptcha site-key ID. See https://cloud.google.com/recaptcha-enterprise/docs/create-key There will be different site keys for stage and prod but this is one we can use for dev/test. In our ToolchainConfig configuration it's the `Captcha.SiteKey`.

pkg/assets/static/landingpage.js
rajivnathan: I was very confused. I had to look at the stopPolling and startPolling at least 3 times because I was sure I was seeing something wrong. Actually I'm still not sure 😂

pkg/configuration/configuration.go
```diff
@@ -249,3 +257,30 @@ func (r VerificationConfig) AWSSenderID() string {
...
 }
+
+func (r VerificationConfig) CaptchaEnabled() bool {
+	return commonconfig.GetBool(r.c.Captcha.Enabled, false)
+}
+
+func (r VerificationConfig) CaptchaScoreThreshold() float32 {
+	threshold := commonconfig.GetString(r.c.Captcha.ScoreThreshold, "")
+	thresholdFloat, err := strconv.ParseFloat(threshold, 32)
+	if err != nil {
```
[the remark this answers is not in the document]
rajivnathan: Good point, done in 77bd4a002e602cd49baa4956191518a9d0bd5af5

pkg/configuration/configuration_test.go
rajivnathan: We test the default on line 65 above. Or do you mean something else?

pkg/signup/service/signup_service.go
rajivnathan: Good point, added a comment in c37ee1c084857f2f92731fe33948ab6dbc514b2e

pkg/signup/service/signup_service.go
rajivnathan: I added this for unit tests actually, the code fails before this if the context is nil. So I think we can leave it as it is or just remove it and adjust to unit tests to always set ctx to a non-nil value?

pkg/configuration/configuration_test.go
alexeykazakov: Yeah I missed that.

pkg/signup/service/signup_service.go
rajivnathan: Good point. Done in 900c1ee7cff3b6cf8288262469e6c3ce4354230c

pkg/verification/captcha/captcha.go
rajivnathan: Thanks for catching it!

pkg/signup/service/signup_service.go
alexeykazakov: What do you mean it fails before that? Where does it fail? I see you return "true" meaning the phone verification required and nothing fails. Or I'm missing something?

alexeykazakov (approved):
Looks really good! Just one very minor comment left but I'm OK with this as is too.

pkg/signup/service/signup_service.go
rajivnathan: What I meant is `IsPhoneVerificationRequired` is called here https://github.com/rajivnathan/registration-service/blob/94d46dd9ae1de2fd4839b3c8512618045d4ff6ed/pkg/signup/service/signup_service.go#L97 but even before it is called there's lots of places where we try to get things from `ctx` so it would fail much earlier https://github.com/rajivnathan/registration-service/blob/recaptcha/pkg/signup/service/signup_service.go#L70-L81. Basically, I don't think we really need to worry about ctx being nil but the ctx.Request nil check is still fine to have.

alexeykazakov: ok

pkg/configuration/configuration.go
```diff
@@ -42,6 +44,15 @@ const (
 	UnitTestsEnvironment = "unit-tests"
 )
 
+// captcha specific configuration
+const (
+	CaptchaSecretName = "captcha-config" // #nosec
```
mfrancisc, line 49: just curious - what is this used for ?

mfrancisc (approved):
Nice work! 👍 

just left one very small comment.

PR #317, opened 2023-04-27
delete impersonate headers

MatousJobanek:


mfrancisc (approved):

pkg/proxy/proxy_test.go
```diff
@@ -425,135 +425,170 @@ func (s *TestProxySuite) TestProxy() {
...
-								req, err := http.NewRequest(tc.ProxyRequestMethod, reqPath, nil)
-								require.NoError(s.T(), err)
-								require.NotNil(s.T(), req)
-
-								for hk, hv := range tc.ProxyRequestHeaders {
-									for _, v := range hv {
-										req.Header.Add(hk, v)
-									}
+						for _, firstHeader := range rejectedHeaders {
+							for _, additionalHeader := range rejectedHeaders {
```
rajivnathan, line 445: I'm confused, aren't `firstHeader` and `additionalHeader` the same here? Why do we add them twice?

rajivnathan: Talked to Matous, it's intentional so that we can test combinations of the rejectedHeaders

rajivnathan (approved):
Thanks!

PR #318, opened 2023-05-02
Remove headers as part of middleware

rajivnathan:
Moves the changes from https://github.com/codeready-toolchain/registration-service/pull/317 into middleware so that it applies more broadly.

### DIFF
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -93,6 +93,7 @@ func (p *Proxy) StartProxy() *http.Server {
 	// middleware before routing
 	router.Pre(
 		middleware.RemoveTrailingSlash(),
+		p.stripInvalidHeaders(),
 		p.addUserContext(), // get user information from token before handling request
 	)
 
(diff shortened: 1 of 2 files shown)


pkg/proxy/proxy.go
```diff
@@ -258,6 +259,19 @@ func (p *Proxy) addUserContext() echo.MiddlewareFunc {
 	}
 }
 
+func (p *Proxy) stripInvalidHeaders() echo.MiddlewareFunc {
+	return func(next echo.HandlerFunc) echo.HandlerFunc {
+		return func(ctx echo.Context) error {
+			for header := range ctx.Request().Header {
+				if strings.HasPrefix(header, "Impersonate-") {
+					ctx.Request().Header.Del(header)
```
alexeykazakov, line 267: Let's log this. In case we want to check in the future if anyone tried to set the impersonate headers ;)

pkg/proxy/proxy_test.go
```diff
@@ -427,6 +427,7 @@ func (s *TestProxySuite) TestProxy() {
 
 				rejectedHeaders := []headerToAdd{
 					{},
+					{"impersonate-user", "myvalue"},
 					{"Impersonate-User", "myvalue"},
 					{"Impersonate-Group", "developers"},
 					{"Impersonate-Group", "admins"},
```
alexeykazakov, line 433:
```suggestion
					{"Impersonate-gRoup", "admins"},
```

alexeykazakov (approved):
Looks good. Just one minor suggestion.

PR #320, opened 2023-05-03
remove TransformUsername from registration-service

ranakan19:
character-limit restriction introduced in TransformUsername in https://github.com/codeready-toolchain/toolchain-common/pull/309 breaks the functionality that limits crtadmin usersignups to be created from registration-service. 
Example :
Username : `johny-long-username-crtadmin` would be reduced to `johny-long-username` using TransformUsername, letting the usersignup with the actual username `johny-long-username-crtadmin` be created.

Since the registration-service only uses TransformUsername to check if the username has a suffix `crtadmin`, a new function is introduces which checks just that, and remove dependency on transformUsername.

### DIFF
--- a/pkg/signup/service/signup_service.go
+++ b/pkg/signup/service/signup_service.go
@@ -20,8 +20,6 @@ import (
 	"github.com/codeready-toolchain/toolchain-common/pkg/condition"
 	"github.com/codeready-toolchain/toolchain-common/pkg/hash"
 	"github.com/codeready-toolchain/toolchain-common/pkg/states"
-	"github.com/codeready-toolchain/toolchain-common/pkg/usersignup"
-
 	"github.com/gin-gonic/gin"
 	errs "github.com/pkg/errors"
 	apiv1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


mfrancisc (approved):

pkg/signup/service/signup_service.go
```diff
@@ -136,6 +133,11 @@ func (s *ServiceImpl) newUserSignup(ctx *gin.Context) (*toolchainv1alpha1.UserSi
 	return userSignup, nil
 }
 
+func isCRTAdmin(username string) bool {
+	newUsername := regexp.MustCompile("[^A-Za-z0-9]").ReplaceAllString(strings.Split(username, "@")[0], "-")
```
alexeykazakov, line 137: Why do we need to do this replacement?

ranakan19: yeah, its not needed to replace the special characters, but it doesn't harm to replace them. At first I just copied the first line of TransformUsername to get the string before @, but then kept the whole replace and regex for the sake of consistency.

alexeykazakov: OK. I see... Both `@` split and transformation are probably redundant in the context of `-crtadmin` username checks but yeah.. I'm fine with keeping it.

alexeykazakov (approved):

MatousJobanek (approved):

PR #327, opened 2023-06-13
disable http2 when Upgrade header is set to SPDY/3.1

MatousJobanek:
https://issues.redhat.com/browse/CRT-1765

e2e tests doesn't make any sense in this case because we use insecure connections for non-prod environments which automatically use http/1.1

However, when working on this, I noticed there is a room for an improvement in the sense of reusing the RoundTripper from the cached ToolchainCluster rest.Config. This could technically improve the unit testing and also reusing the certificates set for the member clusters. I'll do that in a separate PR


rajivnathan (approved):
Nice 🙌

PR #329, opened 2023-06-16
Add logging to record when a user's access token is missing claims

sbryzak:
This PR will log details of a user when their JWT token is missing either the user_id or account_id claims.

Fixes https://issues.redhat.com/browse/CRT-1804


pkg/middleware/middleware.go
```diff
@@ -66,6 +67,12 @@ func (m *JWTMiddleware) HandlerFunc() gin.HandlerFunc {
 			m.respondWithError(c, http.StatusUnauthorized, err.Error())
 			return
 		}
+
+		if token.UserID == "" || token.AccountID == "" {
+			logrus.Infof("Missing essential claims from token - [user_id:%s][account_id:%s] for user [%s], sub [%s]",
```
alexeykazakov, line 72: Can you use our logger instead? Same as you use in signup_service.go

alexeykazakov (commented):
Looks good overall but can we use our logger instead?

pkg/middleware/middleware.go
sbryzak: no problem, fixed

alexeykazakov (approved):

MatousJobanek (approved):

MatousJobanek:
@xcoulon could you please help with creating a query for Loki to filter/select these logs?

PR #333, opened 2023-06-22
Capture raw claims in log

sbryzak:
If the user_id or account_id claim isn't set, log the actual claims part of the token string received from RHSSO.


alexeykazakov (approved):
Looks good. We could also log the header segment just in case though.

PR #336, opened 2023-07-10
Update UserSignup on each request if UserID or AccountID annotations missing

sbryzak:
This PR introduces a mechanism by which the UserSignup resource will be checked each time it is loaded, and if it is missing either the userID or accountID annotations, and these values exist within the user's access token claims, then the UserSignup resource will be updated with the new values.

Fixes https://issues.redhat.com/browse/SANDBOX-281


pkg/signup/service/signup_service.go
```diff
@@ -359,6 +359,37 @@ func (s *ServiceImpl) DoGetSignup(provider ResourceProvider, userID, username st
...
+		if accountID != "" {
+			userSignup.Annotations[toolchainv1alpha1.SSOAccountIDAnnotationKey] = accountID
+			updateUserSignup = true
+		}
+	}
+
+	if updateUserSignup {
+		userSignup, err = s.UpdateUserSignup(userSignup)
+		if err != nil {
+			return nil, errs.Wrapf(err, "error when attempting to update UserSignup with updated userID or accountID: %s", userSignup.Status)
```
alexeykazakov, line 389: There is a chance that the update will return 409 here if the UserSignup was updated after we obtained it in the line 348. In this case we just need to retry instead of failing. I'm afraid that UI can start showing errors for these cases making our API less stable.

I see a few options of how we could mitigate it:
- Log 409 errors here but do not return an error. We will try to update the annotation on the next GET /signup call again.
- Use a poll loop to update the UserSignup with some short timeout. We could even do the update in a separate go routine but it could potentially cause issues in other places since we are increasing chances of getting 409 in other places...
- Combination of those two ^^^

sbryzak: I've added a loop around the fetch/compare/update operation on the UserSignup, which will retry if there is a failure in updating the UserSignup and give up and return an error after repeated failed attempts.

MatousJobanek: btw, there is already logic that does the same: https://github.com/codeready-toolchain/registration-service/blob/1c1f3bc60765ae8cf377ad9fdea1becd76740492/pkg/verification/service/verification_service.go#L163-L187

pkg/signup/service/signup_service.go
```diff
@@ -333,30 +333,83 @@ func (s *ServiceImpl) reactivateUserSignup(ctx *gin.Context, existing *toolchain
...
+		userSignup, err = s.UpdateUserSignup(userSignup)
+		if err == nil {
+			// If there was no error during update, then break out of the loop here
+			break
+		}
+
+		// If we've reached this point, it means there was an error while attempting to update the UserSignup.  The
+		// attempts counter will be incremented and the for-loop will repeat
+		attempts++
 	}
```
MatousJobanek, line 413: could you please reuse the already existing logic from:
https://github.com/codeready-toolchain/registration-service/blob/1c1f3bc60765ae8cf377ad9fdea1becd76740492/pkg/verification/service/verification_service.go#L426-L453

an example of how it is already used: https://github.com/codeready-toolchain/registration-service/blob/1c1f3bc60765ae8cf377ad9fdea1becd76740492/pkg/verification/service/verification_service.go#L163-L187

pkg/signup/service/signup_service.go
```diff
@@ -336,30 +336,84 @@ func (s *ServiceImpl) reactivateUserSignup(ctx *gin.Context, existing *toolchain
...
+	return s.DoGetSignup(ctx, s.Services().InformerService(), userID, username)
 }
 
-func (s *ServiceImpl) DoGetSignup(provider ResourceProvider, userID, username string) (*signup.Signup, error) {
-	// Retrieve UserSignup resource from the host cluster, using the specified UserID and username
-	userSignup, err := s.DoGetUserSignupFromIdentifier(provider, userID, username)
-	// If an error was returned, then return here
-	if err != nil {
-		if apierrors.IsNotFound(err) {
+func (s *ServiceImpl) DoGetSignup(ctx *gin.Context, provider ResourceProvider, userID, username string) (*signup.Signup, error) {
```
alexeykazakov, line 349: It doesn't have to be gin.Context, does it? Can it be just context.Context?

pkg/proxy/handlers/spacelister.go
```diff
@@ -60,7 +62,9 @@ func (s *SpaceLister) ListUserWorkspaces(ctx echo.Context) ([]toolchainv1alpha1.
 	userID, _ := ctx.Get(context.SubKey).(string)
 	username, _ := ctx.Get(context.UsernameKey).(string)
 
-	signup, err := s.GetSignupFunc(userID, username)
+	c, _ := gin.CreateTestContext(httptest.NewRecorder())
```
alexeykazakov, line 65: It doesn't look right to use a text context in runtime. Let's just make the context param optional.

pkg/signup/service/signup_service.go
sbryzak: I've copied the logic from the verification service and am now updating in a similar way.

pkg/signup/service/signup_service.go
sbryzak: The contexts don't seem to be compatible.

pkg/proxy/handlers/spacelister.go
sbryzak: Done.  This does mean though that if the context is nil the updating feature won't work.

pkg/signup/service/signup_service.go
```diff
@@ -336,28 +336,77 @@ func (s *ServiceImpl) reactivateUserSignup(ctx *gin.Context, existing *toolchain
...
+		// If an error was returned, then return here
+		if getError != nil {
+			if apierrors.IsNotFound(getError) {
+				return nil
+			}
+			return getError
+		}
+
+		// Otherwise if the returned userSignup is nil, return here also
+		if userSignup == nil || ctx == nil {
```
alexeykazakov, line 366: Check if the ctx == nil as the first thing in the `pollUpdateSignup()` function. If it's nil then no need to get the usersignup in line 356.

pkg/signup/service/signup_service.go
```diff
@@ -336,28 +336,77 @@ func (s *ServiceImpl) reactivateUserSignup(ctx *gin.Context, existing *toolchain
...
+		// values exist within the claims of the current user's Access Token then set the values in the UserSignup and update
+		// the resource.
+		userIDValue, userIDFound := userSignup.Annotations[toolchainv1alpha1.SSOUserIDAnnotationKey]
+		accountIDValue, accountIDFound := userSignup.Annotations[toolchainv1alpha1.SSOAccountIDAnnotationKey]
+
+		updateUserSignup := false
+
+		if !userIDFound || userIDValue == "" {
+			userID := ctx.GetString(context.UserIDKey)
+			if userID != "" {
```
alexeykazakov, line 380: And same here. In the very begging of the `pollUpdateSignup()` function check if the context is nil or both `ctx.GetString() for UserID and AccountID are empty strings then just move on without even trying to do anything. It will safe time in many cases.

pkg/signup/service/signup_service_test.go
```diff
@@ -247,8 +247,10 @@ func (s *TestSignupServiceSuite) TestGetSignupFailsWithNotFoundThenOtherError()
 		return nil, errors2.NewInternalError(errors.New("something quite unfortunate happened"), "something bad")
 	}
 
+	c, _ := gin.CreateTestContext(httptest.NewRecorder())
```
alexeykazakov, line 250: Let's use `nil` instead of the test context at least in some tests to make sure it doesn't fail if the context is nil. Or rather use it only in the tests which test the cases when the context is expected.

pkg/signup/service/signup_service_test.go
```diff
@@ -1889,6 +1910,99 @@ func (s *TestSignupServiceSuite) TestIsPhoneVerificationRequired() {
...
+		require.NoError(s.T(), err)
+
+		require.NotContains(s.T(), modified.Annotations, toolchainv1alpha1.SSOUserIDAnnotationKey)
+		require.NotContains(s.T(), modified.Annotations, toolchainv1alpha1.SSOAccountIDAnnotationKey)
+	})
+
+	s.Run("userID annotation updated when set in context", func() {
+		c, _ := gin.CreateTestContext(httptest.NewRecorder())
+		c.Set(context.UserIDKey, "888888")
+		//c.Set(context.AccountIDKey, "1234567890")
```
alexeykazakov, line 1959:
```suggestion
```

pkg/signup/service/signup_service_test.go
```diff
@@ -1889,6 +1910,99 @@ func (s *TestSignupServiceSuite) TestIsPhoneVerificationRequired() {
...
+		require.NoError(s.T(), err)
+
+		// Confirm that the userID annotation wasn't updated
+		require.Equal(s.T(), "888888", modified.Annotations[toolchainv1alpha1.SSOUserIDAnnotationKey])
+		require.Equal(s.T(), "1234567890", modified.Annotations[toolchainv1alpha1.SSOAccountIDAnnotationKey])
+	})
+
+	s.Run("userID and accountID annotations not overridden when already set and context values different", func() {
+		c, _ := gin.CreateTestContext(httptest.NewRecorder())
+		// Set the userID context value to empty string
```
alexeykazakov, line 1990: Copy-paste :)

pkg/signup/service/signup_service_test.go
```diff
@@ -1889,6 +1910,99 @@ func (s *TestSignupServiceSuite) TestIsPhoneVerificationRequired() {
...
+		c.Set(context.UserIDKey, "7777777")
+		c.Set(context.AccountIDKey, "0987654321")
+
+		_, err := s.Application.SignupService().GetSignup(c, userSignup.Name, userSignup.Spec.Username)
+		require.NoError(s.T(), err)
+
+		modified, err := s.FakeUserSignupClient.Get(userSignup.Name)
+		require.NoError(s.T(), err)
+
+		// Confirm that both annotations are updated
```
alexeykazakov, line 2000:
```suggestion
		// Confirm that both annotations are NOT updated
```

pkg/signup/service/signup_service_test.go
```diff
@@ -1889,6 +1910,99 @@ func (s *TestSignupServiceSuite) TestIsPhoneVerificationRequired() {
...
+		require.Equal(s.T(), "888888", modified.Annotations[toolchainv1alpha1.SSOUserIDAnnotationKey])
+		require.Equal(s.T(), "1234567890", modified.Annotations[toolchainv1alpha1.SSOAccountIDAnnotationKey])
+	})
```
alexeykazakov, line 2003: We also need to cover the following cases:
- both userID and accountID are present in the context but they already match what in the annotations. There shouldn't be any affect on the usersignup but that test would trigger the code we need to make sure doesn't throws any errors.
- no context present (it's nil). The annotations are not changed.
- the cases to test error handling. When getting a user signup returns an error. When updating the usersignup returns an error.
- the case when the usersignup is not found (it's not an error). In case it's not already covered by a different test.

pkg/signup/service/signup_service.go
sbryzak: We still need the UserSignup even after the update function.

pkg/signup/service/signup_service.go
sbryzak: Same as above, we still need to load the UserSignup.

pkg/signup/service/signup_service.go
alexeykazakov: Ah, right.

pkg/signup/service/signup_service_test.go
sbryzak: Fixed

pkg/signup/service/signup_service_test.go
sbryzak: Fixed

pkg/signup/service/signup_service_test.go
sbryzak: Fixed

pkg/signup/service/signup_service_test.go
sbryzak: I've added a specific test just for this case.

pkg/signup/service/signup_service_test.go
sbryzak: Added tests for the first 2 items, and for testing error handling during an update.  we have existing tests that cover error handling for a get request, plus the not found case.

pkg/signup/service/signup_service.go
```diff
@@ -572,3 +621,30 @@ func getAppsURL(appRouteName string, signup signup.Signup) string {
 	appsURL := signup.ConsoleURL[index:]
 	return fmt.Sprintf("https://%s%s", appRouteName, appsURL)
 }
+
+func pollUpdateSignup(ctx *gin.Context, updater func() error) error {
+	// Attempt to execute an update function, retrying a number of times if the update fails
+	attempts := 0
+	for {
+		attempts++
```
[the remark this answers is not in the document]
sbryzak: i guess the only times when a request will fail will be either when there is a network connectivity issue (unlikely since registration service is running within the same cluster) or when the UserSignup is somehow updated elsewhere between registration service loading it and attempting the update.  In the second case, a wait() won't really help here as a simple reload and re-attempt to update will be sufficient without any delays required.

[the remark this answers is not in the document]
alexeykazakov: OK. Good point.

pkg/signup/service/signup_service_test.go
```diff
@@ -1889,6 +1910,139 @@ func (s *TestSignupServiceSuite) TestIsPhoneVerificationRequired() {
...
+	require.NoError(s.T(), err)
+
+	s.Run("confirm nothing changed when context nil", func() {
+		_, err := s.Application.SignupService().GetSignup(nil, userSignup.Name, userSignup.Spec.Username)
+		require.NoError(s.T(), err)
+
+		modified, err := s.FakeUserSignupClient.Get(userSignup.Name)
+		require.NoError(s.T(), err)
+
+		require.NotContains(s.T(), modified.Annotations, toolchainv1alpha1.SSOUserIDAnnotationKey)
```
alexeykazakov, line 1951: Can we test it with some annotation present? So we can make sure the annotations are not actually changed instead of being deleted.

alexeykazakov (commented):
Looks good. Just one minor comment.

pkg/signup/service/signup_service_test.go
sbryzak: I've changed the test ordering so that the nil context test now occurs after one of the annotations are actually set, to confirm that it's not changed.

alexeykazakov (approved):
Thank you for addressing my comments!

PR #340, opened 2023-07-25
Moved PollUpdateSignup function to common package

sbryzak:
This PR simply consolidates the two duplicate `pollUpdateSignup` functions from the signup and verification services into a single place where it is reused by both.


pkg/verification/service/verification_service.go
```diff
@@ -4,6 +4,7 @@ import (
 	"crypto/rand"
 	"errors"
 	"fmt"
+	signup2 "github.com/codeready-toolchain/registration-service/pkg/signup"
```
alexeykazakov, line 7: `signup2` looks a bit ugly :)

alexeykazakov (approved):

pkg/verification/service/verification_service.go
sbryzak: fixed

mfrancisc (approved):

PR #341, opened 2023-08-23
Added support for sender id configuration per country

sbryzak:
This PR modifies registration service to allow sender IDs to be configured for individual country codes.

Fixes https://issues.redhat.com/browse/SANDBOX-323

Tests have been conducted manually as they required live Twilio keys.

image


pkg/verification/sender/amazon_sns_sender.go
```diff
@@ -27,7 +27,7 @@ func NewAmazonSNSSender(cfg AWSSenderConfiguration) NotificationSender {
 	}
 }
 
-func (s *AmazonSNSSender) SendNotification(_ *gin.Context, content, phoneNumber string) error {
+func (s *AmazonSNSSender) SendNotification(_ *gin.Context, content, phoneNumber, _ string) error {
```
alexeykazakov, line 30: Maybe add a TODO note to implement it for AWS?

alexeykazakov (commented):
The code looks fine, however I don't see any actual unit test for this code. Am I missing something?

pkg/verification/sender/amazon_sns_sender.go
sbryzak: added TODO

sbryzak:
Unit test added

pkg/verification/sender/twilio_sender_test.go
```diff
@@ -0,0 +1,93 @@
...
+			{
+				SenderID:     "RED HAT",
+				CountryCodes: []string{"44"},
+			},
+		},
+	}
+
+	sender := sender2.NewTwilioSender(cfg, httpClient)
+
+	err := sender.SendNotification(ctx, "Test Message", "+440000000000", "44")
```
alexeykazakov, line 78: Can we also add a test which would send to a number which doesn't have the country code representation in the config? Basically another notification to any country code but "44". To check it doesn't use the sender ID and not just fail.

alexeykazakov (commented):
Thank you for the test! Looks good. But could you add one more test? See below.

pkg/verification/sender/twilio_sender_test.go
sbryzak: Done

alexeykazakov (approved):
Thanks for adding the test. The tests are 70% identical so you could use sub-tests to reuse the common part (twillio configuration and gock setup) but it's minor.

sbryzak:
I've added subtests and moved the common stuff to a reusable function.

PR #346, opened 2023-09-11
bugfix: proxy rejects calls when UserSignup is not in a ready state attempt 3

mfrancisc:
Disable usersignup ready condition check also for `ListUserWorkspaces` handler.

See discussion: https://redhat-internal.slack.com/archives/CHK0J6HT6/p1694437075698399?thread_ts=1694433405.899989&cid=CHK0J6HT6


pkg/proxy/handlers/spacelister.go
```diff
@@ -61,12 +61,13 @@ func (s *SpaceLister) ListUserWorkspaces(ctx echo.Context) ([]toolchainv1alpha1.
...
-	signup, err := s.GetSignupFunc(nil, userID, username, true)
+
+	signup, err := s.GetSignupFunc(nil, userID, username, false)
 	if err != nil {
 		ctx.Logger().Error(errs.Wrap(err, "error retrieving signup"))
 		return nil, err
 	}
-	if signup == nil || !signup.Status.Ready {
+	if signup == nil || signup.CompliantUsername == "" {
 		// account exists but is not ready so return an empty list
```
alexeykazakov, line 71:
```suggestion
		// account exists but the compliant username is not set yet, meaning it has not been fully provisioned yet, so return an empty list
```

alexeykazakov (approved):

PR #348, opened 2023-09-20
feat: Implement public API that will open SpaceBindings to end users through proxy part 1

mfrancisc:
- create `GetUserWorkspace` to accommodate required changes for populating the newly added `workspace.Status.AvailableRoles` and `workspace.Status.Bindings` fields
- populate  `workspace.Spec.AvailableRoles`  for `GetUserWorkspace`

Follow up PRs will take care of populating also the`workspace.Status.Bindings` field.

Note: we could create separate handlers for _get workspace_ and _list workspaces_, but this will require more work on separating the test cases. Let's defer this for later maybe.

Jira story: https://issues.redhat.com/browse/ASC-410

Related PRs: 
- https://github.com/codeready-toolchain/toolchain-common/pull/327

Paired with:
- https://github.com/codeready-toolchain/toolchain-e2e/pull/797


pkg/proxy/handlers/spacelister.go
```diff
@@ -37,26 +39,85 @@ func NewSpaceLister(app application.Application) *SpaceLister {
 
 func (s *SpaceLister) HandleSpaceListRequest(ctx echo.Context) error {
```
alexeykazakov, line 40: I still think that having two separate handlers and use routes (https://github.com/codeready-toolchain/registration-service/blob/master/pkg/proxy/proxy.go#L127-L130) would make the code cleaner. This is why we have routes in the first place.

pkg/informers/informers.go
```diff
@@ -22,6 +22,7 @@ type Informer struct {
 	ToolchainStatus   cache.GenericLister
 	UserSignup        cache.GenericLister
 	ProxyPluginConfig cache.GenericLister
+	NSTemplateTier    cache.GenericLister
```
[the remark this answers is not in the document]
mfrancisc: good catch, I missed that. Thanks!

it's done now, plz see: https://github.com/codeready-toolchain/registration-service/pull/348/commits/c3639a39e3a2d367253f90650751c5581f7da4ac

pkg/proxy/handlers/spacelister.go
mfrancisc: I agree, and I can do that split. But as I briefly explained in the PR description this will introduce more changes (at least to the unit tests), and make the PR bigger. If this is not a concern then I'm fine with adding those changes to the current PR.

alexeykazakov: It's totally fine to split your work in multiple PR which compliment each other. For example one PR which introduces an Informer. Another one which adds a new handler for Get. And one more for adding a role.
What I would avoid is splitting your work in the way that the following PR has to refactor the code you introduced in the previous PRs. Like adding everything in one handler and then another PR which refactors the handler and the code you just added. You see what I mean?

mfrancisc: yeah, that was a better way of splitting the PRs. My PR now has the new nstemplatetier informer and available roles, the refactor for adding the new handler for GET will be "minimal" in the handler code, it will be more unit tests separation from what I can see. 

My idea was to:
1) introduce the new available roles (current PR)
2) introduce the list of bindings (at this point there will be some code duplication probably the code in host operator for listing all spacebindings, see: https://github.com/codeready-toolchain/host-operator/blob/c68ff269240fd984839a19fb6743288f31381d1b/controllers/space/space_controller.go#L260-L313)

at this point we delivered the functionality to RHTAP, and we can focus on refactoring if needed:

3) refactor the spacebindings listing and extract common logic to toolchain-common in order use it in both host-operator and registration-service
4) refactor the existing List handler and create a Get handler

The main idea was to first deliver the functionality (and small improvements), and then making the existing code better once RHTAP is using it already.  
But of course I'm open to suggestions.

mfrancisc:
I saw this unit test failure also locally: https://github.com/codeready-toolchain/registration-service/actions/runs/6253097120/job/16977766459?pr=348
```
--- FAIL: TestRunServerSuite (0.43s)
    --- FAIL: TestRunServerSuite/TestServer (0.42s)
        server_test.go:116: 
            	Error Trace:	/home/runner/work/registration-service/registration-service/pkg/server/server_test.go:116
            	            				/opt/hostedtoolcache/go/1.19.13/x64/src/runtime/asm_amd64.s:[159](https://github.com/codeready-toolchain/registration-service/actions/runs/6253097120/job/16977766459?pr=348#step:4:160)4
            	Error:      	Received unexpected error:
            	            	listen tcp :8081: bind: address already in use
            	Test:       	TestRunServerSuite/TestServer
FAIL
```

but I thought that was caused by some other process on my machine using the same port 🤔

alexeykazakov:
> but I thought that was caused by some other process on my machine using the same port

It's a known flaky unit test. We do start a real server in our unit tests to test some parts of the proxy. Never actually tried to fix it... Just restarting the job should fix it.

pkg/proxy/handlers/spacelister.go
alexeykazakov: Maybe I'm missing something but starting with two handlers instead of adding everything to the single one doesn't really make the implementation harder... But again, maybe I'm missing something :)

mfrancisc: No it doesn't make the implementation harder , it's just some more changes to review :). If this is not a concern, I'll add  the changes for the new handler in the current PR.

pkg/proxy/handlers/spacelister.go
```diff
@@ -81,68 +150,98 @@ func (s *SpaceLister) ListUserWorkspaces(ctx echo.Context) ([]toolchainv1alpha1.
...
 		}
+		workspace := createCommonWorkspace(signupName, space, spacebinding)
+		workspaces = append(workspaces, *workspace)
+	}
+	return workspaces
+}
 
-		// TODO right now we get SpaceCreatorLabelKey but should get owner from Space once it's implemented
-		ownerName = space.Labels[toolchainv1alpha1.SpaceCreatorLabelKey]
+func createCommonWorkspace(signupName string, space *toolchainv1alpha1.Space, spaceBinding *toolchainv1alpha1.SpaceBinding, wsAdditionalOptions ...commonproxy.WorkspaceOption) *toolchainv1alpha1.Workspace {
```
alexeykazakov, line 206: Maybe something like `createWorkspaceObject()` would be a better name?

pkg/proxy/handlers/spacelister.go
```diff
@@ -81,68 +150,98 @@ func (s *SpaceLister) ListUserWorkspaces(ctx echo.Context) ([]toolchainv1alpha1.
...
+	return s.GetInformerServiceFunc().GetSpace(spaceName)
+}
+
+func getRolesFromNSTemplateTier(nstemplatetier *toolchainv1alpha1.NSTemplateTier) []string {
+	roles := make([]string, len(nstemplatetier.Spec.SpaceRoles))
+	i := 0
+	for k := range nstemplatetier.Spec.SpaceRoles {
+		roles[i] = k
+		i++
+	}
```
alexeykazakov, line 242: Shouldn't it be this instead:
```suggestion
	for i, k := range nstemplatetier.Spec.SpaceRoles {
		roles[i] = k
	}
```

alexeykazakov (commented):
Looks good! Just a couple of comments.

pkg/proxy/handlers/spacelister.go
mfrancisc: Well, actually I now realized why I had to do it this way. The `SpaceRoles` indexes are strings (name of the role) and not integers.

pkg/proxy/handlers/spacelister.go
mfrancisc: yeah, I was unsure about a proper name (as always 😅 ). thanks for the suggestion! 

Applied: https://github.com/codeready-toolchain/registration-service/pull/348/commits/e83efa484c4ca06b4053e640407cb188ae1e7fdb

mfrancisc:
/retest

reg-service should not crash anymore

mfrancisc:
/retest

really strange , from the regservice seems like it's still crashing because of the missing `nstemplatetiers` role missing: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/codeready-toolchain_registration-service/348/pull-ci-codeready-toolchain-registration-service-master-e2e/1706271468867817472/artifacts/e2e/test/artifacts/registration-service.log

even if I've already merged that [host-operator PR](https://github.com/codeready-toolchain/host-operator/pull/866) . 

Maybe I'm still missing something? 🤔

pkg/proxy/handlers/spacelister.go
alexeykazakov: Ah! It's a map. Not an array.

mfrancisc: correct!

alexeykazakov (approved):
Looks good! But the title is a bit misleading. This PR doesn't expose SpaceBindings yet. It only adds the list of available roles for now, correct?

mfrancisc:
@alexeykazakov thanks for taking a look and approval. 

you're right , the title its a bit misleading. As I tried to explain in the description its just introducing the available roles and new handler for her workspaces. Next PRs will add the spacebindings list. I'll make sure to be more accurate with the PR titles.

pkg/proxy/handlers/spacelister.go
```diff
@@ -81,68 +149,98 @@ func (s *SpaceLister) ListUserWorkspaces(ctx echo.Context) ([]toolchainv1alpha1.
...
 
-	if doGetWorkspace {
-		// specific workspace requested so add label requirement to match the space
-		spaceSelector, err := labels.NewRequirement(toolchainv1alpha1.SpaceBindingSpaceLabelKey, selection.Equals, []string{workspaceName})
-		if err != nil {
-			return nil, err
-		}
-		requirements = append(requirements, *spaceSelector)
+	//  let's only log the issue and consider this as not found
+	if len(spaceBindings) != 1 {
```
MatousJobanek, line 180: we could change that so when 
`len(spaceBindings) == 0` then we would return not found
and if `len(spaceBindings) > 1` then we would return "internal error"

pkg/proxy/handlers/spacelister.go
```diff
@@ -36,25 +37,92 @@ func NewSpaceLister(app application.Application) *SpaceLister {
...
+		// not found
+		r := schema.GroupResource{Group: "toolchain.dev.openshift.com", Resource: "workspaces"}
+		return errorResponse(ctx, apierrors.NewNotFound(r, workspaceName))
+	}
+	return getWorkspaceResponse(ctx, workspace)
+}
+
+func (s *SpaceLister) GetUserWorkspace(ctx echo.Context, signup *signup.Signup) (*toolchainv1alpha1.Workspace, error) {
+	murName := signup.CompliantUsername
+	spaceBinding, err := s.listSpaceBindingForUserAndSpace(ctx, murName)
```
MatousJobanek, line 95: we should keep in mind the Space inheritance - if a user tries to get information about sub-space, then no SpaceBinding will be found. The logic needs to go up in the tree of the parents until it finds at least one.
It's fine to do it in the following PR :+1:

pkg/proxy/handlers/spacelister.go
```diff
@@ -81,68 +149,98 @@ func (s *SpaceLister) ListUserWorkspaces(ctx echo.Context) ([]toolchainv1alpha1.
...
+	return s.GetInformerServiceFunc().GetSpace(spaceName)
+}
+
+func getRolesFromNSTemplateTier(nstemplatetier *toolchainv1alpha1.NSTemplateTier) []string {
+	roles := make([]string, len(nstemplatetier.Spec.SpaceRoles))
+	i := 0
+	for k := range nstemplatetier.Spec.SpaceRoles {
+		roles[i] = k
+		i++
+	}
```
MatousJobanek, line 241: easier to read:
```suggestion
	roles := make([]string, 0, len(nstemplatetier.Spec.SpaceRoles))
	for k := range nstemplatetier.Spec.SpaceRoles {
		roles = append(roles, k)
	}
```

pkg/proxy/handlers/spacelister.go
```diff
@@ -36,25 +37,92 @@ func NewSpaceLister(app application.Application) *SpaceLister {
...
+		// account exists but the compliant username is not set yet, meaning it has not been fully provisioned yet
+		// return not found response when specific workspace request was issued
+		r := schema.GroupResource{Group: "toolchain.dev.openshift.com", Resource: "workspaces"}
+		return errorResponse(ctx, apierrors.NewNotFound(r, workspaceName))
 
-	return listWorkspaceResponse(ctx, workspaces)
+	}
+
+	// get specific workspace
+	workspace, err := s.GetUserWorkspace(ctx, userSignup)
```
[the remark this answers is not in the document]
mfrancisc: Both handlers `HandleSpaceListRequest` and the newly added `HandleSpaceGetRequest` need to retrieve the usersignup in order to get the `signup.CompliantUsername` to be used as MUR name for listing the spacebindings. The _get_ and the _list_ are now in different handlers so both of them need to retrieve the usersignup, but maybe I'm missing something.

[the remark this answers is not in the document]
MatousJobanek: I'm idiot - I selected a wrong code and copy-pasted a wrong method name when doing the comment. Sorry about that. I got lost in the changes between IDE and GH.
I wanted to comment on 
`HandleSpaceListRequest` part:
https://github.com/codeready-toolchain/registration-service/blob/1305bcf1a4583893eb6a7415ba998abc95fad2d6/pkg/proxy/handlers/spacelister.go#L39-L55

you check the UserSignup there and then call the `ListUserWorkspaces` method which does the same again:
https://github.com/codeready-toolchain/registration-service/blob/1305bcf1a4583893eb6a7415ba998abc95fad2d6/pkg/proxy/handlers/spacelister.go#L128-L140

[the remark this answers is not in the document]
mfrancisc: Oh no worries, I see now the issue 🤦‍♂️ . I haven't noticed I duplicated this code while refactoring , thanks for pointing that out! Let me fix it!

pkg/proxy/handlers/spacelister.go
mfrancisc: Good point! thanks!

pkg/proxy/handlers/spacelister.go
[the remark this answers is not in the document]
mfrancisc: I've moved the usersignup retrieveal inside the `ListUserWorkspaces` and `GetUserWorkspace` functions, so that they are the same. I removed that from the handler functions, mainly because the `ListUserWorkspaces` function is used also by the proxy which doesn't have the usersignup available to be passed as parameter.

Plz see: https://github.com/codeready-toolchain/registration-service/pull/348/commits/01759250a7329b5df84baeca99ea984bd28029ec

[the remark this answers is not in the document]
mfrancisc: And I've also added sub tests in HandleSpaceGetRequest for the "get signup error" and "signup has no compliant username set"

pkg/proxy/handlers/spacelister.go
mfrancisc: Good point! Done with tests: https://github.com/codeready-toolchain/registration-service/pull/348/commits/01759250a7329b5df84baeca99ea984bd28029ec

pkg/proxy/handlers/spacelister.go
mfrancisc: Yeah, good point. Let me do that in the next PR. Thanks!

mfrancisc:
@MatousJobanek thanks for your review. I've addressed your comments, PTAL when you have some time!

MatousJobanek (approved):
Thanks for addressing my comments :heart: :rocket:

MatousJobanek (approved):
merge looks good :+1:

mfrancisc:
@MatousJobanek thanks for double checking 🙏

PR #350, opened 2023-10-02
feat: Implement public API that will open SpaceBindings to end users through proxy part 2

mfrancisc:
Populate `workspace.Status.Bindings` with the list of usernames, roles and available actions on those usernames.

This set of PRs will cover all the remaining work for jira: https://issues.redhat.com/browse/ASC-410

Some followup refactor work will remain only. 

Related PRs:
toolchain-common: https://github.com/codeready-toolchain/toolchain-common/pull/329

Paired with:
toolchain-e2e: https://github.com/codeready-toolchain/toolchain-e2e/pull/805


pkg/proxy/handlers/spacelister.go
```diff
@@ -84,38 +96,57 @@ func (s *SpaceLister) GetUserWorkspace(ctx echo.Context) (*toolchainv1alpha1.Wor
...
-	// -------------
-	// TODO recursively get all the spacebindings for the current workspace
-	// and build the Bindings list with the available actions
+	// check if user has access to this workspace
+	userBindings := filterUserSpaceBindings(userSignup.CompliantUsername, allSpaceBindings)
+	if len(userBindings) == 0 {
+		//  let's only log the issue and consider this as not found
+		ctx.Logger().Error(fmt.Sprintf("expected only 1 spacebinding, got 0 for user %s and workspace %s", userSignup.CompliantUsername, workspaceName))
+		return nil, nil
+	} else if len(userBindings) > 1 {
```
alexeykazakov, line 129: Couldn't a user have multiple SpaceBindings in a hierarchy of the spaces? Even if we don't really support it now we could support it in the future, right? Do we really need to make sure there is no than one here?

alexeykazakov: BTW, aren't we already checking that user has access to the workspace? Do we need to do this additional SpaceBinding check?

mfrancisc: > Couldn't a user have multiple SpaceBindings in a hierarchy of the spaces? Even if we don't really support it now we could support it in the future, right? Do we really need to make sure there is no than one here?

We could avoid the check, but right now:

`allSpaceBindings, err := spaceBindingLister.ListForSpace(space, []toolchainv1alpha1.SpaceBinding{})` at line 117 

should return only 1 spacebinding for the given Space and MUR (since it merges them based on the MUR name). So that's why I'm checking that there is no more than one returned from there.

mfrancisc: > BTW, aren't we already checking that user has access to the workspace? Do we need to do this additional SpaceBinding check?

before those changes there was this function here checking for the binding: https://github.com/codeready-toolchain/registration-service/pull/350/files#diff-cffbabd2ac8378a758970ae247dbee7b644f1d94e702181d836d5d5d19a048c5L88

I've replaced that with `filterUserSpaceBindings` which searches for the user binding from the list of all bindings that has already fetched above using the new `spaceBindingLister.ListForSpace` .

I don't see any other place where we check if user has access to the workspace 🤷‍♂️  , but maybe I'm missing something.

alexeykazakov: > before those changes there was this function here checking for the binding:

Ah.. OK. I missed that. So you are replacing the current check with the new one.

alexeykazakov (approved):
Looks good. Thanks.

pkg/proxy/handlers/spacelister.go
```diff
@@ -265,3 +264,43 @@ func errorResponse(ctx echo.Context, err *apierrors.StatusError) error {
...
+// The Bindings list  has the available actions for each entry in the list.
+func generateWorkspaceBindings(space *toolchainv1alpha1.Space, spaceBindings []toolchainv1alpha1.SpaceBinding) []toolchainv1alpha1.Binding {
+	var bindings []toolchainv1alpha1.Binding
+	for _, spaceBinding := range spaceBindings {
+		binding := toolchainv1alpha1.Binding{
+			MasterUserRecord: spaceBinding.Spec.MasterUserRecord,
+			Role:             spaceBinding.Spec.SpaceRole,
+			AvailableActions: []string{},
+		}
+		if name, exists := spaceBinding.GetLabels()[toolchainv1alpha1.SpaceBindingSpaceLabelKey]; exists && name == space.GetName() {
```
MatousJobanek, line 288: We should check the SBR labels instead - if the SpaceBinding was created by the UserSignup controller (as the default SpaceBinding) then the users cannot do anything with it. They can only update/delete those that have corresponding SBR CR in the namespace.

pkg/proxy/handlers/spacelister.go
```diff
@@ -265,3 +264,43 @@ func errorResponse(ctx echo.Context, err *apierrors.StatusError) error {
...
+			// this is a binding that was created on the current space, it can be updated or deleted
+			binding.AvailableActions = []string{UpdateBindingAction, DeleteBindingAction}
```
MatousJobanek, line 290: how does the client (UI) know which SBR it should either update or delete?
I think that the API is missing the `bindingRequest` field, right?
https://docs.google.com/document/d/1hWaFvCfsEXrbqs8ndDR0JPjKROXYJkvj1ZCnio4P1WU/edit#bookmark=id.tttpu2pbe4a6

pkg/proxy/handlers/spacelister.go
```diff
@@ -26,6 +28,15 @@ import (
 	"k8s.io/apimachinery/pkg/selection"
 )
 
+const (
+	// UpdateBindingAction specifies that the current binding can be updated by providing a different Space Role.
+	UpdateBindingAction = "update"
+	// DeleteBindingAction specifies that the current binding can be deleted in order to revoke user access to the Space.
+	DeleteBindingAction = "delete"
+	// OverrideBindingAction specifies that the current binding can be overridden by creating a SpaceBinding containing the same MUR but different Space Role.
```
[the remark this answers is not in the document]
mfrancisc: Thanks!

pkg/proxy/handlers/spacelister.go
```diff
@@ -265,3 +264,43 @@ func errorResponse(ctx echo.Context, err *apierrors.StatusError) error {
...
+
+// filterUserSpaceBindings returns all the spacebindings for a given username
+func filterUserSpaceBindings(username string, allSpaceBindings []toolchainv1alpha1.SpaceBinding) (outputBindings []toolchainv1alpha1.SpaceBinding) {
+	for _, binding := range allSpaceBindings {
+		if binding.Spec.MasterUserRecord == username {
+			outputBindings = append(outputBindings, binding)
+		}
+	}
+	return outputBindings
+}
```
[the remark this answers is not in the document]
mfrancisc: yes this makes it easier! 

I think we would just not be able to report in case there are issues with the system , generating more spacebindings for the same MUR (let's say for whatever reason, a bug in Space controller or SBR controller or other parts of the system like the sandbox-cli ... ). If this is acceptable then yes, the code can be simplified.

pkg/proxy/handlers/spacelister.go
mfrancisc: Ohh jeez 🤦‍♂️ , I missed that field in the original API pr and totally forgot about it! Great catch! 

I've raised https://github.com/codeready-toolchain/api/pull/378 for this. And I'm updating this PR and e2e tests to use it! 

Thanks!

pkg/proxy/handlers/spacelister.go
mfrancisc: Good point, I think the logic was incorrect here 🤦‍♂️ 

I'm now checking for SBR label on the SB resource. Plz take a look mainly at: https://github.com/codeready-toolchain/registration-service/pull/350/commits/64ba7bab696d459bc4b170749685a926b15b6f0e

Thanks!

pkg/proxy/handlers/spacelister.go
mfrancisc: I'm now populating the `bindingRequest` field as well. PTAL at https://github.com/codeready-toolchain/registration-service/pull/350/commits/64ba7bab696d459bc4b170749685a926b15b6f0e

pkg/proxy/handlers/spacelister.go
mfrancisc: There were actually a few cases missing:
- the case in which there should be no available action, since the spacebinding was generated by the system or system admin
- spacebinding was generated from SBR but on the parentSpace, so in this case it can only be overridden , and not updated/deleted as SBRs generated on the current space.

Please check also  [485a52e](https://github.com/codeready-toolchain/registration-service/pull/350/commits/485a52eb2119a0cf3414b82dd4726f8ace6fdf1c)

pkg/proxy/handlers/spacelister.go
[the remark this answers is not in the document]
mfrancisc: as decided earlier today, let's not return this error to the user. I'm not checking only for the user binding as you suggested. Plz check https://github.com/codeready-toolchain/registration-service/pull/350/commits/e8eee83db6691d4923771242f81bf01d7ac41e37

pkg/proxy/handlers/spacelister.go
```diff
@@ -84,38 +96,56 @@ func (s *SpaceLister) GetUserWorkspace(ctx echo.Context) (*toolchainv1alpha1.Wor
...
+		spaceSelector, err := labels.NewRequirement(toolchainv1alpha1.SpaceBindingSpaceLabelKey, selection.Equals, []string{spaceName})
+		if err != nil {
+			return nil, err
+		}
+		return s.GetInformerServiceFunc().ListSpaceBindings(*spaceSelector)
+	}
+	getSpaceFunc := func(spaceName string) (*toolchainv1alpha1.Space, error) {
+		return s.GetInformerServiceFunc().GetSpace(spaceName)
+	}
+	spaceBindingLister := spacebinding.NewLister(listSpaceBindingsFunc, getSpaceFunc)
```
MatousJobanek, line 116: it's just a detail, but the `GetSpace` method already complies with the `getSpaceFunc` format 
```suggestion
	spaceBindingLister := spacebinding.NewLister(listSpaceBindingsFunc, s.GetInformerServiceFunc().GetSpace)
```

pkg/proxy/handlers/spacelister.go
```diff
@@ -84,38 +96,56 @@ func (s *SpaceLister) GetUserWorkspace(ctx echo.Context) (*toolchainv1alpha1.Wor
...
 	}
 
-	// -------------
-	// TODO recursively get all the spacebindings for the current workspace
-	// and build the Bindings list with the available actions
+	// check if user has access to this workspace
+	userBinding := filterUserSpaceBinding(userSignup.CompliantUsername, allSpaceBindings)
+	if userBinding == nil {
+		//  let's only log the issue and consider this as not found
+		ctx.Logger().Error(fmt.Sprintf("expected 1 spacebinding, got 0 for user %s and workspace %s", userSignup.CompliantUsername, workspaceName))
```
MatousJobanek, line 127:
```suggestion
		ctx.Logger().Error(fmt.Sprintf("unauthorized access - there is no SpaceBinding present for the user %s and the workspace %s", userSignup.CompliantUsername, workspaceName))
```

MatousJobanek (approved):
Looks great, thanks a lot for addressing my comments :+1: :1st_place_medal: :rocket:

pkg/proxy/handlers/spacelister.go
mfrancisc: Good point! Thanks for the suggestion!

pkg/proxy/handlers/spacelister.go
mfrancisc: this looks better ! thanks !

PR #351, opened 2023-10-05
allow for approved status false after user sign up deactivated

mmulholla:
Fix required for:  https://github.com/codeready-toolchain/host-operator/pull/868

it should not cause a regression because this condition:
```approvedCondition.Reason != toolchainv1alpha1.UserSignupUserDeactivatedReason```
can only be true without https://github.com/codeready-toolchain/host-operator/pull/868

This is the issue it fixes as a result of https://github.com/codeready-toolchain/host-operator/pull/868

```
=== NAME  TestSignupOK/test_activation-deactivation_workflow
    host.go:711: waiting for UserSignup 'testuser-e808fd7d-bf83-4cdc-abe8-1b72c80c1a9e' in namespace 'toolchain-host-05094958' to match criteria
    host.go:1982: waiting until Space 'testuser-e808fd7d-bf' in namespace 'toolchain-host-05094958' is deleted
    host.go:2028: waiting until SpaceBindings with labels 'map[toolchain.dev.openshift.com/space:testuser-e808fd7d-bf]' in namespace 'toolchain-host-05094958' are deleted
    regsvc.go:16: invoking http request: GET https://registration-service-toolchain-host-05094958.apps.martin-4.12-092615.qe.openshiftappsvc.org/api/v1/signup
    regsvc.go:26: response status code: 200
    regsvc.go:33: 
           Error Trace:   /Users/martinmulholland/codeready-toolchain/toolchain-e2e/testsupport/regsvc.go:33
                                   /Users/martinmulholland/codeready-toolchain/toolchain-e2e/test/e2e/parallel/registration_service_test.go:805
                                   /Users/martinmulholland/codeready-toolchain/toolchain-e2e/test/e2e/parallel/registration_service_test.go:381
           Error:         Not equal: 
                          expected: 404
                          actual  : 200
           Test:          TestSignupOK/test_activation-deactivation_workflow
           Messages:      unexpected response status with body: {"name":"testuser-e808fd7d-bf83-4cdc-abe8-1b72c80c1a9e","compliantUsername":"testuser-e808fd7d-bf","username":"testuser-e808fd7d-bf83-4cdc-abe8-1b72c80c1a9e","givenName":"","familyName":"","company":"","status":{"ready":false,"reason":"PendingApproval","verificationRequired":false}}
```

### DIFF
--- a/pkg/signup/service/signup_service.go
+++ b/pkg/signup/service/signup_service.go
@@ -438,7 +438,8 @@ func (s *ServiceImpl) DoGetSignup(ctx *gin.Context, provider ResourceProvider, u
 	// Check UserSignup status to determine whether user signup is complete
 	approvedCondition, approvedFound := condition.FindConditionByType(userSignup.Status.Conditions, toolchainv1alpha1.UserSignupApproved)
 	completeCondition, completeFound := condition.FindConditionByType(userSignup.Status.Conditions, toolchainv1alpha1.UserSignupComplete)
-	if !approvedFound || !completeFound || approvedCondition.Status != apiv1.ConditionTrue {
+	if !approvedFound || !completeFound ||
+		( approvedCondition.Status != apiv1.ConditionTrue && approvedCondition.Reason != toolchainv1alpha1.UserSignupUserDeactivatedReason) {
 		log.Info(nil, fmt.Sprintf("usersignup: %s is pending approval", userSignup.GetName()))
 		signupResponse.Status = signup.Status{
 			Reason:               toolchainv1alpha1.UserSignupPendingApprovalReason,
--- a/pkg/signup/service/signup_service.go
+++ b/pkg/signup/service/signup_service.go
@@ -436,11 +436,13 @@ func (s *ServiceImpl) DoGetSignup(ctx *gin.Context, provider ResourceProvider, u
 	}
 
 	// Check UserSignup status to determine whether user signup is complete
-	approvedCondition, approvedFound := condition.FindConditionByType(userSignup.Status.Conditions, toolchainv1alpha1.UserSignupApproved)
+	_, approvedFound := condition.FindConditionByType(userSignup.Status.Conditions, toolchainv1alpha1.UserSignupApproved)
 	completeCondition, completeFound := condition.FindConditionByType(userSignup.Status.Conditions, toolchainv1alpha1.UserSignupComplete)
 	if !approvedFound || !completeFound ||
-		(approvedCondition.Status != apiv1.ConditionTrue && approvedCondition.Reason != toolchainv1alpha1.UserSignupUserDeactivatedReason) {
+		condition.IsFalseWithReason(userSignup.Status.Conditions,
+			toolchainv1alpha1.UserSignupApproved, toolchainv1alpha1.UserSignupPendingApprovalReason) {
 		log.Info(nil, fmt.Sprintf("usersignup: %s is pending approval", userSignup.GetName()))
+
 		signupResponse.Status = signup.Status{
 			Reason:               toolchainv1alpha1.UserSignupPendingApprovalReason,
 			VerificationRequired: states.VerificationRequired(userSignup),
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/signup/service/signup_service.go [hunk 1]
```diff
@@ -438,7 +438,8 @@ func (s *ServiceImpl) DoGetSignup(ctx *gin.Context, provider ResourceProvider, u
 	// Check UserSignup status to determine whether user signup is complete
 	approvedCondition, approvedFound := condition.FindConditionByType(userSignup.Status.Conditions, toolchainv1alpha1.UserSignupApproved)
 	completeCondition, completeFound := condition.FindConditionByType(userSignup.Status.Conditions, toolchainv1alpha1.UserSignupComplete)
-	if !approvedFound || !completeFound || approvedCondition.Status != apiv1.ConditionTrue {
+	if !approvedFound || !completeFound ||
+		( approvedCondition.Status != apiv1.ConditionTrue && approvedCondition.Reason != toolchainv1alpha1.UserSignupUserDeactivatedReason) {
```
alexeykazakov, line 442: I think we should replace it by:
```suggestion
		condition.IsFalseWithReason(userSignup.Status.Conditions, toolchainv1alpha1.UserSignupApproved, toolchainv1alpha1.UserSignupPendingApprovalReason {
```
This is the actual condition we set for UserSignup when it's pending approval:
```
  - reason: PendingApproval
    status: "False"
    type: Approved
```

alexeykazakov (commented):
I think the unit tests should be updated to.

pkg/signup/service/signup_service.go
mmulholla: I can do that if you prefer but my concern would be that is is more strict than the current code which could lead to other behavior changes.

alexeykazakov: Yeah, I still think it's better to rely on the `PendingApproval` reason.

pkg/signup/service/signup_service.go
```diff
@@ -438,7 +438,8 @@ func (s *ServiceImpl) DoGetSignup(ctx *gin.Context, provider ResourceProvider, u
 	// Check UserSignup status to determine whether user signup is complete
 	approvedCondition, approvedFound := condition.FindConditionByType(userSignup.Status.Conditions, toolchainv1alpha1.UserSignupApproved)
 	completeCondition, completeFound := condition.FindConditionByType(userSignup.Status.Conditions, toolchainv1alpha1.UserSignupComplete)
-	if !approvedFound || !completeFound || approvedCondition.Status != apiv1.ConditionTrue {
+	if !approvedFound || !completeFound ||
+		(approvedCondition.Status != apiv1.ConditionTrue && approvedCondition.Reason != toolchainv1alpha1.UserSignupUserDeactivatedReason) {
```
MatousJobanek, line 442: this would mean that if the approved is found, but the status is not true (is pending approval), then the if statement would be evaluated as "false", which would mean (in this context) as approved, which is wrong, right?
We want to mark the UserSignup as pending approval when:
1. approved condition is either not found or the status is false
2. or the complete condition is not found (it's almost the same as the previous prerequisite, but let's keep it there to be 100% sure)UserSignupUserDeactivatedReason
3. and at the same time the reason in approved condition is not `UserSignupUserDeactivatedReason`

It shouldn't happen that the approved condition would have a reason  `UserSignupUserDeactivatedReason` and the complete condition wouldn't be found, so no need to deal with that situation.

Please try to play with the `condition.` package - you may found some useful functions there that could simplify it.

MatousJobanek: I'm idiot :man_facepalming: 
It's Monday and I cannot read the if statement correctly. Your version should work fine.
Could you please add/update some unit tests?

pkg/signup/service/signup_service.go
mmulholla: Updated

pkg/signup/service/signup_service.go
mmulholla: Have updated the tests.

mmulholla: Also changed the if statement per Alexey's comment .

pkg/signup/service/signup_service_test.go
```diff
@@ -1093,7 +1093,8 @@ func (s *TestSignupServiceSuite) TestGetSignupDeactivated() {
 	// given
 	s.ServiceConfiguration(configuration.Namespace(), true, "", 5)
 
-	us := s.newUserSignupCompleteWithReason("Deactivated")
+	us := s.newUserSignupComplete()
+	us.Status.Conditions = deactivated()
```
alexeykazakov, line 1097: Maybe I'm missing something but this test doesn't seem to check the `signup` payload status in lines 1132-1133. So it doesn't really test the changed signup service. Could you please add it to the test? Check the status.

mmulholla:
Had to rename the branch because it was picking up the toolchain-e2e branch in my fork which was causing failures but are needed for host-operator changes. New PR is: https://github.com/codeready-toolchain/registration-service/pull/354

PR #352, opened 2023-10-06
move proxy metrics to a separate server

ranakan19:
move proxy metrics to a separate server
host-operator PR - https://github.com/codeready-toolchain/host-operator/pull/884
paired e2e PR - https://github.com/codeready-toolchain/toolchain-e2e/pull/816


deploy/registration-service.yaml
```diff
@@ -98,6 +98,7 @@ objects:
               ports:
                 - containerPort: 8080
                 - containerPort: 8081
+                - containerPort: 8082
```
alexeykazakov, line 101: This file is used only in dev environments. The one which is used in production (and e2e tests) are defined here: https://github.com/codeready-toolchain/host-operator/blob/master/deploy/registration-service/registration-service.yaml
Please create a PR to update that file too.

But what I'm wonder is why the e2e-test still pass? Do we have any e2e tests for this metric service? Without updating https://github.com/codeready-toolchain/host-operator/blob/master/deploy/registration-service/registration-service.yaml the e2e-tests for this PR should fail...

alexeykazakov (commented):
We also need to expose the 8082 port in our container: https://github.com/codeready-toolchain/registration-service/blob/master/build/Dockerfile#L19

deploy/registration-service.yaml
xcoulon: could we also add names to the container ports (see for example https://kubernetes.io/docs/concepts/services-networking/service/#field-spec-ports) to make them more explicit, and reuse these names in the services as well?

MatousJobanek: > This file is used only in dev environments.

It's actually not used even in dev environment. This file is a legacy thing from the time when we were able to deploy reg-service on its own - we don't do that anymore (or at least I'm not aware of anyone who would be doing that)

MatousJobanek: but anyway, Alexey is right that you need to create PR in host-repo with this change. Let's do it in two steps:
1. create PR in host-operator repo changing only the values in the template (I don't think that we verify the actual content of the deployment in e2e tests)
3. when the host-operator PR is merged, create a PR in e2e tests that will be paired with this reg-service one that will verify that the `Service` is created.

pkg/metrics/metrics.go
```diff
@@ -58,9 +64,37 @@ func newHistogramVec(name, help string, labels ...string) *prometheus.HistogramV
...
+//nolint:unparam
+func prometheusHandler(ctx echo.Context) error {
+	h := promhttp.HandlerFor(Reg, promhttp.HandlerOpts{DisableCompression: true, Registry: Reg})
+	h.ServeHTTP(ctx.Response().Writer, ctx.Request())
+	return nil
+}
+
+func StartMetricsServer() *http.Server {
+	// start server
+	router := echo.New()
```
MatousJobanek, line 86: `router` looks like a copy-paste zombie from the proxy code :-)

pkg/metrics/metrics_test.go
```diff
@@ -64,6 +67,31 @@ func TestRegisterCustomMetrics(t *testing.T) {
...
+		rec := httptest.NewRecorder()
+		ctx := e.NewContext(req, rec)
+
+		//when
+		err := prometheusHandler(ctx)
+		require.NoError(t, err)
+		// then
+		// check the status code is what we expect, and the content-type
+		require.Equal(t, http.StatusOK, rec.Code)
+		require.Equal(t, "text/plain; version=0.0.4; charset=utf-8", rec.Header().Get("Content-Type"))
```
MatousJobanek, line 91: does it make sense to verify that the body also contains the actual registered metric? you know, just check that the name of the metric is there, no need to verify the actual values...

deploy/registration-service.yaml
ranakan19: Sorry for the late reply,
@MatousJobanek @alexeykazakov Thanks for your comments and catching the missing host-repo changes. I did have the changes in host-repo, that's how i was able to test it locally indeed, but had missed creating the PR on host-repo. I've updated the host-operator and e2e PR in the description.

pkg/metrics/metrics_test.go
ranakan19: removed

pkg/metrics/metrics.go
ranakan19: fixed

ranakan19:
/retest
host-operator PR merged

pkg/metrics/metrics_test.go
```diff
@@ -1,20 +1,25 @@
...
 	"time"
 )
 
 func TestHistogramVec(t *testing.T) {
 	// given
+	reg := prometheus.NewRegistry()
 	m := newHistogramVec("test_histogram_vec", "test histogram description", "status_code", "kube_verb")
 	getSuccess, getFailure, listSuccess, listFailure := getExpectedLabelPairs()
-	RegisterCustomMetrics()
+	//RegisterCustomMetrics()
```
alexeykazakov, line 21:
```suggestion
```

alexeykazakov (approved):
Looks good!

pkg/metrics/metrics_test.go
ranakan19: I always forget to use the "commit suggestion" feature 😅 
[94a372a](https://github.com/codeready-toolchain/registration-service/pull/352/commits/94a372ad764a62a01e8276e3456d31c9941b88d4) done

MatousJobanek (approved):
Looks nice. Thanks for the work @ranakan19 :+1: :rocket:

PR #354, opened 2023-10-09
Fix pending approval error

mmulholla:
This replaces PR: https://github.com/codeready-toolchain/registration-service/pull/351 - please see this PR for earlier comments.

### DIFF
--- a/pkg/signup/service/signup_service_test.go
+++ b/pkg/signup/service/signup_service_test.go
@@ -1093,7 +1093,8 @@ func (s *TestSignupServiceSuite) TestGetSignupDeactivated() {
 	// given
 	s.ServiceConfiguration(configuration.Namespace(), true, "", 5)
 
-	us := s.newUserSignupCompleteWithReason("Deactivated")
+	us := s.newUserSignupComplete()
+	us.Status.Conditions = deactivated()
 	err := s.FakeUserSignupClient.Tracker.Add(us)
 	require.NoError(s.T(), err)
 
@@ -2222,7 +2223,12 @@ func deactivated() []toolchainv1alpha1.Condition {
 		{
 			Type:   toolchainv1alpha1.UserSignupComplete,
 			Status: apiv1.ConditionTrue,
-			Reason: "Deactivated",
+			Reason: toolchainv1alpha1.UserSignupUserDeactivatedReason,
+		},
+		{
+			Type:   toolchainv1alpha1.UserSignupApproved,
+			Status: apiv1.ConditionFalse,
+			Reason: toolchainv1alpha1.UserSignupUserDeactivatedReason,
 		},
 	}
 }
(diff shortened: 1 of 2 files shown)


pkg/signup/service/signup_service_test.go
```diff
@@ -1093,7 +1093,8 @@ func (s *TestSignupServiceSuite) TestGetSignupDeactivated() {
 	// given
 	s.ServiceConfiguration(configuration.Namespace(), true, "", 5)
 
-	us := s.newUserSignupCompleteWithReason("Deactivated")
+	us := s.newUserSignupComplete()
+	us.Status.Conditions = deactivated()
```
alexeykazakov, line 1097: I'm copying my comment from the old PR:

Maybe I'm missing something but this test doesn't seem to check the signup payload status in lines 1132-1133. So it doesn't really test the changed signup service. Could you please add it to the test? Check the status.

mmulholla: This is not how I read the test, I now set the conditions for deactivated as they would be now as a result of my host operator changes. ie this condition is added: 
```
		{
			Type:   toolchainv1alpha1.UserSignupApproved,
			Status: apiv1.ConditionFalse,
			Reason: toolchainv1alpha1.UserSignupUserDeactivatedReason,
		}
```

Before the fix, with this condition present, DoGetSignup would then: 

```return signupResponse, nil```    

from here:  https://github.com/codeready-toolchain/registration-service/blob/b4cc745d3ba379250ba91eb1120acf80c87bcae2/pkg/signup/service/signup_service.go#L447

But now because of the fix  DoGetSignup now returns:

```return nil, nil```

from here: https://github.com/codeready-toolchain/registration-service/blob/43236b87b84f4b122f13b8967d09db6cc786b226/pkg/signup/service/signup_service.go#L465

therefore  by checking both responses are nil when the extra condition is present confirms the fix works.

Comment (mmulholla) [reply pkg/signup/service/signup_service_test.go:1097]: Oh, btw, thx for copying the comment over. My bad on branch names but I live an learn :-).

Comment (alexeykazakov) [reply pkg/signup/service/signup_service_test.go:1097]: You are right!
What actually is missing is a test which checks that a UserSignup with the `UserSignupPendingApprovalReason` is returning a Signup with the corresponding status.
Basically `doGetSignup()` for the following UserSignup
```
status:
  conditions:
  - reason: PendingApproval
    status: "False"
    type: Approved
  - status: "False"
    type: Complete
```
should return a `Signup` with the following status:
```
{"ready": "false", "reason": "PendingApproval"}
```
There is probably a test for such status already which we need to extend to cover that case.

mfrancisc: Maybe this test https://github.com/codeready-toolchain/registration-service/blob/b4cc745d3ba379250ba91eb1120acf80c87bcae2/pkg/signup/service/signup_service_test.go#L975 is verifying the condition that @alexeykazakov is suggesting.

mmulholla: I think the current tests already cover this:

https://github.com/codeready-toolchain/registration-service/blob/b4cc745d3ba379250ba91eb1120acf80c87bcae2/pkg/signup/service/signup_service_test.go#L980

https://github.com/codeready-toolchain/registration-service/blob/b4cc745d3ba379250ba91eb1120acf80c87bcae2/pkg/signup/service/signup_service_test.go#L989

alexeykazakov (approved):

PR #355, opened 2023-10-11
feat: create usernames response type

mfrancisc:
This PR defines the response for the API `/usernames/<username>`. 
The API implementation will be done as a follow up work.

Jira: https://issues.redhat.com/browse/ASC-411

See also : https://redhat-internal.slack.com/archives/CHK0J6HT6/p1696509583462699


pkg/usernames/usernames.go
```diff
@@ -0,0 +1,19 @@
+package usernames
+
+// Response represents the result of a search request done by using the MUR name or an email address.
+// The result will specify if a user exits in the system or not , or if there are multiple matches for the given search query
+type Response struct {
+	// For is the search string that was specified in the search request url. Example: /usernames/<search string>
+	For string `json:"for"`
```
mfrancisc, line 7: I used the original names for the design doc since maybe there is already an agreement with RHTAP on those.
But personally I would suggest to use `Query` or `Search` here.

pkg/usernames/usernames.go
```diff
@@ -0,0 +1,19 @@
...
+	// Valid can be true or false.
+	// 	true indicates that a match was found
+	// 	false indicates that no match was found
+	Valid string `json:"valid"`
```
mfrancisc, line 11: and alternative name could be `Found`

pkg/usernames/usernames.go
```diff
@@ -0,0 +1,19 @@
...
+	// Matched is a list of the username matches found using the given search query.
+	// There might be more than one match, for example when using email id for the search request.
+	Matched []match `json:"matched,omitempty"`
```
mfrancisc, line 14: Alternatively this can be `Results`, `Matches`, `QueryResults` or `SearchResults`

pkg/usernames/usernames.go
rajivnathan: The example indicates that <search string> is a path parameter so maybe it should be named according to that? eg. `UsernameParam`.

rajivnathan: That's assuming the search string is even needed in the response. It's kind of weird because the request contains the same information so not sure why it needs to be duplicated in the response but I'm guessing there was a reason that was discussed and agreed upon with RHTAP folks?

pkg/usernames/usernames.go
rajivnathan: Is this needed? Isn't found the same as when `Matched` contains some results and not found is the same as when `Matched` is empty?

Should this be more about validation of the search parameter instead?

pkg/usernames/usernames.go
mfrancisc: > That's assuming the search string is even needed in the response.

I was thinking about the same thing. Maybe we can avoid it, since it should be already known to the client/UI that is sending the request.

pkg/usernames/usernames.go
mfrancisc: The plan is In the first phase to support search only by username. So in this case the response will be just:
GET `/usernames/john`
```
{
  for: john
  valid: true
}
```
Later we will support search also by email id, and the response might return multiple results, the valid will be set to false in that case:

GET /usernames/john@redhat.com
```
{
  for: john@redhat.com
  valid: false
  matched: [
 	{ 
       username: johny
  	},
 	{ 
       username: john-2
  	},
 	...
  ]}
```

rajivnathan: ah ok, makes sense for it to be `Valid` rather than `Found` in that case

pkg/usernames/usernames.go
alexeykazakov: Keeping `for` in the struct would make sense if it was a Kube API. But for a custom REST API as this I don't see any value in it.

pkg/usernames/usernames.go
```diff
@@ -0,0 +1,19 @@
+package usernames
```
alexeykazakov, line 1: I know we have a mixture of plural and singular package names but my understanding is that in golang it's preferable to use singular package names.

pkg/usernames/usernames.go
alexeykazakov: I would just return 404 if nothing found. So we can drop `Valid/Found` completely.
And do we really need to add any sub-fields there at all? Can it be just an array?

`GET /usernames/john` - 200 OK:
```
{"john"}
```
`GET /usernames/johnnn` - 404 NOT FOUND

`GET /usernames/john@domain.com` - 200 OK:
```
{"johny", "johny"}
```

pkg/usernames/usernames.go
mfrancisc: I named it using the plural form mainly because the endpoint will be `/usernames/<username>` . But I agree, in Golang package names are in singular form usually . 

I can rename it.

alexeykazakov: Maybe we should name the struct as `Usernames`. Because this is the actual REST resource type name: `/usernames/...`

pkg/usernames/usernames.go
mfrancisc: Makes sense to me to use the http status code as an indicator for found/not found. 

But in your last example, should it be 200 or 404 (since there are multiple users found) ?
```
GET /usernames/john@domain.com - 404 OK:

{"johny", "johny"}
```

Also I would keep the field name: `{"username":"johny"}`, it gives us the option to add more fields and details about the users (let's imagine we are asked to add the family name or sso id or anything else).

pkg/usernames/usernames.go
mfrancisc: I haven't named the struct `Usernames` because the package is called already `Usernames` or even if we name it `Username` it will look like `username.Usernames` from a client perspective , which is not a good practice in Go.

While `username.Response` looks better IMO. But I'm open to suggestions.

pkg/usernames/usernames.go
alexeykazakov: > So the validation logic should be on the client side then? eg. if the api returns multiple results then the client displays an error to users?

I would say it's up to the client (UI) to decide what to do in this case. Showing an error that there are multiple users found or show all the options to the user so he/she can pick which username should be used.

alexeykazakov: > Also I would keep the field name: {"username":"johny"}, it gives us the option to add more fields and details about the users (let's imagine we are asked to add the family name or sso id or anything else).

We could add it as:
```
{
    {"username" : "johny"},
    {"username" : "john"}
}
```
but I don't think we will be able to add anything besides the username anytime soon. Due to PII concerns. It's unlikely that we will need that but it's possible :)
I can keep this API as simple as possible and introduce a v2 version of it in the future if we really need it. But on the other hand that variant with "username": is not a big deal to have it now either. So I don't have a strong opinion.

@MatousJobanek any thought?

pkg/usernames/usernames.go
MatousJobanek: Agree with that - it doesn't make much sense to keep it there. Didn't realize it when I put it into the doc. Thanks

pkg/usernames/usernames.go
MatousJobanek: I really like this approach. Let's do that. If there is no user with the name then return 404. 
```
GET /usernames/doesnotexist - 404 NotFound:
GET /usernames/john@domain.com - 404 NotFound:
```
If there is then return the list of the matching entries - right now only one:
```
GET /usernames/johny - 200 OK:
[
    {"username" : "johny"}
]
```
later when we support emails as well, we could return more of them:
```
GET /usernames/johny@rh.com - 200 OK:
[
    {"username" : "johny"},
    {"username" : "johny2"}
]
```
we could easily extend/play with this format in the future without introducing any breaking changes with the current API

MatousJobanek: cc @rohitkrai03

pkg/usernames/usernames.go
mfrancisc: renamed the package to `username` (singular form) . Plz see : [8f86a3c](https://github.com/codeready-toolchain/registration-service/pull/355/commits/8f86a3c62e557249046644ab2d19e521e172cb9d)

alexeykazakov (approved):

PR #359, opened 2023-10-13
Update UserSignup.Spec.IdentityClaims if missing or different to the user's token

sbryzak:
This PR introduces a mechanism for updating the `IdentityClaims` property values stored in `UserSignup.Spec` if they are missing or different to the values presented by a user's access token simply by having the user visit the registration service landing page.  This will assist in the migration of all users to the usage of `IdentityClaims` as the single source of truth for all token claim values.

Fixes https://issues.redhat.com/browse/SANDBOX-437

e2e test PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/821


pkg/signup/service/signup_service_test.go
```diff
@@ -2104,6 +2104,73 @@ func (s *TestSignupServiceSuite) TestGetSignupUpdatesUserSignupAnnotations() {
...
+		_, err := s.Application.SignupService().GetSignup(c, userSignup.Name, userSignup.Spec.Username)
+		require.NoError(s.T(), err)
+
+		modified, err := s.FakeUserSignupClient.Get(userSignup.Name)
+		require.NoError(s.T(), err)
+
+		require.Equal(s.T(), "Jonathan", modified.Spec.IdentityClaims.GivenName)
+
+		// Confirm that some other properties were not changed
+		require.Equal(s.T(), "cocochanel", modified.Spec.IdentityClaims.PreferredUsername)
```
alexeykazakov, line 2134: This test depends on the previous `"PreferredUsername property updated when set in context"` so maybe worth moving it under that test as a sub-test. But I guess it's fine to keep it this way too. Maybe worth adding a comment that it's coming from the previous test.

alexeykazakov (approved):
Looks good!

pkg/signup/service/signup_service_test.go
sbryzak: good catch, i'll move it under the other test

PR #366, opened 2023-10-26
fix: proxy first request slowness issue 

mfrancisc:
The slowness was caused by the member cache being populated only with the first request. 
I'm moving the the cluster cache initialization before starting the services , so that the first request will have it already populated.

Jira: https://issues.redhat.com/plugins/servlet/samlsso?redirectTo=%2Fbrowse%2FSANDBOX-449


pkg/controller/health_check.go
```diff
@@ -84,9 +83,5 @@ func (c *healthCheckerImpl) APIProxyAlive(ctx *gin.Context) bool {
 		return false
 	}
 	defer resp.Body.Close()
-	_, err = io.ReadAll(resp.Body)
```
xcoulon, line 87: do we still need the line above if we don't read the body at all?

xcoulon (approved):
Looks good! Just left a comment on the healthcheck change

pkg/controller/health_check.go
rajivnathan: @mfrancisc what is the memory consumption issue with reading the body for the health endpoint?

[The documentation](https://pkg.go.dev/net/http#Client.Get) for http.Get() says:
> When err is nil, resp always contains a non-nil resp.Body. Caller should close resp.Body when done reading from it.

So unless there's some reason to do otherwise, we should probably follow the documentation. If there is a performance concern, I found a [blog post](https://manishrjain.com/must-close-golang-http-response#:~:text=Now%2C%20it’s%20still%20possible%20that%20you%20forget%20to%20close%20the%20body%20somewhere,Instead%2C%20only%20use%20a%20client%20that%20you%20created) that mentioned a workaround that we can investigate.

rajivnathan (commented):
Excellent investigation and fix! 🥇 Just have a question about the health endpoint change.

pkg/controller/health_check.go
rajivnathan: Maybe I am misunderstanding the documentation? I understood it as it's optimal to read the body before closing it but I saw some debate about whether reading the body is actually required. I guess that it's not required but there may be benefits to doing it.

This PR is old but indicates reading the body enables reusing the TCP connection https://github.com/google/go-github/pull/317

We read the body before closing in the auth key manager too: https://github.com/codeready-toolchain/registration-service/blob/9ce1f44f6e48665508583f43e312bfbd215d2d2d/pkg/auth/keymanager.go#L154C1-L163

mfrancisc: The memory profile I've ran was showing a big bubble (high memory consumption) caused by the `io.ReadAll` function (even if the json is small). But this was not part of the slowness fix, just to be clear, so it's not mandatory. I can remove there are concerns.

from the doc https://pkg.go.dev/net/http#Client.Get , it mentions that you must close the resp , but not that is also mandatory to read it. The drawback with not reading it is with the persistent TCP connection that might not be reused:

> If the Body is not both read to EOF and closed, the Client's underlying RoundTripper (typically Transport) may not be able to re-use a persistent TCP connection to the server for a subsequent "keep-alive" request.

mfrancisc: > do we still need the line above if we don't read the body at all?
yes closing the response body seems mandatory

mfrancisc: restored, please see: https://github.com/codeready-toolchain/registration-service/pull/366/commits/bd401abfa43419ee9b7dd6e0c08896d68e06bad8

we can address this separately in case we want to have a look at the memory profile as well, since there seems to be other places where we are doing the same thing as @rajivnathan pointed out.

alexeykazakov (approved):

pkg/controller/health_check.go
rajivnathan: Another option we can consider is using a HEAD request instead of a GET request and then I believe there is no response body, which sounds good enough for the health endpoint.

rajivnathan (approved):

pkg/controller/health_check.go
mfrancisc: good idea @rajivnathan !

PR #368, opened 2023-11-08
Removing the unused Parameters

fbm3307:
This is to drop unused parameters from the embedded MUR.Status.UserAccounts.Clusters type.

Related PRs
API - https://github.com/codeready-toolchain/api/pull/387
Host-Operator - https://github.com/codeready-toolchain/host-operator/pull/925

### DIFF
--- a/pkg/signup/service/signup_service.go
+++ b/pkg/signup/service/signup_service.go
@@ -474,9 +474,6 @@ func (s *ServiceImpl) DoGetSignup(ctx *gin.Context, provider ResourceProvider, u
 		signupResponse.ProxyURL = status.Status.HostRoutes.ProxyURL
 		for _, member := range status.Status.Members {
 			if member.ClusterName == mur.Status.UserAccounts[0].Cluster.Name {
-				signupResponse.ConsoleURL = member.MemberStatus.Routes.ConsoleURL
-				signupResponse.CheDashboardURL = member.MemberStatus.Routes.CheDashboardURL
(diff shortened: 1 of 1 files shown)


pkg/signup/service/signup_service.go
```diff
@@ -474,9 +474,6 @@ func (s *ServiceImpl) DoGetSignup(ctx *gin.Context, provider ResourceProvider, u
 		signupResponse.ProxyURL = status.Status.HostRoutes.ProxyURL
 		for _, member := range status.Status.Members {
 			if member.ClusterName == mur.Status.UserAccounts[0].Cluster.Name {
-				signupResponse.ConsoleURL = member.MemberStatus.Routes.ConsoleURL
-				signupResponse.CheDashboardURL = member.MemberStatus.Routes.CheDashboardURL
-				signupResponse.APIEndpoint = member.APIEndpoint
```
MatousJobanek, line 479: this should stay there - these parameters are not from the MUR, but from MemberStatus.
In addition to that, we need to keep the signupResponse fields untouched - they are used by the clients.

We should only drop the fields from the MasterUserRecord.Status (as you did in the api PR) and make sure that these fields are not used anywhere. When you update the api dependency in the corresponding repos:
* host-operator
* registration-service
* toolchain-common
* toolchain-e2e

then you can easily check if the fields were used anywhere or not

fbm3307: Okay got it

fbm3307:
closing this PR as the changes are not required

PR #371, opened 2023-11-14
feat: Introduce SpaceBindingLister function in handleRequestAndRedirect

mfrancisc:
This PR : 
- adds the [spacebinding lister ](https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/spacebinding/spacebindinglister.go) to the `handleRequestAndRedirect` proxy handler. This enables users to issue requests for subSpaces.
- splits spacelister get and list handlers into separate files

TODO (in next PR):
 - split `spacelister_test.go` into `spacelister_get_test.go` and `spacelister_list_test.go`

Jira: https://issues.redhat.com/browse/ASC-437

### DIFF
--- a/pkg/proxy/handlers/spacelister.go
+++ b/pkg/proxy/handlers/spacelister.go
@@ -54,7 +54,7 @@ func NewSpaceLister(app application.Application, proxyMetrics *metrics.ProxyMetr
 func (s *SpaceLister) HandleSpaceListRequest(ctx echo.Context) error {
 	// list all user workspaces
 	requestReceivedTime := ctx.Get(context.RequestReceivedTime).(time.Time)
-	workspaces, err := s.ListUserWorkspaces(ctx)
+	workspaces, err := s.ListUserWorkspaces(ctx, "")
 	if err != nil {
 		s.ProxyMetrics.RegServWorkspaceHistogramVec.WithLabelValues(fmt.Sprintf("%d", http.StatusInternalServerError), metrics.MetricsLabelVerbList).Observe(time.Since(requestReceivedTime).Seconds()) // using list as the default value for verb to minimize label combinations for prometheus to process
 		return errorResponse(ctx, apierrors.NewInternalError(err))
@@ -147,7 +147,7 @@ func (s *SpaceLister) GetUserWorkspace(ctx echo.Context) (*toolchainv1alpha1.Wor
 	), nil
 }
 
-func (s *SpaceLister) ListUserWorkspaces(ctx echo.Context) ([]toolchainv1alpha1.Workspace, error) {
+func (s *SpaceLister) ListUserWorkspaces(ctx echo.Context, workspaceName string) ([]toolchainv1alpha1.Workspace, error) {
 	userID, _ := ctx.Get(context.SubKey).(string)
 	username, _ := ctx.Get(context.UsernameKey).(string)
 
@@ -160,15 +160,53 @@ func (s *SpaceLister) ListUserWorkspaces(ctx echo.Context) ([]toolchainv1alpha1.
 		// account exists but the compliant username is not set yet, meaning it has not been fully provisioned yet, so return an empty list
 		return []toolchainv1alpha1.Workspace{}, nil
 	}
-
 	murName := signup.CompliantUsername
+
+	if workspaceName != "" {
+		// if a workspace was provided
+		// recursively get all the spacebindings, starting from this workspace and going up to the parent workspaces till the "root" of the workspace tree.
+		return s.hierarchicalWorkspaceSearch(ctx, workspaceName, signup)
+	}
+
+	// get all spacebindings with given mur since no workspace was provided
 	spaceBindings, err := s.listSpaceBindingsForUser(murName)
 	if err != nil {
 		ctx.Logger().Error(errs.Wrap(err, "error listing space bindings"))
 		return nil, err
 	}
-
 	return s.workspacesFromSpaceBindings(signup.Name, spaceBindings), nil
+}
+
+// hierarchicalWorkspaceSearch lists all spacebindings starting from a space and going to it's parentSpaces
+// if any of those spaces contains a SpaceBinding for the given user, then a workspace object will be returned with the found role.
+func (s *SpaceLister) hierarchicalWorkspaceSearch(ctx echo.Context, workspaceName string, signup *signup.Signup) ([]toolchainv1alpha1.Workspace, error) {
+	listSpaceBindingsFunc := func(spaceName string) ([]toolchainv1alpha1.SpaceBinding, error) {
+		spaceSelector, err := labels.NewRequirement(toolchainv1alpha1.SpaceBindingSpaceLabelKey, selection.Equals, []string{spaceName})
+		if err != nil {
+			return nil, err
+		}
+		return s.GetInformerServiceFunc().ListSpaceBindings(*spaceSelector)
+	}
+	spaceBindingLister := spacebinding.NewLister(listSpaceBindingsFunc, s.GetInformerServiceFunc().GetSpace)
+	space, err := s.GetInformerServiceFunc().GetSpace(workspaceName)
+	if err != nil {
+		ctx.Logger().Error(err, "failed to get space")
+		return nil, err
+	}
+	allSpaceBindings, err := spaceBindingLister.ListForSpace(space, []toolchainv1alpha1.SpaceBinding{})
+	if err != nil {
+		ctx.Logger().Error(err, "failed to list space bindings")
+		return nil, err
+	}
+	// check if user has access to this workspace
+	murName := signup.CompliantUsername
+	userBinding := filterUserSpaceBinding(murName, allSpaceBindings)
+	if userBinding == nil {
+		//  let's only log the issue and consider this as not found
+		ctx.Logger().Error(fmt.Sprintf("unauthorized access - there is no SpaceBinding present for the user %s and the workspace %s", murName, workspaceName))
+		return nil, nil
+	}
+	return []toolchainv1alpha1.Workspace{*createWorkspaceObject(signup.Name, space, userBinding)}, nil
 }
 
 func (s *SpaceLister) listSpaceBindingsForUser(murName string) ([]toolchainv1alpha1.SpaceBinding, error) {
--- a/pkg/proxy/handlers/spacelister.go
+++ b/pkg/proxy/handlers/spacelister.go
@@ -2,30 +2,19 @@ package handlers
 
 import (
 	"encoding/json"
-	"fmt"
-	"net/http"
-	"sort"
-
-	"time"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 	"github.com/codeready-toolchain/registration-service/pkg/application"
 	"github.com/codeready-toolchain/registration-service/pkg/application/service"
 	"github.com/codeready-toolchain/registration-service/pkg/context"
-	"github.com/codeready-toolchain/registration-service/pkg/log"
 	"github.com/codeready-toolchain/registration-service/pkg/metrics"
 	"github.com/codeready-toolchain/registration-service/pkg/signup"
 	commonproxy "github.com/codeready-toolchain/toolchain-common/pkg/proxy"
-	"github.com/codeready-toolchain/toolchain-common/pkg/spacebinding"
 	"github.com/gin-gonic/gin"
 
 	"github.com/labstack/echo/v4"
 	errs "github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/selection"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


alexeykazakov (commented):
Can we add more tests for this?

- Create a hierarchy of spaces (not just parent - child but a bit more complicated parent - child - grandchild or something like this).
- Add some "noise" space with a role for the same user to make sure it doesn't included in the list or SBs for the desired space hierarchy.

Also it would be cool to cover the following cases:
- correct SBs are collected depending on what workspace you are listing for in the same hierarchy (parent, child or grandchild).
- the case when no SBs are found at all

Also, is there any e2e tests for this?

pkg/proxy/handlers/spacelister.go
```diff
@@ -147,7 +147,7 @@ func (s *SpaceLister) GetUserWorkspace(ctx echo.Context) (*toolchainv1alpha1.Wor
 	), nil
 }
 
-func (s *SpaceLister) ListUserWorkspaces(ctx echo.Context) ([]toolchainv1alpha1.Workspace, error) {
+func (s *SpaceLister) ListUserWorkspaces(ctx echo.Context, workspaceName string) ([]toolchainv1alpha1.Workspace, error) {
```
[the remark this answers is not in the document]
mfrancisc: Done in https://github.com/codeready-toolchain/registration-service/pull/371/commits/00137718fdab8bb0c265c6162db1814582339db5

Please let me know if this looks better.

MatousJobanek:
> Can we add more tests for this?

Talking about `proxy_test.go`. We discussed that we would cover basically the happy-path(s). 
The biggest issue with the unit test is the structure of them which is caused by the structure of the whole registration-service code. It's technically about mocking, mocking, and mocking, and when you look at all the mock, some of the tests almost don't verify the actual logic (we already saw some bugs that we missed because of that). I wish we could finally had time to rewrite the registration-service code into something that is easier to test and maintain.

That being said, I can imagine that covering the more complex structure and scenarios could be a challenge with the current structure and setup. Apart from that, all the advanced scenarios are already tested in the unit tests of the function in the toolchain-common repo. So we just need to verify that the function is being used when it is supposed to be used, I don't think that we need to replicate the same scenarios here as well.
In other words, I would test what really makes sense given the effort that it requires and delegate the more complicated scenarios to the methods/functions called in the proxy. 

If you are talking about the SpaceLister test, then yeah, covering the sub-spaces scenario is definitely needed. We could also cover the new `hierarchicalWorkspaceSearch` separately there.

> Also, is there any e2e tests for this?

I'm afraid that this is still not possible. We cannot create sub-spaces with any tier - the only tier that can be used is `appstudio-env` one, which doesn't grant any permissions to end users. We need to merge this PR first: https://github.com/codeready-toolchain/host-operator/pull/902 to be able to cover sub-spaces via proxy more properly.

mfrancisc:
> Can we add more tests for this?

I was looking into the `proxy_test.go` and indeed as @MatousJobanek said , it seems like it's testing headers and http responses , more then actual "permissions" on given spaces. 

I was thinking about creating a new unit test in `proxy_test.go` or `spacelister_test.go` which will invoke directly `ListUserWorkspaces` (the function that is being updated with this PR) . Instead of testing the `handleRequestAndRedirect` which may require extra mocking and setup without actually providing more value in kind of test coverage. But please let me know what do you think?

> Also, is there any e2e tests for this?

I was looking what actually I can do there, since as @MatousJobanek mentioned , the user has no permissions to the namespaces created via `appstudio-env`, it can be accessed only using the kubeconfig contained by the secret inside the namespace.

mfrancisc:
as for the unit test, @alexeykazakov @MatousJobanek could you please take a look at https://github.com/codeready-toolchain/registration-service/pull/371/commits/cd293e443dd6347b2b64456ce118a5e80213fd73 . Please let me know if this is something close to what you were thinking about and if something is still missing. Thank you!

alexeykazakov:
> as for the unit test, @alexeykazakov @MatousJobanek could you please take a look at https://github.com/codeready-toolchain/registration-service/commit/cd293e443dd6347b2b64456ce118a5e80213fd73 . Please let me know if this is something close to what you were thinking about and if something is still missing. Thank you!

This is exactly what I was looking for. Thank you!

pkg/proxy/handlers/spacelister.go
```diff
@@ -147,7 +147,10 @@ func (s *SpaceLister) GetUserWorkspace(ctx echo.Context) (*toolchainv1alpha1.Wor
 	), nil
 }
 
-func (s *SpaceLister) ListUserWorkspaces(ctx echo.Context) ([]toolchainv1alpha1.Workspace, error) {
+// ListUserWorkspaces returns a list of Workspaces for the current user.
+// If a workspaceName is provided, it means the request was made for a specific workspace, the function will check if the user has access to that workspace, if not it will return 404 Unauthorized.
+// If no workspaceName is provided, the function will list all SpaceBindings for the user and return all the workspaces found from this list.
```
alexeykazakov, line 152:
```suggestion
// If no workspaceName is provided (workspaceName is an empty sting), the function will list all SpaceBindings for the user and return all the workspaces found from this list.
```

alexeykazakov (approved):
Looks good. Thank you for addressing my comments.

pkg/proxy/handlers/spacelister.go
mfrancisc: thanks!

pkg/proxy/handlers/spacelister.go
```diff
@@ -160,17 +163,55 @@ func (s *SpaceLister) ListUserWorkspaces(ctx echo.Context) ([]toolchainv1alpha1.
...
 	murName := signup.CompliantUsername
+
+	if workspaceName != "" {
+		// if a workspace was provided
+		// recursively get all the spacebindings, starting from this workspace and going up to the parent workspaces till the "root" of the workspace tree.
+		return s.hierarchicalWorkspaceSearch(ctx, workspaceName, signup)
+	}
+
+	// get all spacebindings with given mur since no workspace was provided
 	spaceBindings, err := s.listSpaceBindingsForUser(murName)
```
MatousJobanek, line 175: it this case it should go "down" in the structure of sub-spaces and collect all sub-spaces the user's SpaceBinding is inherited for.
It can be done as a separate story ;-)

pkg/proxy/handlers/spacelister.go
```diff
@@ -160,17 +163,55 @@ func (s *SpaceLister) ListUserWorkspaces(ctx echo.Context) ([]toolchainv1alpha1.
...
+		return nil, err
+	}
+	// check if user has access to this workspace
+	murName := signup.CompliantUsername
+	userBinding := filterUserSpaceBinding(murName, allSpaceBindings)
+	if userBinding == nil {
+		//  let's only log the issue and consider this as not found
+		ctx.Logger().Error(fmt.Sprintf("unauthorized access - there is no SpaceBinding present for the user %s and the workspace %s", murName, workspaceName))
+		return nil, nil
+	}
```
MatousJobanek, line 211: is this really needed? the validation is already one in https://github.com/codeready-toolchain/registration-service/blob/3d66cb33031e62543a58332f27708fbf1c5e9394/pkg/proxy/proxy.go#L511-L524

pkg/proxy/handlers/spacelister.go
```diff
@@ -147,7 +147,10 @@ func (s *SpaceLister) GetUserWorkspace(ctx echo.Context) (*toolchainv1alpha1.Wor
 	), nil
 }
 
-func (s *SpaceLister) ListUserWorkspaces(ctx echo.Context) ([]toolchainv1alpha1.Workspace, error) {
+// ListUserWorkspaces returns a list of Workspaces for the current user.
+// If a workspaceName is provided, it means the request was made for a specific workspace, the function will check if the user has access to that workspace, if not it will return 404 Unauthorized.
+// If no workspaceName is provided (workspaceName is an empty sting), the function will list all SpaceBindings for the user and return all the workspaces found from this list.
+func (s *SpaceLister) ListUserWorkspaces(ctx echo.Context, workspaceName string) ([]toolchainv1alpha1.Workspace, error) {
```
MatousJobanek, line 153: TBH, it took me a while to understand the usage and the cases when the method is being used. It's most likely because it's mixing two  use-cases:
1. Get list of workspaces the user has access to - this is done via listing the SpaceBindings by the username
2. Validate access for the given workspace and get information about it - but this is done by collecting all SpaceBindings for the given space and all parent-spaces

in other words - two different solutions:
1. getting all (inherited) SpaceBindings of the Space and filtering them
2. getting all SpaceBindings of the User

at the same time there is also the other case that collects all SpaceBindings for the Space: https://github.com/codeready-toolchain/registration-service/blob/3d66cb33031e62543a58332f27708fbf1c5e9394/pkg/proxy/handlers/spacelister.go#L108-L128

I'm wondering if there is a better way to use/organize the code to easily follow the use-cases.

pkg/proxy/handlers/spacelister.go
mfrancisc: 🤔 
ok, this seems a bit different use case. I guess we could:
- get all spacebindings
- for each space from the spacebindings get the list of subSpaces (using the `toolchainv1alpha1.ParentSpaceLabelKey` label)

So yeah I think this is a separate task 😅 . Let me create a story for it.

pkg/proxy/handlers/spacelister.go
mfrancisc: I've changed things a bit in https://github.com/codeready-toolchain/registration-service/pull/371/commits/796bed8496eb74d47f5ec2a1896cf47a85c87c02. I'm now using `GetUserWorkspace` which does that validation , but I need that so the function can return only 1 workspace (instead of all the workspaces of the user) which makes it makes it easier to test since I can directly call `GetUserWorkspace` from the hierarchical unit test cases.

pkg/proxy/handlers/spacelister.go
mfrancisc: You're right! the code is a bit "messy". I've done some refactor, attempting to remove some of the duplication, and make those paths more clear.

In https://github.com/codeready-toolchain/registration-service/pull/371/commits/796bed8496eb74d47f5ec2a1896cf47a85c87c02#diff-80be0a15eaa87db8f407cfa1ecfa516691ddee93b5fedb1cebc3347fec7801c8 I've:
- introduced `GetUserWorkspace` also in the proxy handler (instead of the new function which basically was doing the same thing, as you pointed out)
- moved the hierarchical unit tests to the `HandleSpaceGetRequest` subtest 
- splitted the get and list handlers functions from `spacelister` ( this was not strictly required and I can revert it, but it makes the code a bit easier to navigate IMO)

I'm planning also to split the `spacelister_test.go` into 2 separate tests as I've added into the PR description, if this makes sense. 

Please let me know WDYT.

pkg/proxy/handlers/spacelister_get.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	s.ProxyMetrics.RegServWorkspaceHistogramVec.WithLabelValues(fmt.Sprintf("%d", http.StatusOK), metrics.MetricsLabelVerbGet).Observe(time.Since(requestReceivedTime).Seconds())
+	return getWorkspaceResponse(ctx, workspace)
+}
+
+func (s *SpaceLister) GetUserWorkspace(ctx echo.Context, workspaceName string) (*toolchainv1alpha1.Workspace, error) {
+	userSignup, err := s.GetProvisionedUserSignup(ctx)
+	if err != nil {
+		cause := errs.Wrap(err, "error retrieving userSignup")
+		ctx.Logger().Error(cause)
```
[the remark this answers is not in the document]
mfrancisc: Done https://github.com/codeready-toolchain/registration-service/pull/371/commits/4a93f77bcef6c5d82b7a856223e97d561bbc60be

pkg/proxy/proxy.go
```diff
@@ -159,27 +159,44 @@ func (p *Proxy) health(ctx echo.Context) error {
...
 	}
 
 	// before proxying the request, verify that the user has a spacebinding for the workspace and that the namespace (if any) belongs to the workspace
-	workspaces, err := p.spaceLister.ListUserWorkspaces(ctx)
-	if err != nil {
-		return "", nil, crterrors.NewInternalError(errs.New("unable to retrieve user workspaces"), err.Error())
+	var workspaces []toolchainv1alpha1.Workspace
+	if workspaceName != "" {
+		// when a workspace name was provided
+		// recursively get all the spacebindings, starting from this workspace and going up to the parent workspaces till the "root" of the workspace tree.
```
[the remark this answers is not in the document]
mfrancisc: See https://github.com/codeready-toolchain/registration-service/pull/371/commits/1da2b2a15d7eccf67a54fe16fac5bea65ee11102

Thanks!

pkg/proxy/handlers/spacelister_get.go
```diff
@@ -0,0 +1,173 @@
...
+	"github.com/codeready-toolchain/toolchain-common/pkg/spacebinding"
+	"github.com/labstack/echo/v4"
+	errs "github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/selection"
+)
+
+func (s *SpaceLister) HandleSpaceGetRequest(ctx echo.Context) error {
```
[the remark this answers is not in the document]
mfrancisc: I've converted all the methods to functions in those two files, and I'm passing the spaceLister as a parameter.
Please see https://github.com/codeready-toolchain/registration-service/pull/371/commits/192aef489335f6cf21784b753fe8138e4c542576

I can also revert this split if it looks cleaner only to me. The main point was to ease the file navigation, I was a bit lost scrolling up and down searching for things, since the `spacelister.go` file was getting bigger and bigger.

pkg/proxy/proxy.go
```diff
@@ -159,27 +159,44 @@ func (p *Proxy) health(ctx echo.Context) error {
...
-	proxyPluginName, workspace, err := getWorkspaceContext(ctx.Request())
+	proxyPluginName, workspaceName, err := getWorkspaceContext(ctx.Request())
 	if err != nil {
 		return "", nil, crterrors.NewBadRequest("unable to get workspace context", err.Error())
 	}
 
-	ctx.Set(context.WorkspaceKey, workspace) // set workspace context for logging
-	cluster, err := p.app.MemberClusterService().GetClusterAccess(userID, username, workspace, proxyPluginName)
+	ctx.Set(context.WorkspaceKey, workspaceName) // set workspace context for logging
+	cluster, err := p.app.MemberClusterService().GetClusterAccess(userID, username, workspaceName, proxyPluginName)
```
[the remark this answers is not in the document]
mfrancisc: Good point. I've moved the cluster retrieval to the end in https://github.com/codeready-toolchain/registration-service/pull/371/commits/e936a0b2c8b6403172e2a54ae6f36956115aa26d

But I haven't done the simplification of `GetClusterAccess` yet, since that will require some unit test updates, let's do that in the next PR if that is ok for you. 

Thanks for the suggestion.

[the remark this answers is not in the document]
mfrancisc: Never mind I have reverted this change in https://github.com/codeready-toolchain/registration-service/pull/371/commits/138336d1cefde5b7b318b0282b76461c064e25b0 

Since the tests needed to be updated even just for moving the function down. So let's address this separately if that sounds good.

MatousJobanek (approved):
Looks great, thanks a lot for the additional changes :+1: :rocket:

PR #374, opened 2023-11-20
close fake server at the end of the test

MatousJobanek:
Let's close the server at the end of the test to get rid of the annoying error 
```
listen tcp :8081: bind: address already in use
```

### DIFF
--- a/pkg/server/server_test.go
+++ b/pkg/server/server_test.go
@@ -1,6 +1,7 @@
 package server_test
 
 import (
+	"errors"
 	"io"
 	"net/http"
 	"testing"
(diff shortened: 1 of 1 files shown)


mfrancisc (commented):
Thanks for looking into it 🙏 

It seems to be still there the issue :( 
https://github.com/codeready-toolchain/registration-service/actions/runs/6931916142/job/18854704510?pr=374

I was trying to looking into it as well.

pkg/server/server_test.go
```diff
@@ -105,17 +106,20 @@ func (s *TestServerSuite) TestServer() {
...
 	go func() {
-		err := srv.ListenAndServe()
-		require.NoError(t, err)
+		if err := srv.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			require.NoError(t, err)
+		}
 	}()
-	return srv
+	t.Cleanup(func() {
+		require.NoError(t, srv.Close())
```
ranakan19, line 121: should this be a srv.Stop() instead of a srv.Close()?

alexeykazakov:
This should fix it: https://github.com/codeready-toolchain/registration-service/pull/375

PR #375, opened 2023-11-20
Fix broken test

filariow:
related to #374

Seems to me the tests on `pkg/proxy/...` and `pkg/server/...` are requesting the same port for proxy server. That causes a collision as tests are executed in concurrent goroutines.

Signed-off-by: Francesco Ilario <filario@redhat.com>

### DIFF
--- a/pkg/server/server_test.go
+++ b/pkg/server/server_test.go
@@ -109,7 +109,10 @@ func startFakeProxy(t *testing.T) *http.Server {
 	mux := http.NewServeMux()
 	mux.HandleFunc("/proxyhealth", fakehealth)
 
-	srv := &http.Server{Addr: ":8091", Handler: mux, ReadHeaderTimeout: 2 * time.Second}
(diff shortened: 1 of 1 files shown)


pkg/server/server_test.go
```diff
@@ -110,7 +109,7 @@ func startFakeProxy(t *testing.T) *http.Server {
 	mux := http.NewServeMux()
 	mux.HandleFunc("/proxyhealth", fakehealth)
 
-	srv := &http.Server{Addr: ":" + proxy.ProxyPort, Handler: mux, ReadHeaderTimeout: 2 * time.Second}
+	srv := &http.Server{Addr: ":8091", Handler: mux, ReadHeaderTimeout: 2 * time.Second}
```
alexeykazakov, line 112: Let's add a comment here why we need to use a different port so we don't forget about this :)

alexeykazakov (approved):
Great catch!

rajivnathan (approved):
Great catch!

sbryzak (approved):

PR #379, opened 2023-11-30
make required captcha score configurable

mfrancisc:
Use new configuration options:
- `toolchainconfig.Spec.Host.RegistrationService.Verification.Captcha.RequiredScore` which blocks usersginup below this value (if it's not a reactivation)
- `toolchainconfig.Spec.Host.RegistrationService.Verification.Captcha. AllowLowScoreReactivation` is enabled by default and allows reactivations to proceed even when captcha score is below the configured `RequiredScore`.

Related PR:
- toolchain-common: https://github.com/codeready-toolchain/toolchain-common/pull/344
- sandbox-sre: https://github.com/codeready-toolchain/sandbox-sre/pull/1348

Jira: https://issues.redhat.com/browse/SANDBOX-495

### DIFF
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -281,9 +281,9 @@ func (r VerificationConfig) CaptchaScoreThreshold() float32 {
 	return float32(thresholdFloat)
 }
 
-func (r VerificationConfig) AutomaticVerificationThreshold() float32 {
+func (r VerificationConfig) CaptchaRequiredScore() float32 {
 	const defaultAutomaticVerificationThreshold float32 = 0.6
-	threshold := commonconfig.GetString(r.c.Captcha.AutomaticVerificationThreshold, "0.6")
+	threshold := commonconfig.GetString(r.c.Captcha.RequiredScore, "0.6")
 	thresholdFloat, err := strconv.ParseFloat(threshold, 32)
 	if err != nil {
 
@@ -293,6 +293,10 @@ func (r VerificationConfig) AutomaticVerificationThreshold() float32 {
 		return defaultAutomaticVerificationThreshold
 	}
 	return float32(thresholdFloat)
+}
+
+func (r VerificationConfig) CaptchaAllowLowScoreReactivation() bool {
+	return commonconfig.GetBool(r.c.Captcha.AllowLowScoreReactivation, true)
 }
 
 func (r VerificationConfig) CaptchaSiteKey() string {
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -281,6 +281,20 @@ func (r VerificationConfig) CaptchaScoreThreshold() float32 {
 	return float32(thresholdFloat)
 }
 
+func (r VerificationConfig) AutomaticVerificationThreshold() float32 {
+	const defaultAutomaticVerificationThreshold float32 = 0.6
+	threshold := commonconfig.GetString(r.c.Captcha.AutomaticVerificationThreshold, "0.6")
+	thresholdFloat, err := strconv.ParseFloat(threshold, 32)
+	if err != nil {
+
+		if threshold != "" {
+			log.Error(nil, err, fmt.Sprintf("unable to parse automatic verification threshold, using default value '%.1f'", defaultAutomaticVerificationThreshold))
+		}
+		return defaultAutomaticVerificationThreshold
+	}
+	return float32(thresholdFloat)
+}
+
 func (r VerificationConfig) CaptchaSiteKey() string {
 	return commonconfig.GetString(r.c.Captcha.SiteKey, "")
 }
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -282,8 +282,8 @@ func (r VerificationConfig) CaptchaScoreThreshold() float32 {
 }
 
 func (r VerificationConfig) CaptchaRequiredScore() float32 {
-	const defaultAutomaticVerificationThreshold float32 = 0.6
-	threshold := commonconfig.GetString(r.c.Captcha.RequiredScore, "0.6")
+	const defaultAutomaticVerificationThreshold float32 = 0
+	threshold := commonconfig.GetString(r.c.Captcha.RequiredScore, "0")
 	thresholdFloat, err := strconv.ParseFloat(threshold, 32)
 	if err != nil {
 
--- a/pkg/verification/service/verification_service.go
+++ b/pkg/verification/service/verification_service.go
@@ -353,10 +353,12 @@ func (s *ServiceImpl) VerifyPhoneCode(ctx *gin.Context, userID, username, code s
 
 func checkRequiredCaptchaScore(ctx *gin.Context, signup *toolchainv1alpha1.UserSignup, cfg configuration.RegistrationServiceConfig) error {
 	captchaScore, found := signup.Annotations[toolchainv1alpha1.UserSignupCaptchaScoreAnnotationKey]
-	fscore, parseErr := strconv.ParseFloat(captchaScore, 32)
-	if found && parseErr == nil && float32(fscore) < cfg.Verification().CaptchaRequiredScore() {
-		log.Error(ctx, errors.New("captcha score too low"), "automatic verification disabled, manual approval required for user")
-		return crterrors.NewForbiddenError("verification failed", "verification is not available at this time")
+	if found {
+		fscore, parseErr := strconv.ParseFloat(captchaScore, 32)
+		if parseErr == nil && float32(fscore) < cfg.Verification().CaptchaRequiredScore() {
+			log.Info(ctx, "captcha score %v is too low, automatic verification disabled, manual approval required for user", float32(fscore))
+			return crterrors.NewForbiddenError("verification failed", "verification is not available at this time")
+		}
 	}
 	return nil
 }
--- a/pkg/verification/service/verification_service.go
+++ b/pkg/verification/service/verification_service.go
@@ -228,12 +228,27 @@ func (s *ServiceImpl) VerifyPhoneCode(ctx *gin.Context, userID, username, code s
 		return crterrors.NewInternalError(lookupErr, fmt.Sprintf("error retrieving usersignup: %s", userID))
 	}
 
-	// require manual approval if captcha score below automatic verification threshold
-	captchaScore, found := signup.Annotations[toolchainv1alpha1.UserSignupCaptchaScoreAnnotationKey]
-	fscore, parseErr := strconv.ParseFloat(captchaScore, 32)
-	if found && parseErr == nil && float32(fscore) < cfg.Verification().AutomaticVerificationThreshold() {
-		log.Error(ctx, errors.New("captcha score too low"), "automatic verification disabled, manual approval required for user")
-		return crterrors.NewForbiddenError("verification failed", "verification is not available at this time")
+	// check if it's a reactivation
+	if activationCounterString, foundActivationCounter := signup.Annotations[toolchainv1alpha1.UserSignupActivationCounterAnnotationKey]; foundActivationCounter && cfg.Verification().CaptchaAllowLowScoreReactivation() {
+		activationCounter, err := strconv.Atoi(activationCounterString)
+		if err != nil {
+			log.Error(ctx, err, "activation counter is not an integer value, checking required captcha score")
+			// require manual approval if captcha score below automatic verification threshold
+			if err = checkRequiredCaptchaScore(ctx, signup, cfg); err != nil {
+				return err
+			}
+		} else if activationCounter == 1 {
+			// check required captcha score if it's not a reactivation
+			if err = checkRequiredCaptchaScore(ctx, signup, cfg); err != nil {
+				return err
+			}
+		}
+	} else {
+		// when allowLowScoreReactivation is not enabled
+		// require manual approval if captcha score below automatic verification threshold for all users
+		if err := checkRequiredCaptchaScore(ctx, signup, cfg); err != nil {
+			return err
+		}
 	}
 
 	annotationValues := map[string]string{}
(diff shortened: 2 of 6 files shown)


pkg/verification/service/verification_service_test.go
```diff
@@ -896,41 +896,279 @@ func (s *TestVerificationServiceSuite) TestVerifyPhoneCode() {
...
+
+			ctx, _ := gin.CreateTestContext(httptest.NewRecorder())
+			err = s.Application.VerificationService().VerifyPhoneCode(ctx, userSignup.Name, userSignup.Spec.Username, "123456")
+			require.NoError(s.T(), err)
+
+			userSignup, err = s.FakeUserSignupClient.Get(userSignup.Name)
+			require.NoError(s.T(), err)
+
+			require.False(s.T(), states.VerificationRequired(userSignup))
+		})
```
alexeykazakov, line 1171: Alternatively you could organize the tests as an array of cases where you provide:
- the score and activation counter annotations
- configuration
- expected result
And then go over them to test. It will reduce code duplication and will be easy to add more cases.

Could you please also add tests to cover cases with:
- there is no score annotation at all.
- score annotation is not valid

alexeykazakov (commented):
Looks great! Just a few minor comments.

pkg/verification/service/verification_service.go
```diff
@@ -228,12 +228,27 @@ func (s *ServiceImpl) VerifyPhoneCode(ctx *gin.Context, userID, username, code s
...
+				return err
+			}
+		} else if activationCounter == 1 {
+			// check required captcha score if it's not a reactivation
+			if err = checkRequiredCaptchaScore(ctx, signup, cfg); err != nil {
+				return err
+			}
+		}
+	} else {
+		// when allowLowScoreReactivation is not enabled
```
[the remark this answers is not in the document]
mfrancisc: thanks!

pkg/verification/service/verification_service.go
```diff
@@ -336,6 +351,16 @@ func (s *ServiceImpl) VerifyPhoneCode(ctx *gin.Context, userID, username, code s
...
 }
 
+func checkRequiredCaptchaScore(ctx *gin.Context, signup *toolchainv1alpha1.UserSignup, cfg configuration.RegistrationServiceConfig) error {
+	captchaScore, found := signup.Annotations[toolchainv1alpha1.UserSignupCaptchaScoreAnnotationKey]
+	fscore, parseErr := strconv.ParseFloat(captchaScore, 32)
+	if found && parseErr == nil && float32(fscore) < cfg.Verification().CaptchaRequiredScore() {
+		log.Error(ctx, errors.New("captcha score too low"), "automatic verification disabled, manual approval required for user")
+		return crterrors.NewForbiddenError("verification failed", "verification is not available at this time")
+	}
+	return nil
```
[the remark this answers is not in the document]
mfrancisc: yeah, better to parse the score only when found. Good catch!

pkg/verification/service/verification_service_test.go
mfrancisc: > Could you please also add tests to cover cases 

Thanks! Done in the same commit. But I'm not 100% sure what should happen in those cases, seems like currently the verification process just goes on, ignoring that the captcha score is missing/invalid. So I just did that. 

Not sure if we should log/return an error or if this is already being done in other parts of the verification process, I guess before going to the phone verification process.

alexeykazakov: If the score is missing that we should treat it as a high score. For invalid scores ideally we would want to return an error but it's OK if we just log the error and ignore it for now.

pkg/configuration/configuration.go
```diff
@@ -281,6 +281,24 @@ func (r VerificationConfig) CaptchaScoreThreshold() float32 {
 	return float32(thresholdFloat)
 }
 
+func (r VerificationConfig) CaptchaRequiredScore() float32 {
+	const defaultAutomaticVerificationThreshold float32 = 0.6
+	threshold := commonconfig.GetString(r.c.Captcha.RequiredScore, "0.6")
```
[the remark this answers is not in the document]
mfrancisc: Right now it is already using the hard-coded `0.6` value : https://github.com/codeready-toolchain/registration-service/blob/ec03997ea2af344078922dec3dddd65a2a0435bc/pkg/verification/service/verification_service.go#L234
We can set it to 0, but this means we have to open and merge a PR in sandbox-sre with the `0.6` threshold before merging this. Not sure if it makes any difference 🤷‍♂️ . Any preference ?

pkg/verification/service/verification_service.go
```diff
@@ -228,12 +228,27 @@ func (s *ServiceImpl) VerifyPhoneCode(ctx *gin.Context, userID, username, code s
...
+			if err = checkRequiredCaptchaScore(ctx, signup, cfg); err != nil {
+				return err
+			}
+		}
+	} else {
+		// when allowLowScoreReactivation is not enabled or no activation counter found
+		// require manual approval if captcha score below automatic verification threshold for all users
+		if err := checkRequiredCaptchaScore(ctx, signup, cfg); err != nil {
+			return err
+		}
```
rajivnathan, line 251: I may be missing something but it looks like all cases lead to calling `checkRequiredCaptchaScore(ctx, signup, cfg)`. If CaptchaAllowLowScoreReactivation is true and it's a reactivation then doesn't that mean we don't need to do additional (phone) verification and we can just proceed to activate the user?

So maybe we can simplify the logic by removing the CaptchaAllowLowScoreReactivation check from here and move it to the IsPhoneVerificationRequired function in https://github.com/codeready-toolchain/registration-service/blob/1602277a54b638f1056f52ef77757f41cc61d2d5/pkg/signup/service/signup_service.go#L227C2-L232 so that we skip phone verification at that point instead and keep the logic simple here?

rajivnathan (commented):
I was getting a bit confused by the logic around `checkRequiredCaptchaScore`. There are 3 segments for scores right?

1. `UserScore` > `ScoreThreshold` (no additional verification required)
2. `ScoreThreshold` > `UserScore` > `RequiredScore` (additional verification required)
3. `RequiredScore` > `UserScore` (manual verification required)

Just an idea but maybe it would be easier to understand the logic if `checkRequiredCaptchaScore` is renamed to `checkRequiredManualApproval`?

When zooming out of the scope of this specific PR I think it's easy to confuse `ScoreThreshold` and `RequiredScore`. Something like `AdditionalVerificationThreshold` maybe would have been a bit more clear.

pkg/verification/service/verification_service.go
mfrancisc: > I may be missing something but it looks like all cases lead to calling checkRequiredCaptchaScore(ctx, signup, cfg).

the case in which is `activationCounter > 1` and `CaptchaAllowLowScoreReactivation == true` doesn't lead to `checkRequiredCaptchaScore` . Maybe I can shuffle the if statements a bit to make this more clear 🤔

mfrancisc:
> Just an idea but maybe it would be easier to understand the logic if checkRequiredCaptchaScore is renamed to checkRequiredManualApproval?

I like the name more , thanks for the feedback. Renamed in https://github.com/codeready-toolchain/registration-service/pull/379/commits/d5a6f9ba04cc8e77865bdc3387151ff6f83e7efe

> When zooming out of the scope of this specific PR I think it's easy to confuse ScoreThreshold and RequiredScore. Something like AdditionalVerificationThreshold maybe would have been a bit more clear. And something like CaptchaVerificationThreshold probably would have been better than ScoreThreshold.

This makes sense indeed. As all those captcha configuration threshold can be a bit confusing now. But let's do this in separate PRs , WDYT ?

pkg/verification/service/verification_service.go
rajivnathan: >Not sure, we could move it there but we need to pass the activation counter from the user. Do you think that would make it simpler ?

You're right, that part of the code doesn't reference the UserSignup so I guess it's more complicated. My bad, disregard the suggestion.

pkg/configuration/configuration.go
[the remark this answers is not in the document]
rajivnathan: I think 0 as a default makes more sense too, that way it's permissible by default and you only need to set it if you want to limit automatic approval.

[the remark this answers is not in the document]
mfrancisc: makes sense. I've changed the default to `0` in https://github.com/codeready-toolchain/registration-service/pull/379/commits/a7c22771f1d488dbc1525a6e7018b26ac05261ab. Thanks!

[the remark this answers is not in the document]
mfrancisc: and related sandbox-sre PR: https://github.com/codeready-toolchain/sandbox-sre/pull/1348

pkg/verification/service/verification_service_test.go
```diff
@@ -896,41 +896,122 @@ func (s *TestVerificationServiceSuite) TestVerifyPhoneCode() {
...
-					toolchainv1alpha1.UserVerificationExpiryAnnotationKey:     now.Add(10 * time.Second).Format(verificationservice.TimestampLayout),
-				},
-				Labels: map[string]string{
-					toolchainv1alpha1.UserSignupUserPhoneHashLabelKey: "+1NUMBER",
-				},
+	s.T().Run("captcha configuration ", func(t *testing.T) {
+		tests := map[string]struct {
+			activationCounterAnnotationValue string
+			captchaScoreAnnotationValue      string
+			allowLowScoreConfiguration       bool
```
alexeykazakov, line 903: Shouldn't it be:
```suggestion
			allowLowScoreReactivationConfiguration       bool
```
?

pkg/verification/service/verification_service_test.go
```diff
@@ -896,41 +896,122 @@ func (s *TestVerificationServiceSuite) TestVerifyPhoneCode() {
...
+				captchaScoreAnnotationValue:      "0.5", // and captcha score is low
+				allowLowScoreConfiguration:       true,
+				expectedErr:                      "verification failed: verification is not available at this time",
+			},
+			"activation counter is invalid and captcha score is ok": {
+				activationCounterAnnotationValue: "x",   // something wrong happened
+				captchaScoreAnnotationValue:      "0.6", // but captcha score is ok
+				allowLowScoreConfiguration:       true,
+			},
+			"allow low score disabled - captcha score below required score and it's a reactivation": {
```
alexeykazakov, line 928:
```suggestion
			"allow low score reactivation disabled - captcha score below required score and it's a reactivation": {
```

pkg/verification/service/verification_service_test.go
```diff
@@ -896,41 +896,122 @@ func (s *TestVerificationServiceSuite) TestVerifyPhoneCode() {
...
+				activationCounterAnnotationValue: "2",   // user is reactivating
+				captchaScoreAnnotationValue:      "0.5", //  captcha score is low
+				allowLowScoreConfiguration:       false,
+				expectedErr:                      "verification failed: verification is not available at this time",
+			},
+			"allow low score disabled - captcha score below required score and it's not a reactivation": {
```
alexeykazakov, line 934:
```suggestion
			"allow low score reactivation disabled - captcha score below required score and it's not a reactivation": {
```

pkg/verification/service/verification_service_test.go
```diff
@@ -896,41 +896,122 @@ func (s *TestVerificationServiceSuite) TestVerifyPhoneCode() {
...
+				activationCounterAnnotationValue: "1",   // first time user
+				captchaScoreAnnotationValue:      "0.5", //  captcha score is low
+				allowLowScoreConfiguration:       false,
+				expectedErr:                      "verification failed: verification is not available at this time",
+			},
+			"allow low score disabled - captcha score ok": {
```
alexeykazakov, line 940:
```suggestion
			"allow low score reactivation disabled - captcha score ok": {
```

pkg/verification/service/verification_service_test.go
```diff
@@ -896,41 +896,122 @@ func (s *TestVerificationServiceSuite) TestVerifyPhoneCode() {
...
+				activationCounterAnnotationValue: "1",   // first time user
+				captchaScoreAnnotationValue:      "0.6", //  captcha score is ok
+				allowLowScoreConfiguration:       false,
+			},
+			"no score annotation": {
```
alexeykazakov, line 945: Let's also add a test for the case with no activation counter annotation set at all.

alexeykazakov (approved):
Thank you for addressing the comments! The test look much better now! I left a few minor comments. But overall looks good!

pkg/verification/service/verification_service_test.go
mfrancisc: good point! done , please see https://github.com/codeready-toolchain/registration-service/pull/379/commits/c09beb3d59e404b08a99285b2d7e6437a5b4b3df

pkg/verification/service/verification_service.go
```diff
@@ -336,6 +351,23 @@ func (s *ServiceImpl) VerifyPhoneCode(ctx *gin.Context, userID, username, code s
 	return
 }
 
+func checkRequiredCaptchaScore(ctx *gin.Context, signup *toolchainv1alpha1.UserSignup, cfg configuration.RegistrationServiceConfig) error {
```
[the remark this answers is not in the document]
MatousJobanek: never mind - I made the comment before I saw that you didn't use `0` is the default value, so I understand why it was confusing.
for the background of my question - I was just wondering if we need to do all these checks when the value is `0`, couldn't we just skip it? However, I understand that logical-wise it doesn't change anything.

MatousJobanek (approved):
:+1:

pkg/verification/service/verification_service.go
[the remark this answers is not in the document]
mfrancisc: ok, got it! thanks for explaining that and for the review 👍

rajivnathan (approved):
Nice 👍

PR #382, opened 2023-12-22
Support oc web login in proxy

alexeykazakov:
This PR enables web login support which is available in oc 4.14+:
```
oc login -w <proxy_url>
```
This will open a browser window where the user will be asked to log into the SSO.

Requires:
- https://github.com/codeready-toolchain/api/pull/392
- https://github.com/codeready-toolchain/toolchain-common/pull/348

I don't think we can properly test it in our e2e tests so this PR comes with unit tests only.


pkg/proxy/proxy_test.go
```diff
@@ -123,524 +123,611 @@ func (s *TestProxySuite) TestProxy() {
...
-					// then
-					require.NoError(s.T(), err)
-					require.NotNil(s.T(), resp)
-					defer resp.Body.Close()
-					assert.Equal(s.T(), http.StatusInternalServerError, resp.StatusCode)
-				})
-			})
+			s.checkPlainHTTPErrors(fakeApp)
+			s.checkWebsocketsError()
+			s.checkWebLogin()
```
alexeykazakov, line 128: Please review the `s.checkWebLogin()` function with the new tests. The rest of the change here is just to make the linter happy. I moved other tests to separate functions as is without changing anything there.

pkg/proxy/proxy.go
```diff
@@ -149,6 +168,78 @@ func (p *Proxy) StartProxy() *http.Server {
...
+	}
+	targetURL.Path = ctx.Request().URL.Path
+	targetURL.RawQuery = ctx.Request().URL.RawQuery
+
+	return p.redirectTo(ctx, targetURL.String())
+}
+
+func (p *Proxy) redirectTo(ctx echo.Context, to string) error {
+	log.InfoEchof(ctx, "redirecting %s to %s", ctx.Request().URL.String(), to)
+	http.Redirect(ctx.Response().Writer, ctx.Request(), to, http.StatusSeeOther)
```
rajivnathan, line 213: Just curious, is there a reason for using `http.StatusSeeOther` specifically here as oppose to `http.StatusFound` for example?

rajivnathan (approved):
Cool stuff 👍

pkg/proxy/proxy.go
alexeykazakov: 302 would probably work just fine but I think 303 is more suitable here. But I don't have any strong opinion though. Those redirect options is often a bit confusing due to some overlapping :)

pkg/proxy/proxy.go
```diff
@@ -149,6 +168,78 @@ func (p *Proxy) StartProxy() *http.Server {
...
+}
+
+// handleSSORequest handles requests to the cluster authentication server and proxy them to SSO instead. Used by web login.
+func (p *Proxy) handleSSORequest(targetURL *url.URL) echo.HandlerFunc {
+	return func(ctx echo.Context) error {
+		req := ctx.Request()
+		director := func(req *http.Request) {
+			origin := req.URL.String()
+			req.URL.Scheme = targetURL.Scheme
+			req.URL.Host = targetURL.Host
```
ranakan19, line 224: curious if there is a difference between req.URL.Host and req.Host? both have the same value being assigned i.e targetURL.Host

alexeykazakov: Surprisingly there seems to be some difference. Setting req.URL.Host is enough for some services (Kube API Server, our own Dev Keycloak). But Akamai which is used by sso.redhat.com also requires req.Host to be set to the same value as the req.URL.Host.

ranakan19 (approved):

pkg/proxy/proxy.go
```diff
@@ -149,6 +168,78 @@ func (p *Proxy) StartProxy() *http.Server {
...
+// openidAuth handles requests to the openID Connect authentication endpoint. Used by web login.
+func (p *Proxy) openidAuth(ctx echo.Context) error {
+	targetURL, err := url.Parse(authorizationEndpointTarget())
+	if err != nil {
+		return err
+	}
+	targetURL.Path = ctx.Request().URL.Path
+	targetURL.RawQuery = ctx.Request().URL.RawQuery
+
+	return p.redirectTo(ctx, targetURL.String())
```
MatousJobanek, line 208: We need to use redirect instead of `handleSSORequest` because this is technically the web login in the browser, is that correct?

pkg/proxy/proxy.go
```diff
@@ -149,6 +168,78 @@ func (p *Proxy) StartProxy() *http.Server {
 	return srv
 }
 
+// unsecured returns true if the request does not require authentication
+func unsecured(ctx echo.Context) bool {
+	uri := ctx.Request().URL.RequestURI()
+	return uri == proxyHealthEndpoint || uri == "/.well-known/oauth-authorization-server" || strings.HasPrefix(uri, authEndpoint)
```
MatousJobanek, line 174: it's just a detail, but could you please (for easy navigation) store the `/.well-known/oauth-authorization-server` path as a constant and use it also in 
```
router.Any("/.well-known/oauth-authorization-server", p.oauthConfiguration)
```

MatousJobanek (approved):
Awesome job :+1: 
Out of curiosity, how did you figure out the expected endpoints/flow? Is this something that you already knew before, something that you saw in OpenShift/`oc` code, or something that you reverse-engineered based on running the `oc login -w` command?
I'm wondering if this is something that is bound to the `oc` implementation/version or something that shouldn't change much in the future. 

I understand that running a proper e2e test would require deploying a keycloak as part of the suite, however, the unit-tests are very opinionated and are following the mocked happy path (I have no idea how to make it better though :wink:), which means that we will need to do the manual verification when there is a new version of OpenShift/`oc` we would like to migrate to.

pkg/proxy/proxy.go
```diff
@@ -121,9 +134,15 @@ func (p *Proxy) StartProxy() *http.Server {
...
 	// routes
 	wg := router.Group("/apis/toolchain.dev.openshift.com/v1alpha1/workspaces")
+	// Space lister routes
 	wg.GET("/:workspace", handlers.HandleSpaceGetRequest(p.spaceLister))
 	wg.GET("", handlers.HandleSpaceListRequest(p.spaceLister))
 	router.GET(proxyHealthEndpoint, p.health)
+	// SSO routes. Used by web login.
+	router.Any("/.well-known/oauth-authorization-server", p.oauthConfiguration)
+	router.Any(fmt.Sprintf("%s*", openidAuthEndpoint()), p.openidAuth)
+	router.Any(fmt.Sprintf("%s*", authEndpoint), p.auth)
```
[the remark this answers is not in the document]
alexeykazakov: Good point. Let me try to add some comment for that.

alexeykazakov:
> Is this something that you already knew before, something that you saw in OpenShift/oc code, or something that you reverse-engineered based on running the oc login -w command?

It's a combination of everything but mostly reverse-engineering :)

> I'm wondering if this is something that is bound to the oc implementation/version or something that shouldn't change much in the future.

It requires oc 1.14 or higher. But yes I do have a concern that it can break in the future versions because we are basically rely on the oc and ocp implementations details here.

> I understand that running a proper e2e test would require deploying a keycloak as part of the suite, however, the unit-tests are very opinionated and are following the mocked happy path (I have no idea how to make it better though 😉), which means that we will need to do the manual verification when there is a new version of OpenShift/oc we would like to migrate to.

Yes. The unit-tests test the implementation of the proxy only. If the implementation of oc or ocp changes it may break the web login and our test won't detect it. I was thinking about how we can test it but it would require more than just a Keycloak (we could rely on our dev keyloak instead of provisioning a built-in one with the test as an option) but it would also require a browser and an oc binary. It's not just an API call to our backend anymore. And adding it all to our e2e tests would make it even harder because our e2e tests runs in the "e2e-tests" mode/environment which uses our built-in mock "sso" to validate tokens against. So our proxy would reject sso tokens received from the "real" sso.

So we are basically talking about a new type of the tests (in addition to unit tests and our e2e-tests). Which we will have to introduce at some point, most likely, if we start relying on the web login through the proxy as a critical feature.

pkg/proxy/proxy.go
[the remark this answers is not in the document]
alexeykazakov: I hope this helps: https://github.com/codeready-toolchain/registration-service/pull/382/commits/5a8e586828d7e638e5cc85c94d28e8dc1e25302c

pkg/proxy/proxy.go
alexeykazakov: See https://github.com/codeready-toolchain/registration-service/pull/382/commits/5a8e586828d7e638e5cc85c94d28e8dc1e25302c

pkg/proxy/proxy.go
alexeykazakov: Done in https://github.com/codeready-toolchain/registration-service/pull/382/commits/5a8e586828d7e638e5cc85c94d28e8dc1e25302c

pkg/proxy/proxy.go
```diff
@@ -121,9 +135,25 @@ func (p *Proxy) StartProxy() *http.Server {
...
 	router.GET(proxyHealthEndpoint, p.health)
+	// SSO routes. Used by web login (oc login -w).
+	// Here is the expected flow for the "oc login -w" command:
+	// 1. "oc login -w <proxy_url>"
+	// 2. oc calls <proxy_url>/.well-known/oauth-authorization-server (wellKnownOauthConfigEndpoint endpoint)
+	// 3. proxy forwards it to <sso_url>/auth/realms/<sso_realm>/.well-known/openid-configuration
+	// 4. oc starts an OAuth flow by opening a browser for <proxy_url>/auth/realms/<realm>/protocol/openid-connect/auth
+	// 5. proxy redirects (the request is not proxied but redirected via 403 See Others response!) the request
+	//    to <sso_url>/auth/realms/<realm>/protocol/openid-connect/auth
+	//    Note: oc uses this hardcoded public (no secret) oauth client name: "openshift-cli-client" which has to exist in SSO to make this flow work.
```
MatousJobanek, line 150: I wanted to just aks if we had to do something special for our SSO client, but it looks like that oc has the client name hard-coded, which means that we need to stick with that.

I guess that this means that we needed to configure the dev keycloak to be aware of this client with some special configuration I guess

pkg/proxy/proxy.go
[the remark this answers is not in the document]
MatousJobanek: awesome, thanks a lot :+1:

pkg/proxy/proxy.go
MatousJobanek: gotcha, thanks for the explanation.

MatousJobanek (approved):
Thanks for providing the additional comments :pray: :1st_place_medal:

PR #383, opened 2024-01-03
Log recaptcha assessment response

rajivnathan:


ranakan19 (approved):

alexeykazakov (approved):

MatousJobanek (approved):
cool :+1: 
but I'm wondering if it's something we could store as an annotation as well. You know, for easy analysis of the suspicious/banned account directly from the UserSignup resource instead of searching through the log storage which has also some limited retention.

rajivnathan:
>but I'm wondering if it's something we could store as an annotation as well. You know, for easy analysis of the suspicious/banned account directly from the UserSignup resource instead of searching through the log storage which has also some limited retention.
Good idea. However, I'm not sure how helpful this information would be. Maybe we can start with these logs and add to annotation on UserSignup later if we feel like there's some specific information we want to include there?

MatousJobanek:
Yeah sure... btw, what the value actually looks like? What does it contain?

PR #385, opened 2024-01-08
update email hash when it doesn't match

MatousJobanek:
https://redhat-internal.slack.com/archives/CHK0J6HT6/p1704721307154849?thread_ts=1704718968.866119&cid=CHK0J6HT6

### DIFF
--- a/pkg/signup/service/signup_service.go
+++ b/pkg/signup/service/signup_service.go
@@ -521,6 +521,21 @@ func (s *ServiceImpl) auditUserSignupAgainstClaims(ctx *gin.Context, userSignup 
 
 	userSignup.Spec.IdentityClaims = c
 
+	// make sure that labels and annotations are initiated
+	if userSignup.Labels == nil {
+		userSignup.Labels = map[string]string{}
+	}
+	if userSignup.Annotations == nil {
+		userSignup.Annotations = map[string]string{}
+	}
+
+	// make sure that he email hash matches the email
+	emailHash := hash.EncodeString(c.Email)
+	if userSignup.Labels[toolchainv1alpha1.UserSignupUserEmailHashLabelKey] != emailHash {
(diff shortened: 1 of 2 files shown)


alexeykazakov (approved):

rajivnathan (approved):

pkg/signup/service/signup_service.go
```diff
@@ -521,6 +521,21 @@ func (s *ServiceImpl) auditUserSignupAgainstClaims(ctx *gin.Context, userSignup
...
+	if userSignup.Labels == nil {
+		userSignup.Labels = map[string]string{}
+	}
+	if userSignup.Annotations == nil {
+		userSignup.Annotations = map[string]string{}
+	}
+
+	// make sure that he email hash matches the email
+	emailHash := hash.EncodeString(c.Email)
+	if userSignup.Labels[toolchainv1alpha1.UserSignupUserEmailHashLabelKey] != emailHash {
```
xcoulon, line 534: out of curiosity, in which cases may we encounter such a mismatch?

MatousJobanek: 1. user signs up, which means that we create/update UserSignup resource
2. the UserSignup CR is populated using the values from the token (let's say email `john@acme.com`)
3. user modifies his/her email in SSO (let's say to `john@gmail.com`)
4. comes to DevSandbox landing page (as already provisioned user) 
5. this `auditUserSignupAgainstClaims` method is executed on `GET` call to keep the claims up-to-date
6. so the email gets updated with the new one from the SSO claim (to `john@gmail.com`), but the hash label stays the same from the time when the usersignup was created (hash of `john@acme.com`)

xcoulon: I see, thanks for this detailed explanation, @MatousJobanek!

xcoulon (approved):

PR #387, opened 2024-01-09
fix: add spacebindingrequests to workspace binding list

mfrancisc:
Users are unable to manage SpaceBindingRequests that are not provisioned ( meaning there is no SpaceBinding associated with it).

This PR adds the SpaceBindingRequests ( that have no SpaceBinding associated) to the list of Bindings in the Workspace, so that RHTAP users can manage them from the UI.

Jira: https://issues.redhat.com/browse/ASC-443

Related PR:
- https://github.com/codeready-toolchain/toolchain-common/pull/351
- https://github.com/codeready-toolchain/toolchain-e2e/pull/877

TODO:

- [ ] Update E2E tests


pkg/proxy/proxy.go
```diff
@@ -121,7 +124,7 @@ func (p *Proxy) StartProxy() *http.Server {
 
 	// routes
 	wg := router.Group("/apis/toolchain.dev.openshift.com/v1alpha1/workspaces")
-	wg.GET("/:workspace", handlers.HandleSpaceGetRequest(p.spaceLister))
+	wg.GET("/:workspace", handlers.HandleSpaceGetRequest(p.spaceLister, commoncluster.GetMemberClusters))
```
filariow, line 127: should we use `p.getMembersFunc` here?

pkg/proxy/proxy.go
```diff
@@ -47,22 +48,23 @@ const (
...
+	app            application.Application
+	cl             client.Client
+	tokenParser    *auth.TokenParser
+	spaceLister    *handlers.SpaceLister
+	metrics        *metrics.ProxyMetrics
+	getMembersFunc commoncluster.GetMemberClustersFunc
 }
 
-func NewProxy(app application.Application, proxyMetrics *metrics.ProxyMetrics) (*Proxy, error) {
+func NewProxy(app application.Application, proxyMetrics *metrics.ProxyMetrics, getMembersFunc commoncluster.GetMemberClustersFunc) (*Proxy, error) {
```
filariow, line 59: is there a need for passing `getMembersFunc` or we could just default here to `cluster.GetMemberClusters`? As tests are in the same package, there we can just overwrite `p.getMembersFunc`

pkg/proxy/handlers/spacelister_get.go
```diff
@@ -98,6 +108,28 @@ func GetUserWorkspace(ctx echo.Context, spaceLister *SpaceLister, workspaceName
...
+				sbrs := &toolchainv1alpha1.SpaceBindingRequestList{}
+				if err := member.Client.List(context.Background(), sbrs, runtimeclient.InNamespace(namespace.Name)); err != nil {
+					return nil, err
+				}
+				allSbrs = append(allSbrs, sbrs.Items...)
+			}
+		}
+	}
+	return allSbrs, nil
+}
```
filariow, line 131: we should forward the context 

```suggestion
// listSpaceBindingRequestsForSpace searches for the SpaceBindingRequests in all the provisioned namespaces for the given Space
func listSpaceBindingRequestsForSpace(ctx echo.Context, GetMembersFunc cluster.GetMemberClustersFunc, space *toolchainv1alpha1.Space) ([]toolchainv1alpha1.SpaceBindingRequest, error) {
	members := GetMembersFunc()
	if len(members) == 0 {
		return nil, errs.New("no member clusters found")
	}
	var allSbrs []toolchainv1alpha1.SpaceBindingRequest
	for _, member := range members {
		if member.Name == space.Status.TargetCluster {
			// list sbrs in all namespaces for the current space
			for _, namespace := range space.Status.ProvisionedNamespaces {
				sbrs := &toolchainv1alpha1.SpaceBindingRequestList{}
				if err := member.Client.List(ctx, sbrs, runtimeclient.InNamespace(namespace.Name)); err != nil {
					return nil, err
				}
				allSbrs = append(allSbrs, sbrs.Items...)
			}
		}
	}
	return allSbrs, nil
}
```

filariow (commented):
great job! just a few nits

pkg/proxy/handlers/spacelister_get.go
```diff
@@ -157,6 +189,26 @@ func generateWorkspaceBindings(space *toolchainv1alpha1.Space, spaceBindings []t
...
+		if alreadyInBindingList(bindings, &sbr) {
+			continue
+		}
+
+		binding := toolchainv1alpha1.Binding{
+			MasterUserRecord: sbr.Spec.MasterUserRecord,
+			Role:             sbr.Spec.SpaceRole,
+			AvailableActions: []string{},
+		}
+		binding.AvailableActions = []string{UpdateBindingAction, DeleteBindingAction}
```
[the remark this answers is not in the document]
mfrancisc: good point 🤦  . thx!

pkg/proxy/proxy.go
mfrancisc: Good point, but I believe that wuold make it more difficult to test in `HandleSpaceGetRequest`, since we are directly invoking `HandleSpaceGetRequest` without creating a `Proxy` instance. 

WDYT?

pkg/proxy/proxy.go
mfrancisc: TBH I would prefer to have tests in `proxy_test` package, so that functions are tested from a client perspective. 

Also I found it easier to follow if we use a composition pattern instead of overwriting properties in tests function, which makes it a bit more "hidden" IMO. But I don't have a strong opinion about this.

pkg/proxy/proxy.go
filariow: yeah, I agree with you on having to pass the function to `HandleSpaceGetRequest`.

Btw, on this line, should we use `p.getMembersFunc` instead of `commoncluster.GetMemberClusters`? To my understanding, this way we are not forwarding the `GetMemberClustersFunc` provided in `NewProxy` to the handler, but defaulting to the one in `common/cluster` lib.

Basically, is the following change correct?
```suggestion
	wg.GET("/:workspace", handlers.HandleSpaceGetRequest(p.spaceLister, p.getMemberClusters))
```

mfrancisc: Ohh now I understand you you mean, yes we should pass that function indeed. Don in https://github.com/codeready-toolchain/registration-service/pull/387/commits/9095d1ccd4d3fefce7a242b3249b557e486ab354 

Thanks and sorry for the confusion.

filariow (approved):
lgtm

pkg/proxy/handlers/spacelister_get.go
```diff
@@ -119,7 +150,7 @@ func filterUserSpaceBinding(username string, allSpaceBindings []toolchainv1alpha
 
 // generateWorkspaceBindings generates the bindings list starting from the spacebindings found on a given space resource and an all parent spaces.
 // The Bindings list  has the available actions for each entry in the list.
-func generateWorkspaceBindings(space *toolchainv1alpha1.Space, spaceBindings []toolchainv1alpha1.SpaceBinding) ([]toolchainv1alpha1.Binding, error) {
+func generateWorkspaceBindings(space *toolchainv1alpha1.Space, spaceBindings []toolchainv1alpha1.SpaceBinding, spacebindingRequests []toolchainv1alpha1.SpaceBindingRequest) ([]toolchainv1alpha1.Binding, error) {
 	var bindings []toolchainv1alpha1.Binding
```
sadlerap, line 154: something I noticed here, though it's not an issue with your patch per se.  `bindings` should be at most of size `len(spaceBindings) + len(spacebindingRequests)`, so maybe we should do this allocation once?

Don't need to do this change in this PR, but it's something to consider.
```suggestion
	bindings := make([]toolchainv1alpha1.Binding, 0, len(spaceBindings) + len(spacebindingRequests))
```

sadlerap (approved):
lgtm, but one optional thing that might be beneficial.

pkg/proxy/handlers/spacelister_get.go
```diff
@@ -77,9 +79,16 @@ func GetUserWorkspace(ctx echo.Context, spaceLister *SpaceLister, workspaceName
 		ctx.Logger().Error(fmt.Sprintf("unauthorized access - there is no SpaceBinding present for the user %s and the workspace %s", userSignup.CompliantUsername, workspaceName))
 		return nil, nil
 	}
+
+	// list all SpaceBindingRequests , just in case there might be some failing to create a SpaceBinding resource.
+	allSpaceBindingRequests, err := listSpaceBindingRequestsForSpace(ctx, GetMembersFunc, space)
```
alexeykazakov, line 84: I'm a bit concerned about adding even more stuff to the `GetUserWorkspace()` function. This function is currently used not only in our WS lister API but it's called for every request to the proxy in case there is a workspace context in the request. And we don't care about failed SBRs in that case (or even about the available user roles either btw).
Maybe it's time to split this function to two? One to remain a lightweight version for the general proxy requests and and with all additional information for the WS lister. Or make this function flexible/configurable so it can be used everywhere.

pkg/proxy/handlers/spacelister_get.go
mfrancisc: makes sense indeed. Thanks!

pkg/proxy/handlers/spacelister_get.go
mfrancisc: Makes sense. Let me take a look an split it so that proxy request do not have to populate the bindings list. Thanks for the suggestion.

mfrancisc: @alexeykazakov @MatousJobanek  I've created 2 functions  `GetUserWorkspace` used by proxy only and `GetUserWorkspaceWithBindings` used by the spacelister.

Please take a look when you have some time. 

PS: there is one new error condition which is not tested https://github.com/codeready-toolchain/registration-service/pull/387/commits/574d4ebfe6d4c145286ddd6cdf7110d038e128d3#diff-23bf3528567b934bf6ab8d0d690dc61a0ac564ba1b43b89cb5ddde609bb150b1R75-R79 , I can add a specific test in the `proxy_test.go` as I'm not sure where that should go in the current `TestProxy` since it's more a combination of different headers and requests.

pkg/proxy/handlers/spacelister_get.go
```diff
@@ -98,6 +142,43 @@ func GetUserWorkspace(ctx echo.Context, spaceLister *SpaceLister, workspaceName
...
+func getUserSignupAndSpace(ctx echo.Context, spaceLister *SpaceLister, workspaceName string) (*signup.Signup, *toolchainv1alpha1.Space, error) {
+	userSignup, err := spaceLister.GetProvisionedUserSignup(ctx)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	space, err := spaceLister.GetInformerServiceFunc().GetSpace(workspaceName)
+	if err != nil {
+		ctx.Logger().Error(errs.Wrap(err, "unable to get space"))
+		return userSignup, nil, nil
```
alexeykazakov, line 155: Why returning `nil` instead of `err`?

pkg/proxy/handlers/spacelister_get.go
alexeykazakov: It's not that important where you add this test. Add it to whatever place you think makes sense. But let's add a test for it if it's not too tricky.

pkg/proxy/handlers/spacelister_get.go
mfrancisc: I've left the functionality unchanged https://github.com/codeready-toolchain/registration-service/blob/81de976a7e62f89d73a44af3ffc61ce8188ab86e/pkg/proxy/handlers/spacelister_get.go#L52-L56

I see we return 404 when when usersignup and space are not found.

pkg/proxy/handlers/spacelister_get.go
mfrancisc: I've added a specific test for this error condition in https://github.com/codeready-toolchain/registration-service/pull/387/commits/98d2640e31aeb72c2b54076d7f7fcb6b9bd10c99

pkg/proxy/handlers/spacelister_get.go
alexeykazakov: Then let's add a comment to the function description stating that if no space found then a "nil" returned instead of an error.

alexeykazakov (commented):
Looks good overall but it looks like we could add more unit tests.
Could you please take a look at the codecov report and cover the missing cases when it's not too tricky?

filariow (approved):
lgtm

mfrancisc:
> Could you please take a look at the codecov report and cover the missing cases when it's not too tricky?

@alexeykazakov thanks for pointing that out, I've added some more unit tests to cover the error cases, please take a look at https://github.com/codeready-toolchain/registration-service/pull/387/commits/1c1fa6f44bf608c4d67ec793ce38d4af45405b39

alexeykazakov (approved):
Looks good. Thanks.

PR #391, opened 2024-01-12
Update JWT version

alexeykazakov:
Fixes https://issues.redhat.com/browse/ASC-448

Requires https://github.com/codeready-toolchain/toolchain-common/pull/354
Paired with https://github.com/codeready-toolchain/toolchain-e2e/pull/881


pkg/auth/tokenparser_test.go
```diff
@@ -321,4 +321,39 @@ func (s *TestTokenParserSuite) TestTokenParser() {
...
+			ID:       uuid.Must(uuid.NewV4()),
+			Username: username0,
+		}
+		email0 := identity0.Username + "@email.tld"
+
+		tests := map[string]struct {
+			aud []string
+		}{
+			"single string": {
+				aud: []string{"aud-claim-1"},
```
mfrancisc, line 337: Not sure "if" and "how", but should we check instead that it supports single string "aud-claim-1" and not a slice containing a single element. I think this is what the rfc says: https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3 and this is what caused the issue in our case.

pkg/auth/tokenparser_test.go
```diff
@@ -321,4 +321,39 @@ func (s *TestTokenParserSuite) TestTokenParser() {
 		require.Equal(s.T(), email0, claims.Email)
 		require.Equal(s.T(), "OriginalSubValue:1234-ABCD", claims.OriginalSub)
 	})
+
+	s.Run("parse valid token with aud claim", func() {
```
mfrancisc, line 325: TBH I'm not 100% sure we need to test this here, as this seems more functionality that belongs to the JWT library. And it should be already tested here https://github.com/golang-jwt/jwt/blob/3c0777d0c9b7acb6c7d1ad1ae39ae39b7c370d8f/parser_test.go#L358-L381 , but I may be wrong.

mfrancisc (commented):
Looks good to me. I have few minor comments related to the test.

mfrancisc:
Seems that it also requires a small change in the e2e tests:
```
Error Trace:	/tmp/toolchain-e2e/test/e2e/parallel/registration_service_test.go:196
        	Error:      	Not equal: 
        	            	expected: "token contains an invalid number of segments"
        	            	actual  : "token is malformed: token contains an invalid number of segments"
```

pkg/auth/tokenparser_test.go
alexeykazakov: The auth field is a string array in the RegisteredClaims: https://pkg.go.dev/github.com/golang-jwt/jwt/v5#ClaimStrings
So we can't set it to a single string on the golang RegisteredClaims struct level.
However when a token is parsed by the jwt v 5 lib it takes into account that the encoded claim can be a string and the parser converts it to a string array to put it into the RegisteredClaims.Audience field.

The only way for us to test it would be to generate a raw token and then parse it but then we would test the jwt library. Not our code. And we don't want to start testing our dependencies. We just need to make sure that we are using the right dependency and this test covers this.

pkg/auth/tokenparser_test.go
alexeykazakov: As I mentioned above this code basically tests that our tokenparser uses the right jwt version. If we roll back our changes from go.mod we are introducing in this PR then the new test would fail.

pkg/auth/tokenparser_test.go
mfrancisc: maybe an e2e test case that set's the `aud` field to a string and one that set's it to a list could help testing this use case. Not sure how much value that can bring.

pkg/auth/tokenparser_test.go
mfrancisc: Right , using the old version would fail because that was expecting `aud` to be a string. My comment was more related to the fact that having a slice with a single string or multiple string doesn't change that much IMO. The field type is the same, but it's just a minor thing.

mfrancisc (approved):
Great job 🚀 

Thanks for the quick fix and for answering my questions.

pkg/auth/tokenparser_test.go
alexeykazakov: The jwt v5 will always generate a token with an array. We can't use it for generating a token with a string with this. We would have to generate and sign a token manually and then hardcode it in the test. Or use the old jwt lib for that. But again that would test the jwt lib itself. I don't think we should go this route of start testing 3rd party libs. Especially when it's quite tricky.

pkg/auth/tokenparser.go
```diff
@@ -30,21 +22,7 @@ type TokenClaims struct {
 	OriginalSub       string `json:"original_sub"`
 	UserID            string `json:"user_id"`
 	AccountID         string `json:"account_id"`
-	jwt.StandardClaims
-}
-
-// Valid checks whether the token claims are valid
-func (c *TokenClaims) Valid() error {
```
alexeykazakov, line 37: The Valid() function was dropped in the Claim interface in v5. So I move the leeway option to the FromString() function below.

pkg/auth/tokenparser.go
```diff
@@ -64,27 +42,32 @@ func NewTokenParser(keyManager *KeyManager) (*TokenParser, error) {
...
+				return nil, errors.New("given key id has unknown type")
+			}
+			// get the public key for kid from keyManager
+			publicKey, err := tp.keyManager.Key(kidStr)
+			if err != nil {
+				return nil, err
+			}
+			return publicKey, nil
+		},
+		jwt.WithLeeway(leeway),
```
alexeykazakov, line 69: This is the leeway option with make our parser to take the leeway window into account when parsing the token instead of relaying the Valid() interface above.

pkg/auth/tokenparser_test.go
```diff
@@ -273,7 +273,27 @@ func (s *TestTokenParserSuite) TestTokenParser() {
 		// validate token
 		_, err = tokenParser.FromString(jwt0string)
 		require.Error(s.T(), err)
-		require.EqualError(s.T(), err, "token is not valid yet")
+		require.EqualError(s.T(), err, "token has invalid claims: token is not valid yet")
+	})
+
+	s.Run("signature is good and token expiration is within leeway", func() {
```
alexeykazakov, line 279: This is a new test to make sure the leeway window is taken into account when we parse the token. It tests that we migrated the leeway thing correctly.

alexeykazakov:
/retest

I've updated the e2e tests to address the failure.

mfrancisc (approved):

PR #396, opened 2024-01-24
fix broken test in pkg/server

filariow:
related to #375

`go test ./pkg/server/server_test.go` is failing consistently with the following error: `"Get \"http://localhost:8081/proxyhealth\": dial tcp [::1]:8081: connect: connection refused"`

After #375, these tests are configured to use a proxy server executing on port `8091`, but healthchecker configuration is hardcoded to `8081`. Changes introduced in this PR allow to configure the healthchecker at runtime.

Signed-off-by: Francesco Ilario <filario@redhat.com>


pkg/server/routes.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 // SetupRoutes registers handlers for various URL paths.
```
alexeykazakov, line 15:
```suggestion
// SetupRoutes registers handlers for various URL paths.
// proxyPort is the API Proxy Server port to be used to setup a route for the health checker for the proxy.
```

alexeykazakov (approved):
Looks good. Just one minor suggestion.

filariow:
> Looks good. Just one minor suggestion.

Thanks, I committed your suggestion

pkg/server/server_test.go
```diff
@@ -37,7 +37,7 @@ func (s *TestServerSuite) TestServer() {
 
 	fake.MockKeycloakCertsCall(s.T())
 	// Setting up the routes.
-	err := srv.SetupRoutes()
+	err := srv.SetupRoutes("8091")
```
rajivnathan, line 40:
```suggestion
	err := srv.SetupRoutes("8091") // uses a different proxy port than the default one to avoid collision with other concurrent tests
```

rajivnathan (approved):
Looks great! Added a minor suggestion. Another minor suggestion is to rename `proxy.ProxyPort` to `proxy.DefaultProxyPort` for added clarity

MatousJobanek (approved):
Nice, thanks for fixing it

cmd/main.go
```diff
@@ -117,7 +117,7 @@ func main() {
 	if err != nil {
 		panic(errs.Wrap(err, "failed to create proxy"))
 	}
-	proxySrv := p.StartProxy()
+	proxySrv := p.StartProxy(proxy.ProxyPort)
```
mfrancisc, line 120: minor and not related to your change - I would drop `Proxy` from the name of the constant, since it's already in the name of the package. And name it something like `proxy.DefaultPort`

mfrancisc (approved):
Looks good 👍

Thanks for fixing it!

cmd/main.go
filariow: thanks for your suggestion, I've renamed that

pkg/server/server_test.go
filariow: thanks for your suggestion, I've applied the change you suggested and renamed the constant to `proxy.DefaultPort`

rajivnathan (approved):

ranakan19 (approved):
Great work!

PR #399, opened 2024-02-04
Refactor remaining service and test code to remove references to old identity properties

sbryzak:
This is an intermediate PR related to the following API PR:

https://github.com/codeready-toolchain/api/pull/393

Since it only changes existing registration service code but doesn't actually update the API dependency it should be safe to merge independent of the other PRs referenced in the above API pull request.

Paired with https://github.com/codeready-toolchain/toolchain-e2e/pull/892


pkg/signup/service/signup_service.go
```diff
@@ -617,7 +585,7 @@ func (s *ServiceImpl) PhoneNumberAlreadyInUse(userID, username, phoneNumberOrHas
 		return errors.NewInternalError(err, "failed listing userSignups")
 	}
 	for _, signup := range userSignups {
-		if signup.Spec.Userid != userID && signup.Spec.Username != username && !states.Deactivated(signup) { // nolint:gosec
+		if signup.Spec.IdentityClaims.Sub != userID && signup.Spec.IdentityClaims.PreferredUsername != username && !states.Deactivated(signup) { // nolint:gosec
```
alexeykazakov, line 588: Let's rename the `userID` param to something like `sub` in the function signature in the line 574, so it's bit less confusing.
```suggestion
		if signup.Spec.IdentityClaims.Sub != sub && signup.Spec.IdentityClaims.PreferredUsername != username && !states.Deactivated(signup) { // nolint:gosec
```

alexeykazakov (approved):

alexeykazakov (approved):
Looks good. Just one minor comment.

But could you please also create a PR for our sre repo to stop using the annotations in the usersignup. Basicaly the same PR as https://github.com/codeready-toolchain/sandbox-sre/pull/1406 but without the go dependency update. And we will need to get that PR merged right after this one so our banning CLI is still working.

PR #400, opened 2024-02-07
add check for ready condition existence

mfrancisc:
When there is not Ready condition or it is incomplete just set it to Unready

### DIFF
--- a/pkg/signup/service/signup_service.go
+++ b/pkg/signup/service/signup_service.go
@@ -453,7 +453,10 @@ func (s *ServiceImpl) DoGetSignup(ctx *gin.Context, provider ResourceProvider, u
 	if err != nil {
 		return nil, errs.Wrap(err, fmt.Sprintf("error when retrieving MasterUserRecord for completed UserSignup %s", userSignup.GetName()))
 	}
-	murCondition, _ := condition.FindConditionByType(mur.Status.Conditions, toolchainv1alpha1.ConditionReady)
+	murCondition, conditionFound := condition.FindConditionByType(mur.Status.Conditions, toolchainv1alpha1.ConditionReady)
+	if !conditionFound {
+		return nil, fmt.Errorf("unable to find condition type: %s", toolchainv1alpha1.ConditionReady)
+	}
 	ready, err := strconv.ParseBool(string(murCondition.Status))
 	if err != nil {
 		return nil, errs.Wrapf(err, "unable to parse readiness status as bool: %s", murCondition.Status)
--- a/pkg/signup/service/signup_service.go
+++ b/pkg/signup/service/signup_service.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"hash/crc32"
 	"regexp"
-	"strconv"
 	"strings"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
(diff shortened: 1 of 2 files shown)


pkg/signup/service/signup_service.go
```diff
@@ -444,7 +444,10 @@ func (s *ServiceImpl) DoGetSignup(ctx *gin.Context, provider ResourceProvider, u
 	if err != nil {
 		return nil, errs.Wrap(err, fmt.Sprintf("error when retrieving MasterUserRecord for completed UserSignup %s", userSignup.GetName()))
 	}
-	murCondition, _ := condition.FindConditionByType(mur.Status.Conditions, toolchainv1alpha1.ConditionReady)
+	murCondition, conditionFound := condition.FindConditionByType(mur.Status.Conditions, toolchainv1alpha1.ConditionReady)
+	if !conditionFound {
+		return nil, fmt.Errorf("unable to find condition type: %s", toolchainv1alpha1.ConditionReady)
```
alexeykazakov, line 449: Is it always an error? What if there is no condition yet because the MUR was just created and takes it time to set the conditions. On the other hand it looks like if there is no condition then it should fail and return an error in line 453 anyway...
You are trying to debug that problem with the reg-service returning an error, right?
Maybe you could look for the `unable to parse readiness status as bool:` error in the logs to see if it's what happened.
But regardless of if the actual reason for that error this code looks buggy. Instead of always depending on the condition to be present and trying to parse it to true/false we should:
- if condition is present and have the status == true then treat the MUR as ready.
- if anything else (it includes, no condition, or status == False or Unknown (anything but True)) then treat is as Not Ready. No errors. Just not ready.

alexeykazakov: OK. I just realized that it's actually exactly the case because the user reported that `unable to parse readiness status as bool:` error. So let's make this code smarter and handle all other cases (condition not present or set to Unknown) gracefully as I mentioned above.

mfrancisc: > - if condition is present and have the status == true then treat the MUR as ready.
> - if anything else (it includes, no condition, or status == False or Unknown (anything but True)) then treat is as Not Ready. No errors. Just not ready.

Sorry but I'm not 100% sure I'm following here, the `DoGetSignup` only builds the `signupResponse` object, it doesn't actually check if the MUR is ready or not, just takes the value of the condition and passes back to the caller. 

Also, from what I can see, I'm not sure where this `signupResponse.Status.Ready` field is being checked, the callers consider MUR/user to be ready if `signupResponse` is not `nil` and it has a `compliantUsername` set, see for example:
https://github.com/codeready-toolchain/registration-service/blob/75f17d8d786695c8a11181268fc3bd56b1f34466/pkg/proxy/service/cluster_service.go#L49
and
https://github.com/codeready-toolchain/registration-service/blob/75f17d8d786695c8a11181268fc3bd56b1f34466/pkg/proxy/handlers/spacelister.go#L52

but maybe I'm missing something.

mfrancisc: @alexeykazakov I've updated the check as discussed today https://github.com/codeready-toolchain/registration-service/pull/400/commits/a17b1531a5f853df5586a5d7835a0ac26f00555b . Could you please take a look when you have some time ?

alexeykazakov (approved):
Looks good! And great tests! Thanks.

rajivnathan (approved):
Nice 👍

PR #401, opened 2024-02-07
reduce readiness failure threshold

mfrancisc:
Reduce number of errors:

```
{"level":"error","ts":"2024-02-07T02:33:06.081Z","logger":"registration-service","msg":"http: Server closed","timestamp":"Wed, 07 Feb 2024 02:33:06 +0000","commit":"b4a49ca","error":"http: Server closed","stacktrace":"github.com/codeready-toolchain/registration-service/pkg/log.(*Logger).Errorf\n\t/home/runner/work/registration-service/registration-service/pkg/log/log.go:183\ngithub.com/codeready-toolchain/registration-service/pkg/log.(*Logger).Error\n\t/home/runner/work/registration-service/registration-service/pkg/log/log.go:169\ngithub.com/codeready-toolchain/registration-service/pkg/log.Error\n\t/home/runner/work/registration-service/registration-service/pkg/log/log.go:111\nmain.main.func2\n\t/home/runner/work/registration-service/registration-service/cmd/main.go:142"}
```

the readiness probe will fail earlier removing the service from load balancing and thus reducing the number of those errors.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -113,7 +113,7 @@ func main() {
 	metricsSrv := proxyMetrics.StartMetricsServer()
 
 	// Start the proxy server
-	p, err := proxy.NewProxy(app, proxyMetrics)
+	p, err := proxy.NewProxy(app, proxyMetrics, cluster.GetMemberClusters)
 	if err != nil {
 		panic(errs.Wrap(err, "failed to create proxy"))
 	}
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
@@ -139,7 +140,11 @@ func main() {
 		log.Infof(nil, "Service Revision %s built on %s", configuration.Commit, configuration.BuildTime)
 		log.Infof(nil, "Listening on %q...", configuration.HTTPAddress)
 		if err := srv.HTTPServer().ListenAndServe(); err != nil {
-			log.Error(nil, err, err.Error())
+			if errors.Is(err, http.ErrServerClosed) {
+				log.Info(nil, err.Error())
+			} else {
+				log.Error(nil, err, err.Error())
+			}
 		}
 	}()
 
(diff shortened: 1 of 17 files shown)


deploy/registration-service.yaml
```diff
@@ -114,7 +114,7 @@ objects:
                 successThreshold: 1
                 timeoutSeconds: 1
               readinessProbe:
-                failureThreshold: 30
+                failureThreshold: 3
```
alexeykazakov, line 117:
```suggestion
                failureThreshold: 1
```

:)

alexeykazakov (commented):
Looks good. But I would set it to 1.

We can't merge it yet though. Let's fix the host operator upgrade first.

metlos:
Please also take into account the discussion on the host-operator PR https://github.com/codeready-toolchain/host-operator/pull/966 that discusses the changes in the registration-service, too.

Sorry for confusing the PRs :)

cmd/main.go
```diff
@@ -139,7 +140,11 @@ func main() {
 		log.Infof(nil, "Service Revision %s built on %s", configuration.Commit, configuration.BuildTime)
 		log.Infof(nil, "Listening on %q...", configuration.HTTPAddress)
 		if err := srv.HTTPServer().ListenAndServe(); err != nil {
-			log.Error(nil, err, err.Error())
+			if errors.Is(err, http.ErrServerClosed) {
+				log.Info(nil, err.Error())
```
rajivnathan, line 144: Should we add some additional context to this log? eg. `The server has closed, this is expected when server shutdown has been initiated.`

rajivnathan (approved):
Nice, I agree it's still good to log something in the case of the `ErrServerClosed` case. Had a small suggestion about the log message to make it a bit more clear why we're logging it as INFO instead of ERROR. But I'll leave it up to you to decide if it's necessary.

cmd/main.go
alexeykazakov: +1 @rajivnathan

alexeykazakov (approved):

cmd/main.go
mfrancisc: makes sense, see [6a877c4](https://github.com/codeready-toolchain/registration-service/pull/401/commits/6a877c4219212c13f1876e8b6d458b4380169797)

thanks for your suggestion!

alexeykazakov (approved):

PR #403, opened 2024-02-13
Optimize binary

filariow:
Removes DWARF and symbol tables from the binary. From [https://pkg.go.dev/cmd/link](https://pkg.go.dev/cmd/link):
- `-s`: Omit the symbol table and debug information
- `-w`: Omit the DWARF symbol table.

Also trims the path in panics so they will start from project's local folder. From [https://pkg.go.dev/cmd/go](https://pkg.go.dev/cmd/go):
- `trimpath`: remove all file system paths from the resulting executable. Instead of absolute file system paths, the recorded file names will begin either a module path@version (when using modules), or a plain import path (when using the standard library, or GOPATH).

As a result binary size is reduced from 63MB to 45MB, and binary inspection and debug are prevented.

### DIFF
--- a/make/build.mk
+++ b/make/build.mk
@@ -3,7 +3,7 @@ GO_PACKAGE_ORG_NAME ?= codeready-toolchain
 GO_PACKAGE_REPO_NAME ?= $(shell basename $$PWD)
 GO_PACKAGE_PATH ?= github.com/${GO_PACKAGE_ORG_NAME}/${GO_PACKAGE_REPO_NAME}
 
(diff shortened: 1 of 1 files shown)


filariow:
@mfrancisc @alexeykazakov thank for your review. I pushed a new commit (de6341d5d867666f4a4c45589854eaa69f30653c) to maintain DWARF and symbol table in the development binary. May you take another look?

alexeykazakov (approved):
OK. Makes sense.

ranakan19 (approved):
LGTM, Thank You for the description - learned something new!

PR #404, opened 2024-02-15
GetSignup: check if user email is banned

mfrancisc:
See: https://redhat-internal.slack.com/archives/CHK0J6HT6/p1707919285542409

Jira: https://issues.redhat.com/browse/SANDBOX-531

### DIFF
--- a/pkg/signup/service/signup_service.go
+++ b/pkg/signup/service/signup_service.go
@@ -488,20 +488,6 @@ func (s *ServiceImpl) DoGetSignup(ctx *gin.Context, provider ResourceProvider, u
 	return signupResponse, nil
 }
 
-func (s *ServiceImpl) isUserEmailBanned(userEmail string) (bool, error) {
-	bannedUsers, err := s.CRTClient().V1Alpha1().BannedUsers().ListByEmail(userEmail)
-	if err != nil {
-		return false, err // error listing banned users
-	}
-
-	for _, bu := range bannedUsers.Items {
-		if bu.Spec.Email == userEmail {
-			return true, nil // user is banned
-		}
-	}
-	return false, nil // user is not banned
-}
-
 // auditUserSignupAgainstClaims compares the properties of the specified UserSignup against the claims contained in the
 // user's access token and updates the UserSignup if necessary.  If updates were made, the function returns true
 // otherwise it returns false.
--- a/pkg/signup/service/signup_service.go
+++ b/pkg/signup/service/signup_service.go
@@ -395,7 +395,7 @@ func (s *ServiceImpl) DoGetSignup(ctx *gin.Context, provider ResourceProvider, u
 	// Query BannedUsers to check the user has not been banned
 	banned, err := s.isUserEmailBanned(userSignup.Spec.IdentityClaims.Email)
 	if err != nil || banned {
-		// return usersignup not found if it's banned
+		// return usersignup not found if it's banned and the error ( if any )
 		return nil, err
 	}
 
--- a/pkg/signup/service/signup_service.go
+++ b/pkg/signup/service/signup_service.go
@@ -91,16 +91,12 @@ func (s *ServiceImpl) newUserSignup(ctx *gin.Context) (*toolchainv1alpha1.UserSi
 	emailHash := hash.EncodeString(userEmail)
 
 	// Query BannedUsers to check the user has not been banned
-	bannedUsers, err := s.CRTClient().V1Alpha1().BannedUsers().ListByEmail(userEmail)
+	banned, err := s.isUserEmailBanned(userEmail)
 	if err != nil {
 		return nil, err
 	}
-
-	for _, bu := range bannedUsers.Items {
-		// If the user has been banned, return an error
-		if bu.Spec.Email == userEmail {
-			return nil, apierrors.NewForbidden(schema.GroupResource{}, "", errs.New("user has been banned"))
-		}
+	if banned {
+		return nil, apierrors.NewForbidden(schema.GroupResource{}, "", errs.New("user has been banned"))
 	}
 
 	verificationRequired, captchaScore := IsPhoneVerificationRequired(s.CaptchaChecker, ctx)
@@ -396,6 +392,13 @@ func (s *ServiceImpl) DoGetSignup(ctx *gin.Context, provider ResourceProvider, u
 		return nil, nil
 	}
 
+	// Query BannedUsers to check the user has not been banned
+	banned, err := s.isUserEmailBanned(userSignup.Spec.IdentityClaims.Email)
+	if err != nil || banned {
+		// return usersignup not found if it's banned
+		return nil, err
+	}
+
 	signupResponse := &signup.Signup{
 		Name:     userSignup.GetName(),
 		Username: userSignup.Spec.IdentityClaims.PreferredUsername,
@@ -479,6 +482,20 @@ func (s *ServiceImpl) DoGetSignup(ctx *gin.Context, provider ResourceProvider, u
 	}
 
 	return signupResponse, nil
+}
+
+func (s *ServiceImpl) isUserEmailBanned(userEmail string) (bool, error) {
+	bannedUsers, err := s.CRTClient().V1Alpha1().BannedUsers().ListByEmail(userEmail)
+	if err != nil {
+		return false, err // error listing banned users
+	}
+
+	for _, bu := range bannedUsers.Items {
+		if bu.Spec.Email == userEmail {
+			return true, nil // user is banned
+		}
+	}
+	return false, nil // user is not banned
 }
 
 // auditUserSignupAgainstClaims compares the properties of the specified UserSignup against the claims contained in the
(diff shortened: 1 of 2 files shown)


pkg/signup/service/signup_service_test.go
```diff
@@ -1608,6 +1608,81 @@ func (s *TestSignupServiceSuite) TestGetSignupReadyConditionStatus() {
...
+	s.ServiceConfiguration(configuration.Namespace(), true, "", 5)
+
+	us := s.newBannedUserSignup()
+	us.Name = service.EncodeUserIdentifier(us.Spec.IdentityClaims.PreferredUsername)
+	err := s.FakeUserSignupClient.Tracker.Add(us)
+	require.NoError(s.T(), err)
+
+	c, _ := gin.CreateTestContext(httptest.NewRecorder())
+
+	toolchainStatus := s.newToolchainStatus(".apps.")
```
xcoulon, line 1622: just curious: why is this part needed?

pkg/signup/service/signup_service_test.go
```diff
@@ -1608,6 +1608,81 @@ func (s *TestSignupServiceSuite) TestGetSignupReadyConditionStatus() {
...
+			},
+		},
+		Spec: toolchainv1alpha1.BannedUserSpec{
+			Email: "jsmith@gmail.com",
+		},
+	})
+	require.NoError(s.T(), err)
+
+	rr := httptest.NewRecorder()
+	ctx, _ := gin.CreateTestContext(rr)
```
xcoulon, line 1645: this looks similar to what's in line 1620

pkg/signup/service/signup_service_test.go
```diff
@@ -1608,6 +1608,81 @@ func (s *TestSignupServiceSuite) TestGetSignupReadyConditionStatus() {
...
+	ctx.Set(context.UsernameKey, "jsmith")
+	ctx.Set(context.SubKey, us.Spec.IdentityClaims.UserID)
+	ctx.Set(context.EmailKey, "jsmith@gmail.com")
+
+	// when
+	response, err := s.Application.SignupService().GetSignup(c, us.Spec.IdentityClaims.UserID, us.Spec.IdentityClaims.PreferredUsername)
```
xcoulon, line 1651: I'm a little confused (but also, not familiar with the code): `response` is a `*signup.Signup`, so why do we need an HTTP recorder in this test?

pkg/signup/service/signup_service_test.go
```diff
@@ -1608,6 +1608,81 @@ func (s *TestSignupServiceSuite) TestGetSignupReadyConditionStatus() {
 	}
 }
 
+func (s *TestSignupServiceSuite) TestGetSignupBannedUserEmail() {
+	// given
+	s.ServiceConfiguration(configuration.Namespace(), true, "", 5)
+
+	us := s.newBannedUserSignup()
+	us.Name = service.EncodeUserIdentifier(us.Spec.IdentityClaims.PreferredUsername)
```
xcoulon, line 1616: this could be done within `s.newBannedUserSignup()`, couldn't it?

pkg/signup/service/signup_service_test.go
mfrancisc: Good catch 🤦‍♂️ 

Cleaned in https://github.com/codeready-toolchain/registration-service/pull/404/commits/72752b29c173cfe6137f0ac2a6b56c58a02707c6

pkg/signup/service/signup_service_test.go
mfrancisc: My understanding is that we need the recorder in order to create the gin context only, so it's just a dependency not actually used. There may be a cleaner way to set this up, but I'm not aware of any atm.

pkg/signup/service/signup_service_test.go
mfrancisc: Actually it doesn't seem useful at all, unless I'm missing something.

Cleaned in https://github.com/codeready-toolchain/registration-service/pull/404/commits/72752b29c173cfe6137f0ac2a6b56c58a02707c6

pkg/signup/service/signup_service_test.go
```diff
@@ -1608,6 +1608,74 @@ func (s *TestSignupServiceSuite) TestGetSignupReadyConditionStatus() {
...
+	require.NoError(s.T(), err)
+
+	rr := httptest.NewRecorder()
+	ctx, _ := gin.CreateTestContext(rr)
+	ctx.Set(context.UsernameKey, "jsmith")
+	ctx.Set(context.SubKey, us.Spec.IdentityClaims.UserID)
+	ctx.Set(context.EmailKey, "jsmith@gmail.com")
+
+	// when
+	response, err := s.Application.SignupService().GetSignup(ctx, us.Spec.IdentityClaims.UserID, us.Spec.IdentityClaims.PreferredUsername)
```
rajivnathan, line 1644: It looks like if you comment out the code that adds the BannedUser to the `FakeBannedUserClient` then it doesn't cause the test to fail. I thought the test should fail on the `require.Nil(s.T(), response)` if there's no BannedUser.

Probably good to change the test to do 2 calls to GetSignup:

1. With just UserSignup and verify `require.NotNil(s.T(), response)`
2. With UserSignup and BannedUser and verify `require.Nil(s.T(), response)`

pkg/signup/service/signup_service.go
```diff
@@ -396,6 +392,13 @@ func (s *ServiceImpl) DoGetSignup(ctx *gin.Context, provider ResourceProvider, u
 		return nil, nil
 	}
 
+	// Query BannedUsers to check the user has not been banned
+	banned, err := s.isUserEmailBanned(userSignup.Spec.IdentityClaims.Email)
```
alexeykazakov, line 396: Alternatively you could just check the Complete condition reason here: https://github.com/codeready-toolchain/registration-service/blob/master/pkg/signup/service/signup_service.go#L438
Something like:
```
	} else if completeCondition.Reason == toolchainv1alpha1.UserSignupUserBannedReason {
		log.Info(nil, fmt.Sprintf("usersignup: %s is banned", userSignup.GetName()))
		// UserSignup is banned
		...
		return ...
	}
```
It would be fewer API requests and easier to test in unit tests I guess.

The question is what we actually want to return in the case of a banned user. If we return `nil` as you do in this PR then UI will treat it as a new user and will show user the "Signup" button which will fail if user clicks on it.
Not ideal.
Maybe we should return something like this instead:

```
		signupResponse.Status = signup.Status{
			Reason:               toolchainv1alpha1.UserSignupUserBannedReason,
		}
		return signupResponse, nil
```
That will be treated be UI as an existing users which is not yet provisioned. And I guess it will show a spinner...
Or we could use `toolchainv1alpha1.UserSignupPendingApprovalReason` and it will showed by UI as "Pending approval".

WDYT guys?

mfrancisc: > Alternatively you could just check the Complete condition reason

Good point, TBH checking the banned resource seemed easy enough, a bit less to test I agree. I can change it if it looks more obvious and there is low chance that the Condition will be dropped or changed.

> The question is what we actually want to return in the case of a banned user. If we return nil as you do in this PR then UI will treat it as a new user and will show user the "Signup" button which will fail if user clicks on it.

I think that we return `nil` response ( 404 Not found ) right now for banned users right ? So I didn't want to change the behaviour, just avoid logging the error when banned users try to login.

mfrancisc: BTW if we do return a specific response, like the ones @alexeykazakov suggested, would make the test more reliable since we don't have to check the `nil` response which may also be caused by usersignup not found.

rajivnathan: Good point about relying on UserSignup status. 👍 I like the idea of returning a response with `toolchainv1alpha1.UserSignupPendingApprovalReason` so that they just see the spinner and cannot keep retrying signup.

pkg/signup/service/signup_service_test.go
mfrancisc: Yeah, good catch, seems like the usersignup is not found , so it doesn't even get to the ban check anymore 🤔 

I think we may want to return a specific reason in the response instead of `nil` as we are discussing the the other thread here https://github.com/codeready-toolchain/registration-service/pull/404#discussion_r1491414417. This will make the test more reliable as well.

pkg/signup/service/signup_service.go
mfrancisc: > The current code does not return 404 if there is no MUR as I can see. It returns 500

You're right, but 500 is even less correct IMO. Since this is intentional and not a server error. I like the idea of returning a specific reason in case of user banned. So let's go with the Pending Approval if there is no objection.

mfrancisc: Please take a look at https://github.com/codeready-toolchain/registration-service/pull/404/commits/ceec4c6db058ca0245c1ce52e6e3e9610a641716 where I'm relying on the Usersignup condition and not anymore on the Banned resource, I've also updated the response to return `UserSignupPendingApprovalReason`.

alexeykazakov (approved):
Looks great now! Thank you for taking care of this!

rajivnathan (approved):
Looks great! 👍

PR #405, opened 2024-02-26
Made UserSignup.Spec.IdentityClaims property required

sbryzak:
Associated API PR: https://github.com/codeready-toolchain/api/pull/400
E2E tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/900

### DIFF
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -178,7 +178,7 @@ func (r AuthConfig) AuthClientConfigRaw() string {
 }
 
 func (r AuthConfig) AuthClientPublicKeysURL() string {
-	return commonconfig.GetString(r.c.AuthClientPublicKeysURL, "https://sso.devsandbox.dev/realms/sandbox-dev/protocol/openid-connect/certs")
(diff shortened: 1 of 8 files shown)


alexeykazakov (approved):

mfrancisc (approved):

pkg/configuration/configuration.go
```diff
@@ -178,7 +178,7 @@ func (r AuthConfig) AuthClientConfigRaw() string {
 }
 
 func (r AuthConfig) AuthClientPublicKeysURL() string {
-	return commonconfig.GetString(r.c.AuthClientPublicKeysURL, "https://sso.devsandbox.dev/auth/realms/sandbox-dev/protocol/openid-connect/certs")
```
alexeykazakov, line 181: Is it a new keycloak thing? Or specific to our dev sso configuration?
We do rely on `auth/realms` path in a few places which is consistent with our production sso.redhat.com.
If we have a different path in dev sso then it won't work OOTB.
See https://github.com/search?q=repo%3Acodeready-toolchain%2Fregistration-service+%22auth%2Frealms%22&type=code

alexeykazakov (changes requested):
Please see my comment above regarding the "auth/realm" path.

alexeykazakov (approved):

PR #411, opened 2024-03-11
Introduce user's subscription metadata into GetUserSignup response

sbryzak:
This PR adds the start date, end date and days remaining properties into the response provided by the GetUserSignup endpoint.

Fixes https://issues.redhat.com/browse/SANDBOX-554

Host Operator PR that grants registration service newly required permissions: https://github.com/codeready-toolchain/host-operator/pull/988

Tests incoming.


pkg/signup/service/signup_service.go
```diff
@@ -461,6 +465,30 @@ func (s *ServiceImpl) DoGetSignup(ctx *gin.Context, provider ResourceProvider, u
...
 	}
+
+	if mur.Status.ProvisionedTime != nil {
+		startDateValue := mur.Status.ProvisionedTime.Format(ISO8601Format)
+		signupResponse.StartDate = &startDateValue
+
+		// #### The rest of this code only makes sense when there is a ProvisionedTime set, so we include it in the same block ####
+
+		// Lookup the user's tier
+		tier, err := provider.GetUserTier(mur.Spec.TierName)
```
alexeykazakov, line 476: Let's use the informer and informer service to get the tier. We also can switch to the informer service to obtain the mur: https://github.com/codeready-toolchain/registration-service/blob/master/pkg/informers/service/informer_service.go

It should work much faster. Thanks to the informer cache.

And don't forget to update the reg-service SA's role so it has permissions to list/get/watch tiers.
For dev environment: https://github.com/codeready-toolchain/registration-service/blob/master/deploy/registration-service.yaml#L33
For prod & e2e-tests: https://github.com/codeready-toolchain/host-operator/blob/master/deploy/registration-service/registration-service.yaml#L33

sbryzak: Thanks for the quick initial review, I've implemented the UserTier informer support, and made all the changes you suggested.  Host operator PR with the new permissions is here: https://github.com/codeready-toolchain/host-operator/pull/988

alexeykazakov (commented):
I was wondering if instead of calculating the life-cycle dates every time the reg-service API is used we should set this data in the UserSignup status in the deactivation controller?
https://github.com/codeready-toolchain/host-operator/blob/master/controllers/deactivation/deactivation_controller.go#L118

The deactivation controller already calculates it and manages the deactivation/deactivating statuses. So should we move this logic there too?

Then in the reg-service we would just read it from the status. Another advantage of having it in the status is that we could use it in other places like sandbox-cli. So it could print the dates when doing `sandbox-cli get usersignup`.

WDYT?

sbryzak:
@alexeykazakov I like this approach, it simplifies things and as you mention makes the information conveniently available in other places.  I will work on making this change today, and undo my changes made to the informer API (which caused some serious breakage for the tests which I still haven't been able to resolve).

xcoulon:
> I was wondering if instead of calculating the life-cycle dates every time the reg-service API is used we should set this data in the UserSignup status in the deactivation controller? https://github.com/codeready-toolchain/host-operator/blob/master/controllers/deactivation/deactivation_controller.go#L118
> 
> The deactivation controller already calculates it and manages the deactivation/deactivating statuses. So should we move this logic there too?
> 
> Then in the reg-service we would just read it from the status. Another advantage of having it in the status is that we could use it in other places like sandbox-cli. So it could print the dates when doing `sandbox-cli get usersignup`.
> 
> WDYT?

what about moving the logic to compute the deactivation time into `toolchain-common`, so we can use it from the registration service and the deactivation controller? My concern is that the proposed [UserSignup.Status.DeactivationTimestamp](https://github.com/codeready-toolchain/api/pull/404/files#diff-be565c332afe4a34d4ae73641e22261dce24bfac3d254b50a295447ae8ecb318R255) will need to be updated when a user is moved to another tier. 
In other words, is there a substantial gain of performance compared to the cost of having to deal with the occasional updates?

alexeykazakov:
> My concern is that the proposed [UserSignup.Status.DeactivationTimestamp](https://github.com/codeready-toolchain/api/pull/404/files#diff-be565c332afe4a34d4ae73641e22261dce24bfac3d254b50a295447ae8ecb318R255) will need to be updated when a user is moved to another tier.

That should not be necessary because the DeactivationController watches both MURs and UserSignups. So when the user is moved to another tier the corresponding MUR is updated. It will trigger the DeactivationController which will update the UserSignup status.

rajivnathan:
Yeah, I like the idea of doing the calculation in the deactivation controller as well.

The issue with calculating it in the reg svc as part of DoGetSignup() is that there's some unnecessary overhead. You'd need to check whether the UserSignup is complete, if so then get the UserTier and calculate the values even though the MUR in most cases would not have changed since the last time the API was called. I'm assuming this API is going to be used in a user-facing dashboard so I'd imagine it'll be called frequently so the less overhead the better.

As Alexey mentioned the deactivation controller already watches both the UserSignup and MUR so it would update the status as soon as the MUR is created/updated and there's no overhead.

xcoulon:
> > My concern is that the proposed [UserSignup.Status.DeactivationTimestamp](https://github.com/codeready-toolchain/api/pull/404/files#diff-be565c332afe4a34d4ae73641e22261dce24bfac3d254b50a295447ae8ecb318R255) will need to be updated when a user is moved to another tier.
> 
> That should not be necessary because the DeactivationController watches both MURs and UserSignups. So when the user is moved to another tier the corresponding MUR is updated. It will trigger the DeactivationController which will update the UserSignup status.

ok, makes sense!

sbryzak:
Renaming branch so that it doesn't conflict with host operator/e2e branches

PR #415, opened 2024-03-19
optimizing EncodeUserIdentifier

filariow:
This PR contains two commits with some optimizations for EncodeUserIdentifier:

#### 66bcbd4 **move regex compilation out of EncodeUserIdentifier**

Moves the regular expressions compilation out of EncodeUserIdentifier.
This means the Regular Expressions are now compiled just when the package is imported, and not all the time the function is invoked.

#### a6a28ec **remove regexp from EncodeUserIdentifier**

A lighter approach that avoids using Regular Expressions.

### DIFF
--- a/pkg/signup/service/signup_service.go
+++ b/pkg/signup/service/signup_service.go
@@ -30,10 +30,7 @@ import (
 )
 
 const (
-	DNS1123NameMaximumLength         = 63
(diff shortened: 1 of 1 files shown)


rajivnathan (approved):
Nice! More efficient 👍

alexeykazakov (approved):
:+1:

mfrancisc (approved):

PR #432, opened 2024-05-15
Calculate and return days remaining in get signup response

sbryzak:
If the UserSignup.Status.ScheduledDeactivationTimestamp property is set, then use it to calculate the number of days remaining in the user's trial and return that value in the get signup response.

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/995


pkg/signup/service/signup_service.go
```diff
@@ -450,6 +456,16 @@ func (s *ServiceImpl) DoGetSignup(ctx *gin.Context, provider ResourceProvider, u
 		return signupResponse, nil
 	}
 
+	if userSignup.Status.ScheduledDeactivationTimestamp != nil {
```
MatousJobanek, line 459: there is no difference if the field is nil or zero 
```suggestion
	if !userSignup.Status.ScheduledDeactivationTimestamp.IsZero() {
```

pkg/signup/service/signup_service.go
```diff
@@ -467,6 +483,23 @@ func (s *ServiceImpl) DoGetSignup(ctx *gin.Context, provider ResourceProvider, u
...
+
+		// #### The rest of this code only makes sense when there is a ProvisionedTime set, so we include it in the same block ####
+
+		if !userSignup.Status.ScheduledDeactivationTimestamp.IsZero() {
+			endDateValue := userSignup.Status.ScheduledDeactivationTimestamp.Format(ISO8601Format)
+			signupResponse.EndDate = &endDateValue
+
+			// Calculate the number of days remaining
+			daysRemaining := time.Until(userSignup.Status.ScheduledDeactivationTimestamp.Time).Hours() / 24
+			signupResponse.DaysRemaining = &daysRemaining
```
MatousJobanek, line 499: this is already calculated above, why do you need to recalculate it here again?

pkg/signup/service/signup_service_test.go
```diff
@@ -1158,6 +1160,9 @@ func (s *TestSignupServiceSuite) TestGetSignupStatusOK() {
 			require.Equal(t, us.Name, response.Name)
 			require.Equal(t, "jsmith", response.Username)
 			require.Equal(t, "ted", response.CompliantUsername)
+			require.NotNil(t, response.DaysRemaining)
+			require.Exactly(t)
+			require.Equal(t, float64(30), *response.DaysRemaining)
```
MatousJobanek, line 1165: the `StartDate` and `EndDate` should be set as well, right? Could you please verify it?

pkg/signup/service/signup_service_test.go
```diff
@@ -1241,6 +1246,28 @@ func (s *TestSignupServiceSuite) TestGetSignupByUsernameOK() {
...
+		}
+		return nil, apierrors.NewNotFound(schema.GroupResource{}, name)
+	}
+	inf.GetMurFunc = func(name string) (*toolchainv1alpha1.MasterUserRecord, error) {
+		if name == mur.Name {
+			return mur, nil
+		}
+		return nil, apierrors.NewNotFound(schema.GroupResource{}, name)
+	}
+	s.Application.MockInformerService(inf)
```
MatousJobanek, line 1262: I don't think that we need anything from the fake informers to be set up
```suggestion
```

pkg/signup/signup.go
```diff
@@ -40,6 +40,12 @@ type Signup struct {
 	// Company from the Identity Provider
 	Company string `json:"company"`
 	Status  Status `json:"status,omitempty"`
+	// StartDate is the date that the user's current subscription started, in ISO8601 format
+	StartDate *string `json:"start_date,omitempty"`
+	// End Date is the date that the user's current subscription will end, in ISO8601 format
+	EndDate *string `json:"end_date,omitempty"`
+	// DaysRemaining is a float representing the number of days remaining in the user's subscription
+	DaysRemaining *float64 `json:"days_remaining,omitempty"`
```
MatousJobanek, line 48: we use camel-case all over the place, not sure why you want to use snake-case in this particular situation
```suggestion
	StartDate *string `json:"startDate,omitempty"`
	// End Date is the date that the user's current subscription will end, in ISO8601 format
	EndDate *string `json:"endDate,omitempty"`
	// DaysRemaining is a float representing the number of days remaining in the user's subscription
	DaysRemaining *float64 `json:"daysRemaining,omitempty"`
```

MatousJobanek: apart from that, why do you want to return the fields as pointers? This is not really need, is it?

MatousJobanek (commented):
I haven't seen any PR for the changes in toolchain-e2e, could you please create one?

pkg/signup/service/signup_service_test.go
```diff
@@ -1158,6 +1160,9 @@ func (s *TestSignupServiceSuite) TestGetSignupStatusOK() {
 			require.Equal(t, us.Name, response.Name)
 			require.Equal(t, "jsmith", response.Username)
 			require.Equal(t, "ted", response.CompliantUsername)
+			require.NotNil(t, response.DaysRemaining)
+			require.Exactly(t)
```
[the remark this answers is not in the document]
sbryzak: IDE hiccup, fixed.

pkg/signup/service/signup_service.go
```diff
@@ -34,6 +37,9 @@ const (
 
 	// NoSpaceKey is the query key for specifying whether the UserSignup should be created without a Space
 	NoSpaceKey = "no-space"
+
+	// ISO8601Format is used to format date values returned to the client
+	ISO8601Format = "2006-01-02T15:04:05.000Z07:00"
```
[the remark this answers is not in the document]
sbryzak: That's fine, I believe RFC3339 is a subset of ISO8601 so it shouldn't make a difference client-side.

pkg/signup/service/signup_service.go
```diff
@@ -467,6 +483,23 @@ func (s *ServiceImpl) DoGetSignup(ctx *gin.Context, provider ResourceProvider, u
...
 	}
+
+	if mur.Status.ProvisionedTime != nil {
+		startDateValue := mur.Status.ProvisionedTime.Format(ISO8601Format)
+		signupResponse.StartDate = &startDateValue
+
+		// #### The rest of this code only makes sense when there is a ProvisionedTime set, so we include it in the same block ####
+
+		if !userSignup.Status.ScheduledDeactivationTimestamp.IsZero() {
+			endDateValue := userSignup.Status.ScheduledDeactivationTimestamp.Format(ISO8601Format)
```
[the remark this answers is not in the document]
sbryzak: Fixed

pkg/signup/service/signup_service.go
sbryzak: There's actually a big difference, if you call a function on a nil pointer it will error.

pkg/signup/service/signup_service.go
sbryzak: Oops, major oversight.. I must have written the above code previously and forgotten about it.

pkg/signup/service/signup_service_test.go
sbryzak: Right, fixed.

pkg/signup/signup.go
sbryzak: You're right, it looks like `omitempty` will still work if a string is equal to its default value, i.e. empty.  Fixed.

pkg/signup/service/signup_service_test.go
sbryzak: You're right, looks like this was unnecessary.

pkg/signup/service/signup_service.go
MatousJobanek: > if you call a function on a nil pointer it will error.
that's not true

in this particular case, if it's the standard `time.Time`, then yes, it would panic because the method doesn't check for nil
In the case of the time from the `k8s.io/apimachinery/pkg/apis/meta/v1` package, the function actually check if the pointer is empty or not  - see https://github.com/kubernetes/apimachinery/blob/703232ea6da48aed7ac22260dabc6eac01aab896/pkg/apis/meta/v1/time.go#L61-L66

```go
// IsZero returns true if the value is nil or time is zero.
func (t *Time) IsZero() bool {
	if t == nil {
		return true
	}
	return t.Time.IsZero()
}
```

pkg/signup/signup.go
```diff
@@ -40,6 +40,12 @@ type Signup struct {
 	// Company from the Identity Provider
 	Company string `json:"company"`
 	Status  Status `json:"status,omitempty"`
+	// StartDate is the date that the user's current subscription started, in ISO8601 format
+	StartDate string `json:"startDate,omitempty"`
+	// End Date is the date that the user's current subscription will end, in ISO8601 format
```
MatousJobanek, line 45: You are still mentioning the `ISO8601`, but it's actually `RFC3339`

pkg/signup/signup.go
```diff
@@ -40,6 +40,12 @@ type Signup struct {
...
+	EndDate string `json:"endDate,omitempty"`
+	// DaysRemaining is a float representing the number of days remaining in the user's subscription
+	DaysRemaining *float64 `json:"daysRemaining,omitempty"`
```
MatousJobanek, line 48: Why do you still keep the pointer for`DaysRemaining` ?

MatousJobanek (commented):
still waiting for the changes in e2e tests

pkg/signup/service/signup_service.go
sbryzak: The `ScheduledDeactivationTimestamp` property itself can be nil, which means it will definitely error if you try to call a function on it.

pkg/signup/signup.go
sbryzak: Fixed

pkg/signup/signup.go
sbryzak: A pointer is correct here.  Otherwise how would you differentiate between `DaysRemaining` not being set (i.e. nil) and `DaysRemaining` actually having a value of `0` ?

pkg/signup/service/signup_service.go
MatousJobanek: again, it won't fail! Please read my comment above and the link to understand why it doesn't fail.
Also, you can try it on your own - just run:
```go
func TestTimestampIsNil(t *testing.T) {
	userSignup := toolchainv1alpha1.UserSignup{}

	require.Nil(t, userSignup.Status.ScheduledDeactivationTimestamp)
	require.True(t, userSignup.Status.ScheduledDeactivationTimestamp.IsZero())
}
```

MatousJobanek: and if you still don't trust me, you can try this:
https://play.golang.com/p/HE0cUu5aGev
so you can really call a function on a pointer that is nil - nil is just a default value of a pointer. It starts failing as soon as you try to touch some of the fields of the type because the memory address it could read it from is invalid.

pkg/signup/signup.go
MatousJobanek: Ok, could you please update the comment so it's clear?

pkg/signup/service/signup_service.go
sbryzak: My apologies, you are correct - I had actually tried this change before however I missed the negate operator at the start of the line so it did actually fail (as the code within the block is attempting to access properties of the object):

```
        --- FAIL: TestRunSignupServiceSuite/TestGetSignupStatusNotComplete/informer_-_with_no_check_for_UserSignup_complete_condition (0.38s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x2174c50]

goroutine 81 [running]:
testing.tRunner.func1.2({0x261f240, 0x3af7c10})
	/home/shane/go/go1.22.2/src/testing/testing.go:1631 +0x49e
testing.tRunner.func1()
	/home/shane/go/go1.22.2/src/testing/testing.go:1634 +0x669
panic({0x261f240?, 0x3af7c10?})
	/home/shane/go/go1.22.2/src/runtime/panic.go:770 +0x136
github.com/codeready-toolchain/registration-service/pkg/signup/service.(*ServiceImpl).DoGetSignup(0xc000ddfad0, 0xc0001c0700, {0x2d802f8, 0xc000c607c0}, {0xc000863740, 0x24}, {0x2886f34, 0x4}, 0x0)
	/home/shane/go/src/github.com/codeready-toolchain/registration-service/pkg/signup/service/signup_service.go:464 +0x1210
github.com/codeready-toolchain/registration-service/pkg/signup/service.(*ServiceImpl).GetSignupFromInformer(0xc000ddfad0, 0xc0001c0700, {0xc000863740, 0x24}, {0x2886f34, 0x4}, 0x0)
	/home/shane/go/src/github.com/codeready-toolchain/registration-service/pkg/signup/service/signup_service.go:368 +0x153
github.com/codeready-toolchain/registration-service/pkg/signup/service_test.(*TestSignupServiceSuite).TestGetSignupStatusNotComplete.func2(0xc000909040)
	/home/shane/go/src/github.com/codeready-toolchain/registration-service/pkg/signup/service/signup_service_test.go:938 +0x763
testing.tRunner(0xc000909040, 0xc000c60680)
	/home/shane/go/go1.22.2/src/testing/testing.go:1689 +0x1da
created by testing.(*T).Run in goroutine 82
	/home/shane/go/go1.22.2/src/testing/testing.go:1742 +0x7d3

Debugger finished with the exit code 0
```

Once I noticed my typo and fixed it all the tests passed.  I don't actually like how the new line reads - "not is zero" is not as immediately easy to visually parse as "value != nil" in my opinion, however the `IsZero()` function does also provide us with the other check for an actual zero time which I guess is nice to have.

pkg/signup/signup.go
sbryzak: updated

alexeykazakov: Why using float64 and not int or uint? Also I wonder if we really need days remaining... EndDate should be enough, right? WDYT?

alexeykazakov (commented):
Looks good to me but I would get rid of the DaysRamaining thing and let the UI calculate it from the EndDate if needed.

sbryzak:
DaysRemaining is now removed.

pkg/signup/service/signup_service_test.go
```diff
@@ -1235,13 +1239,25 @@ func (s *TestSignupServiceSuite) TestGetSignupByUsernameOK() {
 
 	us := s.newUserSignupComplete()
 	us.Name = service.EncodeUserIdentifier(us.Spec.IdentityClaims.PreferredUsername)
+	// Set the scheduled deactivation timestamp 1 day ago
+	us.Status.ScheduledDeactivationTimestamp = util.Ptr(v1.NewTime(time.Now().Add(-time.Hour * 24)))
```
MatousJobanek, line 1243: could you please explain why you set time 1 day ago? it's more logical to set the deactivation time in the future instead, right?

pkg/signup/service/signup_service_test.go
```diff
@@ -1257,12 +1273,22 @@ func (s *TestSignupServiceSuite) TestGetSignupByUsernameOK() {
...
 
+	// Confirm the StartDate is the same as the provisionedTime
+	require.Equal(s.T(), provisionedTime.Format(time.RFC3339), response.StartDate)
+
+	// Confirm the end date is about 1 day ago
+	responseEndDate, err := time.Parse(time.RFC3339, response.EndDate)
+	require.NoError(s.T(), err)
+	expectedEndDate := time.Now().Add(-time.Hour * 24)
+	require.WithinDuration(s.T(), expectedEndDate, responseEndDate, time.Minute*5,
+		"endDate in response [%s] not in expected range [%s]", responseEndDate, expectedEndDate.Format(time.RFC3339))
```
MatousJobanek, line 1290: why do we need to verify that the time is "about" 1 day ago? We do set the `ScheduledDeactivationTimestamp` to a specific time, and the value is copied from the UserSignup status, so we can use `require.Equal` expecting the same timestamp to be set also in the Signup response.
```suggestion
	// Confirm the end date is the same as ScheduledDeactivationTimestamp
	require.Equal(s.T(), us.Status.ScheduledDeactivationTimestamp.Format(time.RFC3339), response.EndDate)
```

MatousJobanek (commented):
I don't see any unit test that would verify the case when the ScheduledDeactivationTimestamp/ProvisionedTime is not set - we need to make sure that the empty string is returned in this case.
Sorry if I missed it somewhere.

pkg/signup/service/signup_service_test.go
sbryzak: It made more sense previously before I removed the DaysRemaining field from the response, now since we no longer need to test this functionality I can set the time in the future.  Updated.

pkg/signup/service/signup_service_test.go
sbryzak: Updated to check exact timestamp.

sbryzak:
Added check to existing test to ensure `StartDate` / `EndDate` are now blank when user is not active.

pkg/signup/service/signup_service_test.go
```diff
@@ -1234,13 +1241,26 @@ func (s *TestSignupServiceSuite) TestGetSignupByUsernameOK() {
...
 	us.Name = service.EncodeUserIdentifier(us.Spec.IdentityClaims.PreferredUsername)
+	// Set the scheduled deactivation timestamp 1 day in the future
+	deactivationTimestamp := time.Now().Add(time.Hour * 24).Round(time.Second).UTC()
+	us.Status.ScheduledDeactivationTimestamp = util.Ptr(v1.NewTime(deactivationTimestamp))
 	err := s.FakeUserSignupClient.Tracker.Add(us)
 	require.NoError(s.T(), err)
 
 	mur := s.newProvisionedMUR("ted")
+	// Set the provisioned time 29 days in the past
+	provisionedTime := time.Now().Add(-time.Hour * 24 * 31).Round(time.Second)
```
MatousJobanek, line 1252: You changed the comment, but not the value
```suggestion
	provisionedTime := time.Now().Add(-time.Hour * 24 * 29).Round(time.Second)
```

pkg/signup/service/signup_service_test.go
```diff
@@ -1256,12 +1276,20 @@ func (s *TestSignupServiceSuite) TestGetSignupByUsernameOK() {
...
+	response, err := svc.GetSignup(c, "foo", us.Spec.IdentityClaims.PreferredUsername)
 
 	// then
 	require.NoError(s.T(), err)
 	require.NotNil(s.T(), response)
 
+	// Confirm the StartDate is the same as the provisionedTime
+	require.Equal(s.T(), provisionedTime.UTC().Format(time.RFC3339), response.StartDate)
+
+	// Confirm the end date is about 1 day ago
```
MatousJobanek, line 1288:
```suggestion
	// Confirm the end date is the same as the deactivationTimestamp
```

MatousJobanek (approved):
Nice :+1: Thanks for the extra changes :pray:

pkg/signup/service/signup_service_test.go
sbryzak: Fixed, thanks.

pkg/signup/service/signup_service_test.go
sbryzak: Oops, my fault. Thanks for picking this up.

alexeykazakov (approved):

PR #434, opened 2024-06-13
Add captcha assessment ID to UserSignup

rajivnathan:
https://issues.redhat.com/browse/SANDBOX-626

Adds captcha assessment ID to UserSignup so that assessments can be annotated later on. eg. when a user is banned


alexeykazakov (approved):
Looks good!

MatousJobanek (approved):

PR #436, opened 2024-06-19
feat: export http server metrics

xcoulon:
expose metrics for registration service on a separate server/port
to avoid making them available from the current Route

include a unit test to verify that the new
`sandbox_promhttp_*` metrics are exposed on the
`/metrics` endpoint

see also https://github.com/codeready-toolchain/host-operator/pull/1051 (new `Service` in the resource manifest)
see also https://github.com/codeready-toolchain/toolchain-e2e/pull/1001 (e2e tests) 


pkg/server/metrics_server.go
```diff
@@ -0,0 +1,44 @@
...
+			NextProtos: []string{"http/1.1"}, // disable HTTP/2 for now
+		},
+	}
+	go func() {
+		// service connections
+		if err := srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			log.Error(err, err.Error())
+		}
+	}()
+	return srv, router
```
MatousJobanek, line 43: Correct me if I'm wrong, the original implementation relied on `echo`, which was wrong because we use `gin` in reg-service (echo is used in proxy), so you changed the logic to use `gin` here as well, is that correct?

pkg/middleware/promhttp_middleware.go
```diff
@@ -0,0 +1,49 @@
...
+			duration := time.Since(start)
+			histVec.With(prometheus.Labels{
+				"code":   strconv.Itoa(c.Writer.Status()),
+				"method": c.Request.Method,
+				"path":   c.Request.URL.Path,
+			}).Observe(float64(duration.Milliseconds()))
+		}()
+		c.Next()
+	}
+}
```
MatousJobanek, line 49: they are used in registration-service endpoints only, is that correct? Just asking :wink: We don't need anything like that in proxy

MatousJobanek (approved):
:+1:

pkg/server/metrics_server.go
xcoulon: correct: Proxy uses `echo` and so does it companion server for its own metrics, whereas registration-service and its new companion server for its metrics both use `gin`

pkg/middleware/promhttp_middleware.go
xcoulon: yes, we only uses this middleware in the registration-service for now. We could add something similar for `echo` if we want to collect the same metrics in proxy.

PR #438, opened 2024-06-24
Add static redirect page to sso and then to requested page

aneelac22:
[RHCLOUD-32115]


pkg/assets/static/redirectpage.js
```diff
@@ -0,0 +1,53 @@
+const queryString = window.location.search;
+const urlParams = new URLSearchParams(queryString);
+const link = urlParams.get("link");
+const keyword = urlParams.get("keyword");
+const selectedId = urlParams.get("selectedId");
+const consoleUrl = "https://console.redhat.com/openshift/sandbox";
```
alexeykazakov, line 6: We will probably need to move all specific URLs like console.redhat.com to a configuration so it can be configurable in different sandbox instances but we can keep it as is for now.

MatousJobanek: or we could include it in the response of the `/api/v1/signup` endpoint - we should have these links available there, right?

alexeykazakov: Yes, the configuration itself is exposed to the reg-page through our reg-service `/api/v1`. So UI doesn't have to get access to the configuration directly.

aneelac22: I can use consoleUrl from the JSON data but, when it is not ready or when there is an error I would have get it from other config maybe

pkg/assets/static/redirectpage.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
+<html lang="en" class="pf-m-redhat-font">
+    <head>
+        <meta charset="utf-8" />
+        <meta name="viewport" content="width=device-width, initial-scale=1" />
+        <title>OpenShift Developer Sandbox</title>
+      </head>
+      <body>
+      </body>
+      <script src="redirectpage.js"></script>
```
alexeykazakov, line 10: Doesn't seem to be working...

This is what I tried: `{baseURL}/redirectpage.html?link="import"`
And I expect to get redirected to `{consoleURL}/import/ns/alexeykazakov-dev`
But the JS failed with:
```
redirectpage.js:19 Uncaught SyntaxError: Unexpected token 'export'
jquery.min.js:14 [Deprecation] Listener added for a 'DOMSubtreeModified' mutation event. Support for this event type has been removed, and this event will no longer be fired. See https://chromestatus.com/feature/5083947249172480 for more information.
add @ jquery.min.js:14
```

redirectpage-tests/README.adoc
```diff
@@ -0,0 +1,27 @@
...
+    npm run test
+    ``` 
+
+== Implementation guidelines
+
+1. When user clicks on a catalog item on developers.redhat.com it redirects to this url format
+    `${baseUrl}?link="catalogName"&keyword="keyword"&selectedId="selectedId"`
+2. If the user is not authenticated, it redirects to sso page for user to login
+3. Once the user is authenticated it redirects to the page user wants to access which is in the format
+    `${consoleUrl}/${link}/ns/${defaultUserNamespace}?keyword="keyword"&selectedId="selectedId"` with
```
alexeykazakov, line 21: Hm...
What URL should I use if I want to get redirected to
```
{consoleURL}/catalog/ns/alexeykazakov-dev?category=databases&selectedId=8ef58d8f-edcc-4284-96bb-5006360b6e73
```
Where `alexeykazakov-dev` is my default namespace.
What should I use for `keyword` to get `category=databases` as the result?

aneelac22: In the google doc there is no URL with category? Am I missing something?

alexeykazakov: Yeah, you are rigth.
But could you please clarify what URL should be used to access
```
{consoleURL}/catalog/ns/alexeykazakov-dev?keyword=jboss&selectedId=openshift-helm-charts--https%3A%2F%2Fgithub.com%2Fopenshift-helm-charts%2Fcharts%2Freleases%2Fdownload%2Fredhat-eap8-1.1.2%2Feap8-1.1.2.tgz
```
?
Should we use 
```

Comment (aneelac22) [reply redirectpage-tests/README.adoc:21]: Yes!

Comment (aneelac22) [reply redirectpage-tests/README.adoc:21]: Yes, that is correct

File: pkg/assets/static/redirectpage.js [hunk 4]
```diff
@@ -0,0 +1,189 @@
+const configURL = "/api/v1/authconfig"; // URL to get config
+const queryString = window.location.search;
+const urlParams = new URLSearchParams(queryString);
+const link = urlParams.get("link");
+const keyword = urlParams.get("keyword");
+const selectedId = urlParams.get("selectedId");
+const consoleUrlMock = "https://console.redhat.com/openshift/sandbox";
```
alexeykazakov, line 7: Instead of hardcoding this consleURL (and why it's called consoleUrl**Mock** btw?) let's load it from the config (`/api/v1/authconfig`)
So let's maybe do the following:
- rename `consoleUrlMock` to something like `signupURL`.
- try to load it from `configURL` JSON as `signup-url`.
- if there is no `signup-url` defined in the config payload JSON then fall back to the default value (which is just the base URL: `/`)
- We, the dev sandbox team, will add `https://console.redhat.com/openshift/sandbox` to the config API for our production sandbox.

pkg/assets/static/redirectpage.js
```diff
@@ -0,0 +1,189 @@
...
+  script.onerror = errorFunction;
+  script.onload = loadFunction;
+  script.onreadystatechange = loadFunction;
+  document.head.appendChild(script);
+}
+
+async function getRedirectData() {
+  const xhr = new XMLHttpRequest();
+
+  xhr.open("GET", registrationURL, true);
```
alexeykazakov, line 145: Hm... I don't actually see in my browser any requests to `"/api/v1/signup"`. This is what I see instead:
- redirectpage.html redirects to sso (expected)
- then it redirects to console.openshift.com/sandbox even if I already have an account in dev sandobx. So it doesn't seem to actually check `/api/v1/signup` unless I'm missing something. But anyway, no matter what I do I always end up in console.openshift.com

pkg/assets/static/redirectpage.js
```diff
@@ -0,0 +1,189 @@
...
+    if (xhr.readyState === XMLHttpRequest.DONE) {
+      if (xhr.status >= 200 && xhr.status < 300) {
+        try {
+          const data = JSON.parse(xhr.responseText);
+          handleSuccess(data);
+        } catch (error) {
+          handleError();
+        }
+      } else {
+        handleError();
```
alexeykazakov, line 158: If the signup api returns 404 it's not an error actually. It just means that user has to sign up first. So I wouldn't mix it with other cases when something went wrong.
And yes, in case of 404 we should redirect to the signup URL but we also need to redirect back to the redirectpage after user signs up successfully. See https://issues.redhat.com/browse/RHCLOUD-32115?focusedId=24554951&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-24554951 for details.
The redirect param is not currently supported in HCC but we can add this to our internal landing page for now and later add it to HCC too.

alexeykazakov (commented):
Still not working for me :(
See my comments below. Please let me know if you need any help with setting up dev environment (see [this comment in JIRA](https://issues.redhat.com/browse/RHCLOUD-32115?focusedId=24563919&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-24563919)) so you can properly test it localy.

pkg/assets/static/redirectpage.js
alexeykazakov: > We, the dev sandbox team, will add `https://console.redhat.com/openshift/sandbox` to the config API for our production sandbox.

See https://github.com/codeready-toolchain/registration-service/pull/464

alexeykazakov (changes requested):
See my comments above.

PR #417, opened 2024-03-22
add PublicViewer support

filariow:
linked to:
* https://github.com/codeready-toolchain/host-operator/pull/998
* https://github.com/codeready-toolchain/api/pull/410
* https://github.com/codeready-toolchain/toolchain-common/pull/377

PR #443, opened 2024-07-11
feat: add Public-Viewer support

filariow:
This PR adds support for Public-Viewer into `proxy` as described in JIRA [ASC-532](https://issues.redhat.com//browse/ASC-532)

Supersedes https://github.com/codeready-toolchain/registration-service/pull/417

Linked to: https://github.com/codeready-toolchain/toolchain-e2e/pull/1009

Splitting in multiple PRs:
- [x] #447
- [x] #448
- [x] #449 
- [x] #450 
- [x] #451
- [x] #452 
- [x] #455


alexeykazakov (commented):
It's a big PR. I only half way through it :)

pkg/proxy/handlers/spacelister_list.go
```diff
@@ -40,19 +42,38 @@ func ListUserWorkspaces(ctx echo.Context, spaceLister *SpaceLister) ([]toolchain
 	}
 	// signup is not ready
 	if signup == nil {
-		return []toolchainv1alpha1.Workspace{}, nil
+		return nil, nil
```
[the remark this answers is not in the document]
filariow: No big reason, I'm reverting this change

pkg/proxy/handlers/spacelister_list.go
```diff
@@ -40,19 +42,38 @@ func ListUserWorkspaces(ctx echo.Context, spaceLister *SpaceLister) ([]toolchain
...
+	}
+
+	// get MUR Names
+	murNames := getMURNamesForList(ctx, signup)
+	if len(murNames) == 0 {
+		return nil, nil
```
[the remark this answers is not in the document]
filariow: Updating it with empty slice return

pkg/proxy/handlers/spacelister_get.go
```diff
@@ -54,6 +58,56 @@ func GetUserWorkspace(ctx echo.Context, spaceLister *SpaceLister, workspaceName
...
+			return pvSb, nil
+		}
+		ctx.Logger().Error(fmt.Sprintf("unauthorized access - there is no SpaceBinding present for the user %s and the workspace %s", userSignup.CompliantUsername, workspaceName))
+	}
+
+	return userSpaceBinding, nil
+}
+
+// getUserSpaceBinding retrieves the user space binding for an user and a space.
+func getUserSpaceBinding(ctx echo.Context, spaceLister *SpaceLister, space *toolchainv1alpha1.Space, userSignup *signup.Signup) (*toolchainv1alpha1.SpaceBinding, error) {
```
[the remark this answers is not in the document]
filariow: ca4b882

pkg/proxy/handlers/spacelister_get.go
```diff
@@ -54,6 +58,56 @@ func GetUserWorkspace(ctx echo.Context, spaceLister *SpaceLister, workspaceName
...
+			}
+			return pvSb, nil
+		}
+		ctx.Logger().Error(fmt.Sprintf("unauthorized access - there is no SpaceBinding present for the user %s and the workspace %s", userSignup.CompliantUsername, workspaceName))
+	}
+
+	return userSpaceBinding, nil
+}
+
+// getUserSpaceBinding retrieves the user space binding for an user and a space.
```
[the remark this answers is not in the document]
filariow: ec8b470

pkg/proxy/handlers/spacelister_get.go
```diff
@@ -74,7 +128,6 @@ func GetUserWorkspace(ctx echo.Context, spaceLister *SpaceLister, workspaceName
 	}
 	if len(userSpaceBindings) == 0 {
 		//  let's only log the issue and consider this as not found
```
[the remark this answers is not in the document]
filariow: ec8b470

pkg/context/keys.go
```diff
@@ -25,4 +25,8 @@ const (
 	WorkspaceKey = "workspace"
 	// RequestReceivedTime is the context key for the starting time of a request made
 	RequestReceivedTime = "requestReceivedTime"
+	// PublicViewerEnabled is a boolean value indicating whether PublicViewer support is enabled
+	PublicViewerEnabled = "publicViewerEnabled"
+	// ImpersonateUser is the content key for the impersonated user in proxied call
```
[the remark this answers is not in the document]
filariow: ec8b470

pkg/proxy/handlers/spacelister_list.go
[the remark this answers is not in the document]
filariow: 6c2a595

pkg/proxy/handlers/spacelister_list.go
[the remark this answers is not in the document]
filariow: 6c2a595

pkg/proxy/handlers/spacelister_get_test.go
```diff
@@ -726,3 +735,257 @@ func TestGetUserWorkspace(t *testing.T) {
...
+	robinWS := workspaceFor(t, fakeClient, "robin", "admin", true)
+	batmanWS := workspaceFor(t, fakeClient, "batman", "admin", true)
+
+	tests := map[string]struct {
+		username             string
+		expectedErr          string
+		workspaceRequest     string
+		expectedWorkspace    *toolchainv1alpha1.Workspace
+		overrideInformerFunc func() service.InformerService
+		overrideSignupFunc   func(ctx *gin.Context, userID, username string, checkUserSignupComplete bool) (*signup.Signup, error)
```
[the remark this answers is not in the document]
filariow: great catch, fixed in 7f015b5

pkg/proxy/handlers/spacelister_get_test.go
```diff
@@ -726,3 +735,257 @@ func TestGetUserWorkspace(t *testing.T) {
...
+		fake.NewSpaceBinding("robin-1", "robin", "robin", "admin"),
+		fake.NewSpaceBinding("batman-1", "batman", "batman", "admin"),
+
+		fake.NewSpaceBinding("community-robin", toolchainv1alpha1.KubesawAuthenticatedUsername, "robin", "viewer"),
+	)
+
+	robinWS := workspaceFor(t, fakeClient, "robin", "admin", true)
+	batmanWS := workspaceFor(t, fakeClient, "batman", "admin", true)
+
+	tests := map[string]struct {
```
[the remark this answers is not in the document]
filariow: These tests here only applies to the case in which PublicViewer is enabled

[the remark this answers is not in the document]
alexeykazakov: Sure. That's the point. Why not covering both cases? It's a bit confusing why you have one test function `TestSpaceListerGet()` which tests both cases with enabled and disabled public viewer without actually testing the public viewer access. And then a separate test with public viewer enabled but this time which actually checks the public viewer access.

Any specific reason for not combining them in one test function with a single table which would test everything?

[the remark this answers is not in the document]
filariow: I tried multiple time merging those tests in a single table but I couldn't find a tidy solution. Either the table exploded or the test became unreadable to me. The point here is that we have a set of tests we expect to pass both when PublicViewer is on and off, and a set of tests that have a different behavior if PublicViewer is on.

So, either we duplicate the first set or -as implemented now- we do execute the common ones two times and have another function for the ones with a different behavior. Do you have any suggestions?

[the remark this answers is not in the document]
filariow: oh sorry, I got confused with another Test suite. Yeah, that makes sense. Let me try to merge them!

pkg/proxy/service/cluster_service.go
```diff
@@ -65,7 +65,49 @@ func (s *ServiceImpl) GetClusterAccess(userID, username, workspace, proxyPluginN
...
+	userSignup, err := s.Services().SignupService().GetSignupFromInformer(nil, userID, username, false)
+	if err != nil {
+		return "", err
+	}
+	// if signup has the CompliantUsername set it means that MUR was created and useraccount is provisioned
+	if userSignup == nil || userSignup.CompliantUsername == "" {
+		cause := errs.New("user is not provisioned (yet)")
+		log.Error(nil, cause, fmt.Sprintf("signup object: %+v", userSignup))
+		return "", cause
+	}
```
mfrancisc, line 90: I see this check is now duplicated at lines 99-108. Maybe we could extract this into a function ?

pkg/proxy/proxy.go
```diff
@@ -264,51 +266,221 @@ func (p *Proxy) health(ctx echo.Context) error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	// check whether the user's has access to the home workspace
+	// and whether the requestedNamespace -if any- exists in the workspace.
+	requestedNamespace := namespaceFromCtx(ctx)
+	if err := validateWorkspaceRequest(workspaceName, requestedNamespace, workspaces); err != nil {
+		return nil, crterrors.NewForbiddenError("invalid workspace request", err.Error())
+	}
```
mfrancisc, line 343: I'm not sure I following here - from the comments it seems that we are checking twice if requested namespace belongs to the workspace. But  `listUserWorkspaces` only returns the workspaces for a given user right ?

pkg/proxy/proxy.go
```diff
@@ -264,51 +266,221 @@ func (p *Proxy) health(ctx echo.Context) error {
...
+
+	return nil
+}
+
+// validateSpace checks whether the Space exists.
+func (p *Proxy) validateSpace(workspaceName string) error {
+	if _, err := p.app.InformerService().GetSpace(workspaceName); err != nil {
+		// log the actual error but do not return it so that it doesn't reveal information about a space that may not belong to the requestor
+		log.Error(nil, err, "unable to get target cluster for workspace "+workspaceName)
```
mfrancisc, line 365: shouldn't this be something like "the requested space is not available", as the error we return ? the check doesn't seem related to the target cluster.

pkg/proxy/proxy.go
```diff
@@ -264,51 +266,221 @@ func (p *Proxy) health(ctx echo.Context) error {
...
+
+	// retrieve the requesting user's UserSignup
+	signup, err := p.spaceLister.GetSignupFunc(nil, userID, username, false)
+	if err != nil {
+		return nil, crterrors.NewInternalError(errs.New("unable to get target cluster"), "user not found")
+	}
+
+	// if PublicViewer is enabled and user has no direct access to the workspace, proceed as PublicViewer
+	publicViewerEnabled := context.IsPublicViewerEnabled(ctx)
+	if publicViewerEnabled && !hasDirectAccess(signup, w) {
```
mfrancisc, line 418: maybe I'm making some confusion here - but shouldn't there be a binding for `KubesawAuthenticatedUsername` before giving public viewer access to this workspace ? It seems like we only check if the public viewer feature is enabled.

mfrancisc (commented):
Great job, from a first review looks good 👍 

I have few comments/questions and I'm still reviewing the tests.

pkg/proxy/handlers/spacelister_get.go
```diff
@@ -26,6 +28,8 @@ func HandleSpaceGetRequest(spaceLister *SpaceLister, GetMembersFunc cluster.GetM
 	// get specific workspace
 	return func(ctx echo.Context) error {
 		requestReceivedTime := ctx.Get(regsercontext.RequestReceivedTime).(time.Time)
+		publicViewerEnabled := configuration.GetRegistrationServiceConfig().PublicViewerEnabled()
+		ctx.Set(context.PublicViewerEnabled, publicViewerEnabled)
```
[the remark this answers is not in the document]
filariow: getting from configuration when needed could lead you to fetch two different values for the same request. This can happen if the configuration is updated in the middle of processing a request. Very unlikely, but still possible.

Also, having this in context allows us to easily log the value.

pkg/proxy/handlers/spacelister_get.go
```diff
@@ -53,10 +57,58 @@ func GetUserWorkspace(ctx echo.Context, spaceLister *SpaceLister, workspaceName
...
+	// retry with PublicViewer's signup
+	if userSpaceBinding == nil {
+		if context.IsPublicViewerEnabled(ctx) {
+			pvSb, err := getUserSpaceBinding(ctx, spaceLister, space, toolchainv1alpha1.KubesawAuthenticatedUsername)
+			if err != nil {
+				ctx.Logger().Error(fmt.Sprintf("error checking if SpaceBinding is present for user %s and the workspace %s", toolchainv1alpha1.KubesawAuthenticatedUsername, workspaceName))
+				return nil, err
+			}
+			if pvSb == nil {
+				ctx.Logger().Error(fmt.Sprintf("unauthorized access - there is no SpaceBinding present for the user %s and the workspace %s", userSignup.CompliantUsername, workspaceName))
```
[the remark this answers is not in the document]
filariow: great catch!

pkg/proxy/handlers/spacelister_get.go
```diff
@@ -106,9 +157,16 @@ func GetUserWorkspaceWithBindings(ctx echo.Context, spaceLister *SpaceLister, wo
 	// check if user has access to this workspace
 	userBinding := filterUserSpaceBinding(userSignup.CompliantUsername, allSpaceBindings)
 	if userBinding == nil {
-		//  let's only log the issue and consider this as not found
-		ctx.Logger().Error(fmt.Sprintf("unauthorized access - there is no SpaceBinding present for the user %s and the workspace %s", userSignup.CompliantUsername, workspaceName))
-		return nil, nil
+		// if PublicViewer is enabled, check if the Space is visibile to PublicViewer
+		if context.IsPublicViewerEnabled(ctx) && userSignup.CompliantUsername != toolchainv1alpha1.KubesawAuthenticatedUsername {
```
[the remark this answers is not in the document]
filariow: because `getUserSignupAndSpace` can return the `KubesawAuthenticated`'s UserSignup, in this case we do not repeat the check

pkg/proxy/proxy.go
filariow: that's indeed a nice to have to avoid flooding the target cluster! let me work on this

filariow: I double-checked, and the access for public viewer is already guaranteed from `SpaceLister.GetUserWorkspace`.
It returns information about the Workspace only if the user has direct access or if PublicViewer support is enabled and PublicViewer user has access to the Workspace. So here we can assume that if we have data about the workspace and user has no direct access than the PublicViewer user has direct access to it. Hence, adding this check is not required. 

At the same time it is very easy to add this check, it could increase readability and robustness of the code without impacting much the performances. Do you suggest adding it?

```diff
-	if publicViewerEnabled && !hasDirectAccess(signup, w) {
+	if publicViewerEnabled && !hasDirectAccess(signup, w) && hasDirectAccess(kubesawAuthenticatedSignup, w) {
```

pkg/proxy/service/cluster_service.go
filariow: good catch. done in f9b9b5a0fd1166c3f38d05f4ce80006866522624

pkg/proxy/proxy.go
mfrancisc: oh gotcha! yeah it wasn't clear at the first, I couldn't remember all the checks that were performed before this if statement , adding it explicitly might help indeed , if this doesn't affect performances. Thanks for checking this!

filariow: Yeah, it's easy to forget some of this check. Even I had to double check. Also, the fact that this PR is so big is not helping. I'll try to break it in multiple PRs

pkg/proxy/handlers/spacelister_get.go
filariow, line 32: Update this. This is now done in a dedicated middleware: https://github.com/codeready-toolchain/registration-service/pull/453

pkg/proxy/proxy.go
```diff
@@ -264,51 +266,214 @@ func (p *Proxy) health(ctx echo.Context) error {
...
+		return nil, crterrors.NewForbiddenError("invalid workspace request", fmt.Sprintf("access to workspace '%s' is forbidden", workspaceName))
+	}
+	// workspace was found means we can forward the request
+	return workspace, nil
 }
 
 func (p *Proxy) handleRequestAndRedirect(ctx echo.Context) error {
 	requestReceivedTime := ctx.Get(context.RequestReceivedTime).(time.Time)
+	publicViewerEnabled := configuration.GetRegistrationServiceConfig().PublicViewerEnabled()
+	ctx.Set(context.PublicViewerEnabled, publicViewerEnabled)
```
filariow, line 476: Update this. This is now done in a dedicated middleware: #453

pkg/proxy/proxy.go
```diff
@@ -264,51 +266,214 @@ func (p *Proxy) health(ctx echo.Context) error {
...
+	if _, err := p.app.InformerService().GetSpace(workspaceName); err != nil {
+		// log the actual error but do not return it so that it doesn't reveal information about a space that may not belong to the requestor
+		log.Error(nil, err, "unable to get target cluster for workspace "+workspaceName)
+		return fmt.Errorf("the requested space is not available")
+	}
+	return nil
+}
+
+// checkUserIsProvisioned checks whether the user is Approved, if they are not an error is returned.
+// If public-viewer is enabled, user validation is skipped.
```
mfrancisc, line 371: why skipping the validation if public-viewer is enabled ? I mean should we skip it if both following conditions are true:
- public viewer functionality is enabled
- the requested workspace is a community one

mfrancisc: Also I think this will fail the ban usersignup e2e tests, since it will not block requests from banned users if I'm not mistaken. I'm planning to try and add a unit test for it as well, as I see that is not covered here.

filariow: with public-viewer enabled we just need a valid SSO Token, we don't require the user to have Signed Up

pkg/proxy/proxy.go
filariow: fixed in 8d944fb19e69a25eb1e4a168fc56ae3cdcf10969

pkg/proxy/proxy.go
mfrancisc: right, but only if the request was meant for accessing a community workspace, and not for any workspace right ? But I'm sure I'm missing something.

pkg/proxy/proxy.go
```diff
@@ -264,47 +267,208 @@ func (p *Proxy) health(ctx echo.Context) error {
...
+func (p *Proxy) getClusterAccessAsUserOrPublicViewer(ctx echo.Context, userID, username, proxyPluginName string, workspace *toolchainv1alpha1.Workspace) (*access.ClusterAccess, error) {
+	// retrieve the requesting user's UserSignup
+	userSignup, err := p.spaceLister.GetSignupFunc(nil, userID, username, false)
+	if err != nil {
+		return nil, crterrors.NewInternalError(errs.New("unable to get target cluster"), "user not found")
+	}
+
+	// proceed as PublicViewer if needed
+	publicViewerEnabled := context.IsPublicViewerEnabled(ctx)
+	if publicViewerEnabled && !hasDirectAccess(userSignup, workspace) {
```
alexeykazakov, line 430: Let's maybe add a comment stating that we attempt to access the workspace as a public viewer only if the user doesn't have any role in the workspace. So we always access the workspace using the most powerful role available to the user: the one which is specified by the space binding. And only if there is none then we use the public viewer user. We also do not check if the public viewer actually has any access in the workspace. If there is no role for the public viewer then the request will be rejected by the member cluster.

pkg/proxy/proxy.go
filariow: Whether the user has access to this workspace or not is checked later in the chain. This method only checks that the user is provisioned and it only applies to the scenario in which Public-Viewer is disabled.

The user access check is performed by retrieving the requested workspace with SpaceBindings (cf. L332) and retrieving the Cluster Access (cf L345, L409, L420)

filariow: Regarding banned user, I think you may be correct, I'll double check. Great point!

pkg/proxy/proxy.go
```diff
@@ -264,47 +267,208 @@ func (p *Proxy) health(ctx echo.Context) error {
...
+	// UserSignup complete status is not checked, since it might cause the proxy blocking the request
+	// and returning an error when quick transitions from ready to provisioning are happening.
+	userSignup, err := p.app.SignupService().GetSignupFromInformer(nil, userID, username, false)
+	if err != nil {
+		return err
+	}
+
+	// if the UserSignup is nil or has NOT the CompliantUsername set,
+	// it means that MUR was NOT created and useraccount is NOT provisioned yet
+	// TODO(@filariow): should we also check if the user is Banned?
```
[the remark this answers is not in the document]
filariow: I really like the middleware idea! let me look into it

pkg/proxy/proxy.go
```diff
@@ -632,7 +812,10 @@ func replaceTokenInWebsocketRequest(req *http.Request, newToken string) {
 	req.Header.Set(ph, strings.Join(protocols, ","))
 }
 
-func validateWorkspaceRequest(requestedWorkspace, requestedNamespace string, workspaces []toolchainv1alpha1.Workspace) error {
+// validateWorkspaceRequest checks whether the requested workspace is in the list of workspaces the user has visibility on (retrieved via the spaceLister).
+// If `requestedWorkspace` is zero, this function looks for the home workspace (the one with `status.Type` set to `home`).
+// If `requestedNamespace` is NOT zero, this function checks if the namespace exists in the workspace.
```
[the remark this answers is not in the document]
filariow: IIUIC, if we remove this check we could enable unwanted cross-workspace.

Consider the scenario in which Alice has access to workspaces `alice` - with namespace `alice-tenant`- and `bob` - with namespace `bob-tenant`. Both living in the same cluster.
If we remove the check, Alice could explicitly target the workspace `alice` and access the namespace `bob-tenant`. Something like `.../workspaces/alice/api/v1/namespaces/bob-tenant/pods`. Am I correct?

pkg/proxy/proxy.go
```diff
@@ -264,47 +267,208 @@ func (p *Proxy) health(ctx echo.Context) error {
...
+	}
+
+	// retrieve the ClusterAccess for the cluster hosting the workspace and the given user.
+	return p.app.MemberClusterService().GetClusterAccess(userID, username, workspace.Name, proxyPluginName, publicViewerEnabled)
+}
+
+// hasDirectAccess checks if an UserSignup has access to a workspace.
+// Workspace's bindings are obtained from its `status.bindings` property.
+func hasDirectAccess(signup *signup.Signup, workspace *toolchainv1alpha1.Workspace) bool {
+	if signup == nil || workspace == nil {
```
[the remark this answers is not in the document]
filariow: removed in 75d7982e8db7110ffdb5430899bd1d02a7dcb341

pkg/proxy/proxy.go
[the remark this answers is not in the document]
alexeykazakov: > If we remove the check, Alice could explicitly target the workspace alice and access the namespace bob-tenant. Something like `.../workspaces/alice/api/v1/namespaces/bob-tenant/pods`. Am I correct?

Not really. Our proxy will redirect the requests indeed to the same cluster: `/api/v1/namespaces/bob-tenant/pods` as Alice. But since Alice does not have permissions to access that namespace, kube will reject it.

It basically will still fail but on the kube side. Not on the proxy side.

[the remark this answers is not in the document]
filariow: ok, but if Alice has access to `bob-tenant` namespace because Bob gave her access to his workspace, the request will not fail. Right?

If this is correct, Alice could set a workspace and perform cross-workspace requests by simply using a namespace from another workspace. The only requirements are:
1. workspaces are hosted on the same cluster
2. Alice has access to both workspaces

So Alice can have a kubeconfig with server url set to `https://cluster-domain/workspaces/alice-workspace` and target namespaces from different workspace. To me this is a problem, because as a user I expect my requests to be confined to the workspace I set in the url.

pkg/proxy/proxy.go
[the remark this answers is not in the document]
filariow: I opend a PR for this: https://github.com/codeready-toolchain/registration-service/pull/455

pkg/proxy/proxy.go
```diff
@@ -264,47 +267,208 @@ func (p *Proxy) health(ctx echo.Context) error {
...
+	}
+	return nil
+}
+
+// checkSpaceExists checks whether the Space exists.
+func (p *Proxy) checkSpaceExists(workspaceName string) error {
+	if _, err := p.app.InformerService().GetSpace(workspaceName); err != nil {
+		// log the actual error but do not return it so that it doesn't reveal information about a space that may not belong to the requestor
+		log.Errorf(nil, err, "requested space '%s' is not available", workspaceName)
+		return fmt.Errorf("the requested space is not available")
```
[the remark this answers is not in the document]
filariow: It makes sense. I committed the suggestion. Thanks!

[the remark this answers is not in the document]
filariow: This seems to have broken [unit tests](https://github.com/codeready-toolchain/registration-service/actions/runs/10601872194/job/29383835577?pr=443)  should I revert it or update unit tests accordingly?

pkg/proxy/proxy.go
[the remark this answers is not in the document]
alexeykazakov: This is correct but I still don't see a problem with that. It's not like Alice can get hack the proxy and get access to a namespace she doesn't have access to. She just could create a weird workspace/namespace context and use `/workspaces/alice/api/v1/namespaces/bob-tenant/pods` instead of more appropriate `/workspaces/bob/api/v1/namespaces/bob-tenant/pods`.

This is exactly why we introduce it in the first place. There was a bug in HAC/UI with the workspace switching. The workspace switcher would change the workspace but wouldn't change the namespace. So we hopped that setting such a limit would make it easy to spot bugs/mistakes on the client side (like in HAC): hey, you are using a weird context, make sure you really knows what namespace in what workspace you want to access otherwise you could be scratching your head wondering why it gives you weird result.
So it is not a security problem.

However it has significant side effects. We faced multiple cases when users try to access some namespaces which do not belong to any workspace at all! Some global view-only namespaces which should be available to all authenticated users. Some operators use such approach.
With this limitation in place it's impossible to access such non-workspace namespaces through the proxy even if they are totally accessible using the api server directly.

pkg/proxy/proxy.go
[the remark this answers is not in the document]
alexeykazakov: There is a glitch in github actions atm so I can't see the test logs. Did it break you own new tests or some existing tests?
My first thought is to adjust the test to the changed behavior (which we think is correct) but maybe I'm missing something.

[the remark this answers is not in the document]
filariow: New test indeed. I was convinced it was an already existing one and I didn't want to break backwards compatibility. Sorry for the confusion. Fixed in 30b980bd03aefad976de496610a62258fc3d75c5

pkg/proxy/proxy.go
```diff
@@ -264,47 +267,208 @@ func (p *Proxy) health(ctx echo.Context) error {
...
+
+// getClusterAccessAsUserOrPublicViewer if the requesting user has direct access to the workspace,
+// it returns the ClusterAccess impersonating the requesting user.
+// Otherwise, if PublicViewer support is enabled and PublicViewer user has access to the workspace,
+// it returns the ClusterAccess impersonating the PublicViewer user.
+func (p *Proxy) getClusterAccessAsUserOrPublicViewer(ctx echo.Context, userID, username, proxyPluginName string, workspace *toolchainv1alpha1.Workspace) (*access.ClusterAccess, error) {
+	// retrieve the requesting user's UserSignup
+	userSignup, err := p.spaceLister.GetSignupFunc(nil, userID, username, false)
+	if err != nil {
+		return nil, crterrors.NewInternalError(errs.New("unable to get target cluster"), "user not found")
```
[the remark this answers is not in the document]
filariow: updated in 96b16461e85e6a6d7c89800812167c2bf6ab6f29 and 048f29ba1bc98e564f62713f6bc277a8d2e90e09

pkg/proxy/proxy.go
filariow: done in 048f29ba1bc98e564f62713f6bc277a8d2e90e09

filariow: Also, I think we should check if public-viewer has access before forwarding the request. So this is a gap I need to fill.

filariow: done in 6408242518acad1da5c980985948e8af5dd8b25f

alexeykazakov: > Also, I think we should check if public-viewer has access before forwarding the request. So this is a gap I need to fill.

It would work without that check too. We would forward the request but the request would be rejected by the api server. But this check will reject the request faster (so it will consume less resources). So, yeah, we can have this check here.

pkg/proxy/proxy_community_test.go
```diff
@@ -0,0 +1,306 @@
...
+			},
+			// Given smith2 owns a workspace named communityspace
+			// And   communityspace is publicly visible (shared with PublicViewer)
+			// And   a user named communityuser exists
+			// And   smith2's communityspace is not directly shared with communityuser
+			// When  communityuser requests the list of pods in workspace communityspace
+			// Then  the request is forwarded from the proxy
+			// And   the request impersonates the PublicViewer
+			// And   the request's X-SSO-User Header is set to communityuser's ID
+			// And   the request is successful
```
alexeykazakov, line 141: The ` communityuser` name is a bit confusing. It's just a regular user, right? There is nothing special in this user.
I would maybe use the following setup:
- User `smith` with publicly available workspace `smith-communityspace`
- User `alice` with private workspace `alice-private`
- User `bob` with private workspace `bob-private`
- User `john-not-ready` (not-ready)

Then check that:
- `smith`  and `john-notready` can access `smith-communityspace` only.
- `alice` can access `alice-private` and `smith-communityspace`
- `bob` can access `bob-private` and `smith-communityspace`
- some non-existing user can access `smith-communityspace` only

That would cover all use cases.
And I would also add a test for banned user after the banned user middleware PR is merged. We need that PR get merged first anyway.

pkg/proxy/proxy_community_test.go
```diff
@@ -0,0 +1,306 @@
...
+				ExpectedProxyResponseStatus: http.StatusOK,
+				RequestPath:                 fmt.Sprintf("http://localhost:%s/workspaces/communityspace/api/communityspace/pods", port),
+				ExpectedResponse:            httpTestServerResponse,
+			},
+			// Given A not ready user exists
+			// When  the not ready user requests the list of pods in workspace communityspace
+			// Then  the request is forwarded from the proxy
+			// And   the request impersonates the not ready user
+			// And   the request's X-SSO-User Header is set to not ready user's ID
+			// And   the request is successful
```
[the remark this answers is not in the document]
filariow: great catch, fixed in a2bc3f8

pkg/proxy/proxy_community_test.go
filariow: Thanks for the suggestion, I cleaned it up a bit and used the following users:
- User `smith` with publicly available workspace `smith-community`
- User `alice` with private workspace `alice-private`
- User `bob` as the non-signed-up user
- User `john` as the not-ready user

Now we check:
- `smith` can access `smith-community` as `smith`
- `john` can access `smith-community` as `kubesaw-authenticated`
- `bob` can access `smith-community` as `kubesaw-authenticated`
- `alice` can access `smith-community` as `kubesaw-authenticated`
- `smith` can not access `alice-private`

I'll add the test case for banned user

filariow: added in ea515a6: `eve` is the banned user

pkg/proxy/proxy_community_test.go
```diff
@@ -0,0 +1,374 @@
...
+			// Given banned user eve exists
+			// And   user smith exists
+			// And   smith owns a workspace named smith-community
+			// And   smith-community is publicly visible (shared with PublicViewer)
+			// When  eve requests the list of pods in smith's workspace
+			// Then  the proxy does NOT forward the request
+			// And   the proxy rejects the call with 403 Forbidden
+			"plain http actual request as banned user to community workspace": {
+				ProxyRequestMethod:          "GET",
+				ProxyRequestHeaders:         map[string][]string{"Authorization": {"Bearer " + s.token(eve, authsupport.WithEmailClaim(eveEmail))}},
```
mfrancisc, line 207: probably this is a scenario more suited for the proxy test but could it be that the email claim is not set for any reason, what will/should happen in that case ? the proxy should return an error since it's a mandatory claim I would say.

pkg/proxy/proxy_community_test.go
```diff
@@ -0,0 +1,374 @@
...
+
+				// given
+				fakeApp.Err = nil
+
+				testServer.Config.Handler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+					w.Header().Set("Content-Type", "application/json")
+					// Set the Access-Control-Allow-Origin header to make sure it's overridden by the proxy response modifier
+					w.Header().Set("Access-Control-Allow-Origin", "dummy")
+					w.WriteHeader(http.StatusOK)
+					_, err := w.Write([]byte("my response"))
```
mfrancisc, line 238: not sure I understand why do we need to set the response here , don't we already do this at lines 80-84 ? But maybe it's a different handler/server 😕

pkg/proxy/proxy_community_test.go
```diff
@@ -0,0 +1,374 @@
...
+					fake.Signup(eve.String(), &signup.Signup{
+						Name:              "eve",
+						CompliantUsername: "eve",
+						Username:          "eve@",
+						Status: signup.Status{
+							Ready:  false,
+							Reason: toolchainv1alpha1.UserSignupUserBannedReason,
+						},
+					}),
+				)
```
mfrancisc, line 285: minor - if I'm not mistaken looks like we can move this outside the `tests` loop and initialize it only once, not once per test case.

pkg/proxy/proxy_community_test.go
```diff
@@ -0,0 +1,374 @@
...
+					case "smith-community":
+						return fake.NewSpace("smith-community", "member-2", "smith"), nil
+					case "alice-private":
+						return fake.NewSpace("alice-private", "member-2", "alice"), nil
+					}
+					return nil, fmt.Errorf("space not found error")
+				}
+
+				sbSmithCommunitySmith := fake.NewSpaceBinding("smith-community-smith", "smith", "smith-community", "admin")
+				commSpacePublicViewer := fake.NewSpaceBinding("smith-community-publicviewer", toolchainv1alpha1.KubesawAuthenticatedUsername, "smith-community", "viewer")
```
mfrancisc, line 299: maybe this part also can be moved outside the test cases loop and the section after where you configure the informer listers and getters function if I'm not mistaken.

pkg/proxy/proxy.go
```diff
@@ -269,47 +272,220 @@ func (p *Proxy) health(ctx echo.Context) error {
...
+	// UserSignup complete status is not checked, since it might cause the proxy blocking the request
+	// and returning an error when quick transitions from ready to provisioning are happening.
+	userSignup, err := p.app.SignupService().GetSignupFromInformer(nil, userID, username, false)
+	if err != nil {
+		return err
+	}
+
+	// if the UserSignup is nil or has NOT the CompliantUsername set,
+	// it means that MUR was NOT created and useraccount is NOT provisioned yet
+	// TODO(@filariow): should we also check if the user is Banned?
```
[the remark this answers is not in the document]
filariow: done in 8306e888cafef250917a363f7d2c14ab525996c6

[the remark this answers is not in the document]
filariow: done in b18db564b0b303a54d4c380858576c1f5e59ef92

pkg/proxy/proxy.go
```diff
@@ -269,47 +272,220 @@ func (p *Proxy) health(ctx echo.Context) error {
...
+
+// getClusterAccessAsUserOrPublicViewer if the requesting user exists and has direct access to the workspace,
+// this function returns the ClusterAccess impersonating the requesting user.
+// If PublicViewer support is enabled and PublicViewer user has access to the workspace,
+// this function returns the ClusterAccess impersonating the PublicViewer user.
+// If requesting user does not exists and PublicViewer is disabled or does not have access to the workspace,
+// this function returns an error.
+func (p *Proxy) getClusterAccessAsUserOrPublicViewer(ctx echo.Context, userID, username, proxyPluginName string, workspace *toolchainv1alpha1.Workspace) (*access.ClusterAccess, error) {
+	// retrieve the requesting user's UserSignup
+	userSignup, err := p.spaceLister.GetSignupFunc(nil, userID, username, false)
```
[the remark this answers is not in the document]
filariow: hmmm, good question. Isn't it the same? (cf. https://github.com/codeready-toolchain/registration-service/blob/92998a6271ae9281a9efc15c7e8aa40021f53466/pkg/proxy/handlers/spacelister.go#L37)

pkg/proxy/proxy_community_test.go
filariow: in this case the `addUserContext` middleware will fail
https://github.com/codeready-toolchain/registration-service/blob/92998a6271ae9281a9efc15c7e8aa40021f53466/pkg/proxy/proxy.go#L110

because of this check:
https://github.com/codeready-toolchain/registration-service/blob/92998a6271ae9281a9efc15c7e8aa40021f53466/pkg/auth/tokenparser.go#L80

pkg/proxy/proxy_community_test.go
filariow: great catch, fixed in d5c15036812a2c789cbb7e75ac22237cdf9a7159

pkg/proxy/proxy_community_test.go
filariow: good idea, done in 94e45a03b5b980db97572c7538eca8d1efd41880

pkg/proxy/proxy_community_test.go
filariow: good idea, done in 94e45a03b5b980db97572c7538eca8d1efd41880

alexeykazakov (approved):
Looks good! Thank you for addressing the comments!
However there is a few lines not covered by the tests. Check out the codecov report.
Could you please take a look and see if you can add the missing tests if it's not too tricky?

filariow:
> Looks good! Thank you for addressing the comments! However there is a few lines not covered by the tests. Check out the codecov report. Could you please take a look and see if you can add the missing tests if it's not too tricky?

I've worked on some of them. I could not find an easy fix for the others. AFAIUI, they are happening because the same check was already performed by another function in the chain, like retrieving two times the UserSignup or Space info. I tried refactoring but it was requiring too many changes. Maybe we can refactor later in another PR.  WDYT?

alexeykazakov:
> Maybe we can refactor later in another PR. WDYT?

Yes, sure. As I said let's do it if it's not too tricky. Otherwise it's OK to leave it as is.

pkg/proxy/proxy.go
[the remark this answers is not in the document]
mfrancisc: Correct! Thanks for replacing it.

pkg/proxy/proxy_community_test.go
mfrancisc: Right, I see that it throws an error when these claims are not being set, but I don't see a test checking this unless I'm missing something. If it's missing should we add a test case in the `proxy_community_test.go` suite which checks that if the `claims.Email` and `claims.PreferredUsername` are not set an error is returned ?

pkg/proxy/proxy.go
```diff
@@ -503,11 +682,16 @@ func extractUserToken(req *http.Request) (string, error) {
 func (p *Proxy) newReverseProxy(ctx echo.Context, target *access.ClusterAccess, isPlugin bool) *httputil.ReverseProxy {
 	req := ctx.Request()
 	targetQuery := target.APIURL().RawQuery
+	username, _ := ctx.Get(context.UsernameKey).(string)
+	ctx.Set(context.ImpersonateUser, target.Username())
```
mfrancisc, line 686: minor - should we add a comment explaining why we need this ? I guess it's just for logging purposes right ?

pkg/proxy/proxy_community_test.go
filariow: good idea, let me work on this

filariow: As `proxy_test.go` already contains a similar check for the JWT's `subject` field, I added a test case there: 150dbd02baf7e935b66fd56ccb22db1663b41d35

pkg/proxy/proxy.go
filariow: good idea, done in 599e2d2424b3fb2f5752ad57e7aef0496fe61783

mfrancisc (approved):
Great Job 🚀 🏅 

Thanks for the additional improvements/tests and for addressing my comments.

mfrancisc (approved):

alexeykazakov (approved):

mfrancisc (approved):

pkg/proxy/proxy.go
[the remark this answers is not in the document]
alexeykazakov: Here is the followup PR: https://github.com/codeready-toolchain/registration-service/pull/471

PR #444, opened 2024-07-12
fix: use UTC time in tests

filariow:


rajivnathan (commented):
Was there something that prompted this change?

filariow:
> Was there something that prompted this change?

I was not able to understand the reason why, but these tests started failing on my local with both `go 1.20.8` and `1.22.2`.

rajivnathan:
>I was not able to understand the reason why, but these tests started failing on my local with both go 1.20.8 and 1.22.2.

That's interesting, I just tried without your changes and it's failing for me too. I verified your change resolves it but it would be good to understand why. It looks like it could be related to https://github.com/codeready-toolchain/registration-service/pull/432 cc @sbryzak

filariow:
I think the main reason why we didn't notice this before merging #432 is that the CI instances are running with timezone set to UTC.

rajivnathan (approved):
Ok looks like this was probably just missed from the original PR. I think the fix is good 👍

PR #446, opened 2024-07-22
proposal: simplify labels.Requirement creation

filariow:
We can skip validations performed by [labels.NewRequirement](https://pkg.go.dev/k8s.io/apimachinery/pkg/labels#NewRequirement) as operator is fixed, key is constant, and value is extracted from Custom Resources.

> NewRequirement is the constructor for a Requirement. If any of these rules is violated, an error is returned:
>
> 1. The operator can only be In, NotIn, Equals, DoubleEquals, Gt, Lt, NotEquals, Exists, or DoesNotExist.
> 1. If the operator is In or NotIn, the values set must be non-empty.
> 1. If the operator is Equals, DoubleEquals, or NotEquals, the values set must contain one value.
> 1. If the operator is Exists or DoesNotExist, the value set must be empty.
> 1. If the operator is Gt or Lt, the values set must contain only one value, which will be interpreted as an integer.
> 1. The key is invalid due to its length, or sequence of characters. See validateLabelKey for more details.
>
> The empty string is a valid value in the input values set. Returned error, if not nil, is guaranteed to be an aggregated field.ErrorList 

[SelectorFromSet](https://pkg.go.dev/k8s.io/apimachinery/pkg/labels#SelectorFromSet) does not perform any validation, which means the server will reject the request if the Set contains invalid values.

### DIFF
--- a/pkg/proxy/handlers/spacelister_get.go
+++ b/pkg/proxy/handlers/spacelister_get.go
@@ -19,7 +19,6 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/selection"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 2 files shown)


alexeykazakov (approved):

MatousJobanek (approved):
/lgtm

PR #454, opened 2024-08-15
Set target cluster for social events

alexeykazakov:
Set the target cluster for UserSignups if it's associated with a social event with the target cluster field set.

Part of https://issues.redhat.com/browse/SANDBOX-724

Paired with https://github.com/codeready-toolchain/toolchain-e2e/pull/1035


rajivnathan (approved):
Nice! 🚀

PR #455, opened 2024-08-28
proposal: add banneduser middleware

filariow:
proposal for the BannedUser middleware as discussed [in #443 review|https://github.com/codeready-toolchain/registration-service/pull/443#discussion_r1731999440]

### DIFF
--- a/deploy/registration-service.yaml
+++ b/deploy/registration-service.yaml
@@ -47,6 +47,13 @@ objects:
           - toolchain.dev.openshift.com
         resources:
           - bannedusers
+        verbs:
+          - watch
+          - get
+          - list
+      - apiGroups:
+          - toolchain.dev.openshift.com
+        resources:
           - toolchainclusters
         verbs:
           - get
--- a/pkg/informers/service/informer_service.go
+++ b/pkg/informers/service/informer_service.go
@@ -9,10 +9,12 @@ import (
 	"github.com/codeready-toolchain/registration-service/pkg/informers"
 	"github.com/codeready-toolchain/registration-service/pkg/kubeclient/resources"
 	"github.com/codeready-toolchain/registration-service/pkg/log"
+	"github.com/codeready-toolchain/toolchain-common/pkg/hash"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/selection"
 )
 
 type Option func(f *ServiceImpl)
@@ -144,4 +146,34 @@ func (s *ServiceImpl) GetNSTemplateTier(name string) (*toolchainv1alpha1.NSTempl
 		return nil, err
 	}
 	return tier, err
+}
+
+func (s *ServiceImpl) BannedUsersByEmail(email string) ([]toolchainv1alpha1.BannedUser, error) {
+	hashedEmail := hash.EncodeString(email)
+	r, err := labels.NewRequirement(toolchainv1alpha1.BannedUserEmailHashLabelKey, selection.Equals, []string{hashedEmail})
+	if err != nil {
+		return nil, err
+	}
+
+	ls := labels.NewSelector().Add(*r)
+	objs, err := s.informer.BannedUsers.ByNamespace(configuration.Namespace()).List(ls)
+	if err != nil {
+		return nil, err
+	}
+
+	bb := []toolchainv1alpha1.BannedUser{}
+	for _, obj := range objs {
+		unobj := obj.(*unstructured.Unstructured)
+		bu := &toolchainv1alpha1.BannedUser{}
+		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(unobj.UnstructuredContent(), bu); err != nil {
+			log.Errorf(nil, err, "failed to list BannedUsers")
+			return nil, err
+		}
+
+		// check in case of hash collision
+		if bu.Spec.Email == email {
+			bb = append(bb, *bu)
+		}
+	}
+	return bb, nil
 }
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -4,6 +4,7 @@ import (
 	gocontext "context"
 	"crypto/tls"
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -137,7 +138,7 @@ func (p *Proxy) StartProxy(port string) *http.Server {
 	)
 
 	// routes
-	wg := router.Group("/apis/toolchain.dev.openshift.com/v1alpha1/workspaces")
+	wg := router.Group("/apis/toolchain.dev.openshift.com/v1alpha1/workspaces", p.ensureUserIsNotBanned())
 	// Space lister routes
 	wg.GET("/:workspace", handlers.HandleSpaceGetRequest(p.spaceLister, p.getMembersFunc))
 	wg.GET("", handlers.HandleSpaceListRequest(p.spaceLister))
@@ -157,7 +158,7 @@ func (p *Proxy) StartProxy(port string) *http.Server {
 	router.Any(fmt.Sprintf("%s*", openidAuthEndpoint()), p.openidAuth) // <- this is the step 5 in the flow above
 	router.Any(fmt.Sprintf("%s*", authEndpoint), p.auth)               // <- this is the step 7.
 	// The main proxy route
-	router.Any("/*", p.handleRequestAndRedirect)
+	router.Any("/*", p.handleRequestAndRedirect, p.ensureUserIsNotBanned())
 
 	// Insert the CORS preflight middleware
 	handler := corsPreflightHandler(router)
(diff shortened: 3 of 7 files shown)


pkg/informers/service/informer_service.go
```diff
@@ -145,3 +147,33 @@ func (s *ServiceImpl) GetNSTemplateTier(name string) (*toolchainv1alpha1.NSTempl
 	}
 	return tier, err
 }
+
+func (s *ServiceImpl) BannedUsersByEmail(email string) ([]toolchainv1alpha1.BannedUser, error) {
```
alexeykazakov, line 151: Maybe we want to name it `ListBannedUsersByEmail` just for consistency with other functions but I'm fine with keeping it as is.
```suggestion
func (s *ServiceImpl) ListBannedUsersByEmail(email string) ([]toolchainv1alpha1.BannedUser, error) {
```

pkg/informers/service/informer_service.go
```diff
@@ -145,3 +147,33 @@ func (s *ServiceImpl) GetNSTemplateTier(name string) (*toolchainv1alpha1.NSTempl
...
+			return nil, err
+		}
+
+		// check in case of hash collision
+		if bu.Spec.Email == email {
+			bb = append(bb, *bu)
+		}
+	}
+	return bb, nil
+}
```
alexeykazakov, line 179: Can you please add tests for this new function to `informer_service_test.go`?

pkg/proxy/proxy.go
```diff
@@ -137,7 +138,7 @@ func (p *Proxy) StartProxy(port string) *http.Server {
 	)
 
 	// routes
-	wg := router.Group("/apis/toolchain.dev.openshift.com/v1alpha1/workspaces")
+	wg := router.Group("/apis/toolchain.dev.openshift.com/v1alpha1/workspaces", p.ensureUserIsNotBanned())
```
alexeykazakov, line 141: Instead of adding this middleware to specific routes you can add it to the list of our middleware for all routes:
https://github.com/codeready-toolchain/registration-service/blob/5faca3af920bc27bb3acee382fce5662e1463321/pkg/proxy/proxy.go#L105-L121

And to avoid the situation when it's enforced for unsecured routes (which does not require authentication) you can add the following check: https://github.com/codeready-toolchain/registration-service/blob/5faca3af920bc27bb3acee382fce5662e1463321/pkg/proxy/proxy.go#L390-L392

In this case any new secured route we add in the future it will be covered automatically covered by this middlewere.

pkg/proxy/proxy.go
```diff
@@ -391,18 +392,52 @@ func (p *Proxy) addUserContext() echo.MiddlewareFunc {
...
 		}
 	}
 }
 
+// ensureUserIsNotBanned rejects the request if the user is banned
+func (p *Proxy) ensureUserIsNotBanned() echo.MiddlewareFunc {
+	return func(next echo.HandlerFunc) echo.HandlerFunc {
+		return func(ctx echo.Context) error {
+			errorResponse := func(err *crterrors.Error) error {
+				ctx.Logger().Error(errs.Wrap(err, "workspace list error"))
```
alexeykazakov, line 413:
```suggestion
				ctx.Logger().Error(errs.Wrap(err, "user ban status check error"))
```

pkg/proxy/proxy.go
```diff
@@ -391,18 +392,52 @@ func (p *Proxy) addUserContext() echo.MiddlewareFunc {
...
+				ctx.Response().Writer.Header().Set("Content-Type", "application/json")
+				ctx.Response().Writer.WriteHeader(int(err.Code))
+				return json.NewEncoder(ctx.Response().Writer).Encode(err.Status)
+			}
+
+			email := ctx.Get(context.EmailKey).(string)
```
alexeykazakov, line 419: Maybe worth adding a comment that this middleware is using the context set by the `addUserContext` middleware. So we need to make sure the right order of the middlewares.

pkg/proxy/proxy.go
```diff
@@ -391,18 +392,52 @@ func (p *Proxy) addUserContext() echo.MiddlewareFunc {
...
+			if email == "" {
+				return errorResponse(crterrors.NewUnauthorizedError("unauthenticated request", "anonymous access is not allowed"))
```
alexeykazakov, line 421: I wonder why you actually sets the response here (headers, response code, etc) instead of returning `crterrors.NewUnauthorizedError("unauthenticated request", "anonymous access is not allowed")` here as is as we do in other middlewares like `addUserContext`? I didn't check the code but apparently it works for `addUserContext`?

pkg/proxy/proxy.go
```diff
@@ -391,18 +392,52 @@ func (p *Proxy) addUserContext() echo.MiddlewareFunc {
...
+
+			// retrieve banned users
+			uu, err := p.app.InformerService().BannedUsersByEmail(email)
+			if err != nil {
+				return errorResponse(crterrors.NewInternalError(errs.New("unable to retrieve user"), "could not define ban status"))
+			}
+
+			// if a matching Banned user is found, then user is banned
+			if len(uu) > 0 {
+				return errorResponse(crterrors.NewForbiddenError("user is banned", "user is banned"))
```
alexeykazakov, line 432: We usually don't indicate that user is banned and return a generic forbidden error message for banned users. Just want to give as little details to banned users as possible. Not sure if it's right approach or not though. WDYT?
But it's not a big deal anyway.

pkg/proxy/proxy.go
```diff
@@ -391,18 +392,52 @@ func (p *Proxy) addUserContext() echo.MiddlewareFunc {
...
+
+			email := ctx.Get(context.EmailKey).(string)
+			if email == "" {
+				return errorResponse(crterrors.NewUnauthorizedError("unauthenticated request", "anonymous access is not allowed"))
+			}
+
+			// retrieve banned users
+			uu, err := p.app.InformerService().BannedUsersByEmail(email)
+			if err != nil {
+				return errorResponse(crterrors.NewInternalError(errs.New("unable to retrieve user"), "could not define ban status"))
```
alexeykazakov, line 427: Unless I'm missing something you do not log the root `err` anyway. So this information is lost.

pkg/proxy/proxy.go
```diff
@@ -391,18 +392,52 @@ func (p *Proxy) addUserContext() echo.MiddlewareFunc {
...
+			// if a matching Banned user is found, then user is banned
+			if len(uu) > 0 {
+				return errorResponse(crterrors.NewForbiddenError("user is banned", "user is banned"))
+			}
+
+			// user is not banned
+			return next(ctx)
+		}
+	}
+}
```
alexeykazakov, line 439: We could probably add some unit tests for `ensureUserIsNotBanned()` if it's not too tricky, right? Unless you believe it's already indirectly covered somehow...

alexeykazakov (commented):
Looks good overall. I left a few comments though.

pkg/informers/service/informer_service.go
filariow: done in b3a2d59a659a23d1f72253747637ff28104e56d1

pkg/proxy/proxy.go
filariow: done in 73039cc4deb732a28efae65e3fe22d677c9494a2

pkg/proxy/proxy.go
filariow: I think it's good for now to hide details. The worst think that can happen is that banned users cannot tell they are banned so they could start opening tickets complaining about the service not working properly.

I don't see this becoming a problem in the short term and given that this is the approach used up to now, I'd say to hide details also in this PR. If we start to have tickets for this behavior or the team decides to change the approach, we can create a new story.

fixed in 6c12ec8d0b69a7c6424897acb0be6a8193f9da6a

pkg/proxy/proxy.go
filariow: good idea! Added in function comment: aa1697e219ebe276116b0840b824bb731a87c1de

pkg/proxy/proxy.go
filariow: added in 2061e887edd1e3b9276bf305fa06022b9f52a8fd

pkg/informers/service/informer_service.go
filariow: added in e2625007b558f2d2fa131df46e0f9ffa5ef238a0

pkg/proxy/proxy.go
filariow: done in b822ccf2581c0429e5fb37644e566577702f138b

pkg/proxy/proxy_test.go
```diff
@@ -1168,7 +1245,7 @@ func (s *TestProxySuite) token(userID uuid.UUID, extraClaims ...authsupport.Extr
 		Username: "username-" + userID.String(),
 	}
 
-	extra := append(extraClaims, authsupport.WithEmailClaim("someemail@comp.com"))
+	extra := append([]authsupport.ExtraClaim{authsupport.WithEmailClaim("someemail@comp.com")}, extraClaims...)
```
filariow, line 1248: for context: this change allows us to override the default email set by this auxiliary function. As a result if you don't define an email, you'll have `someemail@comp.com`. Otherwise you'll have the one you are explicitly setting

alexeykazakov (commented):
Runtime code looks good! Thank you for addressing my comments. I'm going to review the test code next.

alexeykazakov (approved):
Looks good! Just one question.
And how are we doing in terms of e2e-tests? We already have some tests which check banned users in proxy, right? Let's make sure that we add some e2e-tests for banned users for the public-viewer proxy PR.

pkg/proxy/proxy_test.go
alexeykazakov: Worth moving this comment to the code.

pkg/proxy/proxy_test.go
```diff
@@ -219,6 +255,8 @@ func (s *TestProxySuite) checkPlainHTTPErrors(fakeApp *fake.ProxyFakeApp) {
 
 		s.Run("internal error if accessing incorrect url", func() {
 			// given
+			fakeApp.Err = errors.New("some-error")
+			defer func() { fakeApp.Err = nil }()
```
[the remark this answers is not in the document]
filariow: We need it because `fakeApp` is shared across multiple test cases. The test cases we are adding after this one are impacted by its changes to `fakeApp`, so we have to cleanup before running the next one.
Without this line the tests we are adding in this PR inherit the change to `fakeApp.Err` and, as a result, `fakeApp` will return the error `some-error` even in the new tests. 

I added this simple cleanup so that the default `fakeApp` behavior is restored to default. Basically, now the test `internal error if get accesses returns an error` undoes the changes it has done to the `fakeApp` before exiting, leaving `fakeApp` in the default state.

pkg/proxy/proxy_test.go
filariow: done in d81542b35ef008098f4c3b09b4fbf1b9aeab453e

filariow:
> Looks good! Just one question. And how are we doing in terms of e2e-tests? We already have some tests which check banned users in proxy, right? Let's make sure that we add some e2e-tests for banned users for the public-viewer proxy PR.

I've opened a PR for improving banned users test cases: https://github.com/codeready-toolchain/toolchain-e2e/pull/1045

deploy/registration-service.yaml
```diff
@@ -47,6 +47,13 @@ objects:
           - toolchain.dev.openshift.com
         resources:
           - bannedusers
+        verbs:
```
mfrancisc, line 50: minor thing - but since you decoupled the update of the deployment manifest in a separate PR in host-operator I would create a dedicated PR for this so that it can be merged together with the host one. WDYT?

mfrancisc: never mind - the manifest was dropped from this repo in this PR https://github.com/codeready-toolchain/registration-service/pull/459

mfrancisc (approved):
Great Job 🚀

rsoaresd (commented):
Great job 🚀 

Just a minor question:

pkg/informers/service/informer_service.go
```diff
@@ -145,3 +147,33 @@ func (s *ServiceImpl) GetNSTemplateTier(name string) (*toolchainv1alpha1.NSTempl
...
+
+	ls := labels.NewSelector().Add(*r)
+	objs, err := s.informer.BannedUsers.ByNamespace(configuration.Namespace()).List(ls)
+	if err != nil {
+		return nil, err
+	}
+
+	bb := []toolchainv1alpha1.BannedUser{}
+	for _, obj := range objs {
+		unobj := obj.(*unstructured.Unstructured)
```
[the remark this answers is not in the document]
filariow: I'm not sure this can happen. If I'm not mistaken it's stored internally as `*unstructured.Unstructured` (cf. https://github.com/kubernetes/client-go/blob/2666bd29867c96168c5e9429fd74fd9bfbedac8b/dynamic/dynamicinformer/informer.go#L138). Also I can not immagine how to write a test for covering this check.

The other functions in this file are using an unchecked casting like this too. I'd say that -if we find out it to be a problem- we can create a PR for fixing this in all the functions. WDYT?

PR #459, opened 2024-09-10
drop registration-service.yaml file

MatousJobanek:
There is no point in keeping the file in registration-service repository (it's not used for anything) and keep it always up to date with what is in host-operator repo. 
The template that is in the host-operator is the only source of truth.


mfrancisc (approved):

alexeykazakov (approved):

fbm3307 (approved):
Thanks for the cleanup

PR #460, opened 2024-09-10
Init cached client and use it for ToolchainConfig loading

MatousJobanek:
* introduce cached client provided by controller-runtime library and backed by the shared informers using cache for host-operator namespace only.
* the client is used to load the latest and the greatest version of ToolchainConfig CR and associated sercrets
* dropped the goroutine that updated the ToolchainConfig in a custom cache "manually"

The cached client will be used as a foundation for the next refactoring changes (dropping custom shared informers and custom http client wrapper)

some comments will be added also inline

depends on https://github.com/codeready-toolchain/host-operator/pull/1084

[KUBESAW-188](https://issues.redhat.com/browse/KUBESAW-188)


cmd/main.go
```diff
@@ -63,16 +62,16 @@ func main() {
 		os.Exit(1)
 	}
 
-	// create runtime client
-	cl, err := configClient(cfg)
+	ctx := controllerruntime.SetupSignalHandler()
```
MatousJobanek, line 65: I used the provided function from the controller-runtime library to handle the setup signal as it was almost the same as the logic that was originally here. This however also returns the cancel context.
I could have mimic the logic and make it more closer to the original version, but there is no point/benefit in doing that.

pkg/proxy/proxy.go
```diff
@@ -175,7 +175,11 @@ func (p *Proxy) StartProxy(port string) *http.Server {
 	// listen concurrently to allow for graceful shutdown
 	go func() {
 		if err := srv.ListenAndServe(); err != nil {
-			log.Error(nil, err, err.Error())
+			if errors.Is(err, http.ErrServerClosed) {
+				log.Info(nil, fmt.Sprintf("%s - this is expected when server shutdown has been initiated", err.Error()))
+			} else {
+				log.Error(nil, err, err.Error())
+			}
```
MatousJobanek, line 182: this is something that I noticed when I was testing the graceful shutdown. The proxy printed an error when the server was shut down
```
{"level":"error","ts":"2024-09-10T09:16:08.384Z","logger":"registration-service","msg":"http: Server closed","timestamp":"Tue, 10 Sep 2024 09:16:08 +0000","commit":"22b13b9","error":"http: Server closed","stacktrace":"github.com/codeready-toolchain/registration-service/pkg/log.(*Logger).Errorf\n\tgithub.com/codeready-toolchain/registration-service/pkg/log/log.go:191\ngithub.com/codeready-toolchain/registration-service/pkg/log.(*Logger).Error\n\tgithub.com/codeready-toolchain/registration-service/pkg/log/log.go:177\ngithub.com/codeready-toolchain/registration-service/pkg/log.Error\n\tgithub.com/codeready-toolchain/registration-service/pkg/log/log.go:111\ngithub.com/codeready-toolchain/registration-service/pkg/proxy.(*Proxy).StartProxy.func4\n\tgithub.com/codeready-toolchain/registration-service/pkg/proxy/proxy.go:178"}
```
which was a bit confusing and not expected

pkg/configuration/configuration.go
```diff
@@ -79,18 +78,9 @@ func GetRegistrationServiceConfig() RegistrationServiceConfig {
...
-	config, secrets, err := commonconfig.LoadLatest(cl, &toolchainv1alpha1.ToolchainConfig{})
-	if err != nil {
-		// return default config
-		logger.Error(err, "failed to force load RegistrationServiceConfig")
-		return RegistrationServiceConfig{cfg: &toolchainv1alpha1.ToolchainConfigSpec{}}, err
-	}
-	return NewRegistrationServiceConfig(config, secrets), nil
+// SetClient sets the client to be used to fetch the configuration
+func SetClient(cl client.Client) {
+	configurationClient = cl
```
MatousJobanek, line 83: This looks like Java setter, I know. It's there to make the current structure and test suites happy. As soon as we use the same cached client everywhere, then we will be able to drop this part.

cmd/main.go
```diff
@@ -154,43 +153,25 @@ func main() {
...
-
-	// We'll accept graceful shutdowns when quit via SIGINT (Ctrl+C) or SIGTERM
-	// (Ctrl+/). SIGKILL, SIGQUIT will not be caught.
-	signal.Notify(stop, os.Interrupt, syscall.SIGTERM, syscall.SIGINT)
-	sigReceived := <-stop
-	log.Infof(nil, "Signal received: %+v", sigReceived.String())
-
-	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+func gracefulShutdown(ctx context.Context, timeout time.Duration, hs ...*http.Server) {
+	<-ctx.Done()
```
mfrancisc, line 160: is this actually required ? I see the reading from the Done channel is already happening in the `Shutdown` function.

mfrancisc (commented):
Nice Job! 

I have only few comments/questions.

cmd/main.go
MatousJobanek: yes, it is - this is the notifier (gate keeper) that tells us when the kill signal was send and when we should start the graceful shutdown.

cmd/main.go
```diff
@@ -200,9 +181,37 @@ func configClient(cfg *rest.Config) (client.Client, error) {
...
+
+	hostCluster, err := runtimecluster.New(cfg, func(options *runtimecluster.Options) {
+		options.Scheme = scheme
+		// cache only in the host-operator namespace
+		options.Namespace = configuration.Namespace()
 	})
+	if err != nil {
+		return nil, err
+	}
+	go func() {
```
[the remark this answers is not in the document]
MatousJobanek: because it's a blocking function - if it wasn't executed in a separate goroutine then the execution would hang here and wouldn't go further.

mfrancisc (approved):
Great Job 🏅 

Thanks for answering my questions.

cmd/main.go
mfrancisc: ok, I see I thought the one in the select statement would suffice but I was wrong. Thanks.

cmd/main.go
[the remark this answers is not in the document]
mfrancisc: makes sense. I've missed that. Thanks for clarifying.

cmd/main.go
MatousJobanek: which select statement :thinking: ?

MatousJobanek: Ah, you mean in the `Shutdown` method - the select happens at the end of the shutdown logic when most of the stuff is already shut down and the logic closes the idled connections.

rsoaresd (approved):
LGTM! Great job 🚀

PR #462, opened 2024-09-20
replace methods of the CRTRestClient with the cached client

MatousJobanek:
This is a follow-up of https://github.com/codeready-toolchain/registration-service/pull/461

It replaces the methods of the custom CRTRestClient with the cached client that is already used instead of the custom shared informers. With this, we will use only one client in both proxy and reg-service which is cached and optimized.

I'm changing the implementation of the methods only to minimize the scope of the changes in the PR. 

Changes in unit-tests and removal of the extra layers will be addressed in many following (and most likely huge) PRs :grimacing: 

[KUBESAW-133](https://issues.redhat.com/browse/KUBESAW-133)


alexeykazakov (approved):
`+68 -221` I like this ratio for these reg-service refactoring PRs :)

mfrancisc (approved):
Nice cleanup 👍

PR #464, opened 2024-09-20
Add signup URL to authconfig payload

alexeykazakov:
Required for https://github.com/codeready-toolchain/registration-service/pull/438
There is a new redirect page in that PR ^^^  which is supposed to do the following:
- it loads the authconfig from `/api/v1/authconfig`
- then if the user is not logged in already then it redirects the user to SSO for login
- then it checks if the user has already signed up. If not then the page redirects the user to the signup page.

This PR adds the signup URL (we store it as `registrationServiceURL` in the configuration) to the authconfig payload, so it's available to the clients in the last step above.


ranakan19 (approved):
Looks good!

MatousJobanek (approved):

PR #466, opened 2024-09-24
use FakeClient as backend for InformersService in unit tests

MatousJobanek:
First PR to migrate unit tests to use FakeClient instead of the mocks. I was able to isolate the changes for the InformerService only. 
* Unit tests now rely on the FakeClient as the back-end of the InformerService in all unit tests where it was used. 
* I removed the "fake informer" as well as other extra mocks.
* I tried to minimize the changes, so I adjusted some formatting here and there accordingly

The main goal of this PR is to start using the FakeClient in unit tests, its purpose is not to improve the unit tests, or increase the coverage. This can (and should) be done later.

[KUBESAW-193](https://issues.redhat.com/browse/KUBESAW-193)

### DIFF
--- a/pkg/proxy/handlers/spacelister_get_test.go
+++ b/pkg/proxy/handlers/spacelister_get_test.go
@@ -12,6 +12,7 @@ import (
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 	"github.com/codeready-toolchain/registration-service/pkg/application/service"
 	rcontext "github.com/codeready-toolchain/registration-service/pkg/context"
+	infservice "github.com/codeready-toolchain/registration-service/pkg/informers/service"
 	"github.com/codeready-toolchain/registration-service/pkg/proxy/handlers"
 	"github.com/codeready-toolchain/registration-service/pkg/proxy/metrics"
 	proxytest "github.com/codeready-toolchain/registration-service/pkg/proxy/test"
@@ -20,14 +21,12 @@ import (
 	commoncluster "github.com/codeready-toolchain/toolchain-common/pkg/cluster"
 	commonproxy "github.com/codeready-toolchain/toolchain-common/pkg/proxy"
 	"github.com/codeready-toolchain/toolchain-common/pkg/test"
-	spacetest "github.com/codeready-toolchain/toolchain-common/pkg/test/space"
 	spacebindingrequesttest "github.com/codeready-toolchain/toolchain-common/pkg/test/spacebindingrequest"
 	"github.com/gin-gonic/gin"
 	"github.com/labstack/echo/v4"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"k8s.io/apimachinery/pkg/labels"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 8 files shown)


alexeykazakov (approved):
:+1:

pkg/proxy/handlers/spacelister_get_test.go
```diff
@@ -248,27 +246,29 @@ func testSpaceListerGet(t *testing.T, publicViewerEnabled bool) {
...
 				expectedErr:     "nstemplatetier error",
 				expectedErrCode: 500,
-				overrideInformerFunc: func() service.InformerService {
-					informerFunc := fake.GetInformerService(fakeClient, fake.WithGetNSTemplateTierFunc(func(_ string) (*toolchainv1alpha1.NSTemplateTier, error) {
-						return nil, fmt.Errorf("nstemplatetier error")
-					}))
-					return informerFunc()
+				mockFakeClient: func(fakeClient *test.FakeClient) {
+					fakeClient.MockGet = func(ctx context.Context, key runtimeclient.ObjectKey, obj runtimeclient.Object, opts ...runtimeclient.GetOption) error {
+						if key.Name == "base1ns" {
```
mfrancisc, line 261: minor - alternatively we could check (also) the object type.

Same for the other similar checks.

pkg/proxy/handlers/spacelister_get_test.go
```diff
@@ -711,20 +706,33 @@ func TestGetUserWorkspace(t *testing.T) {
...
+				// space
+				fake.NewSpace("batman", "member-1", "batman"),
+				fake.NewSpace("robin", "member-1", "robin"),
+
+				fake.NewSpaceBinding("robin-1", "robin", "robin", "admin"),
+				// 2 spacebindings to force the error
+				fake.NewSpaceBinding("batman-1", "batman", "batman", "admin"),
+				fake.NewSpaceBinding("batman-2", "batman", "batman", "maintainer"),
+				fake.NewSpaceBinding("community-robin", toolchainv1alpha1.KubesawAuthenticatedUsername, "robin", "viewer"),
+			)
```
mfrancisc, line 719: not sure if there was any specific reason for declaring this `fakeClient` for each test iteration. Maybe we could move it back outside ( not a big performance improvement but still ) .

mfrancisc (approved):
Nice job, the code looks simpler 👍 

I have left some very minor comments , only aesthetic  stuff so feel free to ignore those.

pkg/proxy/handlers/spacelister_get_test.go
MatousJobanek: We cannot use `TypeMeta` as this is empty. we would need to either check the actual go type of the object, or get the kind from the schema. TBH, both of these options seem to be an overkill for these unit tests. I don't think that we would return an error in other cases than the expected ones.
but I'm fine with adding it there if you see some possible problems with the current approach.

pkg/proxy/handlers/spacelister_get_test.go
MatousJobanek: the reason for moving it there was the mock functions - some of the iterations mock one of the FakeClient functions and we want to do it in the context of the iteration only, so it doesn't affect other test cases.

pkg/proxy/handlers/spacelister_get_test.go
MatousJobanek: casting and checking the type in this way is actually simple and can be done in single-line. so I updated the mocks: 0194c4c

pkg/proxy/handlers/spacelister_get_test.go
mfrancisc: ok makes sense, thanks for pointing that out.

pkg/proxy/handlers/spacelister_get_test.go
mfrancisc: right, that is what I was thinking. Thanks for addressing my comment.

mfrancisc (approved):

PR #469, opened 2024-09-30
use FakeClient as backend for RESTClient based unit tests

MatousJobanek:
It's a follow-up of the previous PR doing changes in unit-tests. This time for the CRTClient use-cases. 
The PR is a bit bigger, but there is no chance to make it smaller. I could separate the removal, but I wanted to highlight what has been removed/replaced by these changes.

The same as before, the PR is about dropping the old mocks of the CRTClient by replacing their usage with the FakeClient. It doesn't try to improve the coverage nor the tests themself. 
In the modified tests, I also dropped the usage of the MockableApplication type. I'll do the same also in the other tests in the following PR(s). There is still so many things to be removed and optimized :-). 

[KUBESAW-193](https://issues.redhat.com/browse/KUBESAW-193)


pkg/application/service/factory/service_factory.go
```diff
@@ -88,6 +88,12 @@ func (s *ServiceFactory) WithVerificationServiceOption(opt verificationservice.V
 	s.verificationServiceOptions = append(s.verificationServiceOptions, opt)
 }
 
+func (s *ServiceFactory) WithSignupService(signupService service.SignupService) {
+	s.signupServiceFunc = func(_ ...signupservice.SignupServiceOption) service.SignupService {
+		return signupService
+	}
+}
```
MatousJobanek, line 95: This is a temp hack to inject specific fake version of signupService without any extra abstraction and additional mocking.

pkg/signup/service/resource_provider.go
```diff
@@ -14,26 +14,26 @@ type ResourceProvider interface {
 	ListSpaceBindings(reqs ...labels.Requirement) ([]toolchainv1alpha1.SpaceBinding, error)
 }
 
-type crtClientProvider struct {
-	cl kubeclient.CRTClient
+type CrtClientProvider struct {
+	Cl kubeclient.CRTClient
```
MatousJobanek, line 18: Also a temp hack until this type is gonna be removed completely

alexeykazakov (approved):
Thanks for taking care of it!

test/util/service_context.go
```diff
@@ -0,0 +1,23 @@
...
+
+	"github.com/codeready-toolchain/registration-service/pkg/kubeclient"
+	"github.com/codeready-toolchain/registration-service/pkg/server"
+	"github.com/codeready-toolchain/registration-service/test/fake"
+	commontest "github.com/codeready-toolchain/toolchain-common/pkg/test"
+	"github.com/stretchr/testify/require"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func NewMemberClusterServiceContext(t *testing.T, cl client.Client) fake.MemberClusterServiceContext {
```
mfrancisc, line 14: Not related to your PR , but I'm a bit confused with the name of the function and the returning struct. 

Why does it refer to `Member` since it seems to contain a rest client for the host namespace and an application type ( which BTW it's again another rest client for the host operator namespace ) ?

test/util/service_context.go
```diff
@@ -0,0 +1,23 @@
+package util
```
mfrancisc, line 1: minor - there is already a file named `service_context.go` in the context package. Not a big deal since that is allowed by Go, but when searching for the file name sometimes can be confusing.

mfrancisc (commented):
I like the new code structure, it's simpler to read and more maintainable, great job 👏 

I have only two minor comments.

test/util/service_context.go
MatousJobanek: I have no clue why the type is named in this way :man_shrugging: 
I left it the same name to not incorporate even more changes and named the function accordingly. I hope that I'll be able to drop both of them soon.

test/util/service_context.go
MatousJobanek: Good catch, I didn't notice that. That was just a quick workaround for an import loop that I faced in unit tests. I believe that I'll drop this file soon as well. 
As for the name, any other name would have the same problem (service, context, etc...). Let's keep it and drop it as part of the next refactoring efforts

test/util/service_context.go
mfrancisc: Ok makes sense. Thanks for answering and taking a look into it. I think ideally we would drop both indeed.

test/util/service_context.go
mfrancisc: I'm fine with keeping it and removing it later 👍

mfrancisc (approved):

PR #470, opened 2024-09-30
fix: log errors inside VerifyPhoneCode

mfrancisc:
This PR moves all the logging inside the `VerifyPhoneCode` function and removes the log error from the caller since the function might return an error that doesn't need to be logged which was already logged as `Info`.


alexeykazakov (approved):
Thanks!

rajivnathan (approved):

xcoulon (approved):

MatousJobanek (approved):
/lgtm

PR #473, opened 2024-10-02
drop CRTClient abstraction

MatousJobanek:
* drops the CRTClient abstraction and all the related types. 
* It uses only the InformerService as the ResourceProvider for getting ifnromation about UserSignups. The InformerService will go away as well, it's just getting harder to make incremental changes :exploding_head: 
    * This also simplifies some of the unit-tests as we don't have to test two different types of implementation of the  ResourceProvider
* I introduced a new type called `namespaced.Client` which holds the `Client` itself plus the host operator namespace. With this type it's much easier to propagate the Client and the namespace in the services. 
* It now calls the client directly (instead of relying on the abstraction & types), in those cases where the CRTClient was used. 

[KUBESAW-197](https://issues.redhat.com/browse/KUBESAW-197)


test/util/service_context.go
```diff
@@ -3,21 +3,16 @@ package util
...
-	application, err := server.NewInClusterApplication(cl, commontest.HostOperatorNs)
-	require.NoError(t, err)
+func NewMemberClusterServiceContext(_ *testing.T, cl client.Client) fake.MemberClusterServiceContext {
 	return fake.MemberClusterServiceContext{
-		CrtClient: crtClient,
-		Svcs:      application,
+		NamespacedClient: namespaced.NewClient(cl, commontest.HostOperatorNs),
+		Svcs:             server.NewInClusterApplication(cl, commontest.HostOperatorNs),
 	}
 }
```
MatousJobanek, line 18: This will go away in one of the following PRs. Just wanted to minimize the number of changes so I left it here for now.

pkg/signup/service/signup_service.go
```diff
@@ -589,33 +595,49 @@ func (s *ServiceImpl) DoGetUserSignupFromIdentifier(provider ResourceProvider, u
...
 // PhoneNumberAlreadyInUse checks if the phone number has been banned. If so, return
-// an internal server error. If not, check if an active (non-deactivated) UserSignup with a different userID and username
+// an internal server error. If not, check if an approved UserSignup with a different userID and username
 // and email address exists. If so, return an internal server error. Otherwise, return without error.
 // Either the actual phone number, or the md5 hash of the phone number may be provided here.
 func (s *ServiceImpl) PhoneNumberAlreadyInUse(userID, username, phoneNumberOrHash string) error {
-	bannedUserList, err := s.CRTClient().V1Alpha1().BannedUsers().ListByPhoneNumberOrHash(phoneNumberOrHash)
-	if err != nil {
+	labelValue := hash.EncodeString(phoneNumberOrHash)
+	if md5Matcher.Match([]byte(phoneNumberOrHash)) {
```
mfrancisc, line 615: is strange that the md5 package doesn't have a function for identifying if a string is md5 hash. 

Also should we add a unit tests for this case ?

pkg/namespaced/client.go
```diff
@@ -0,0 +1,22 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func NewClient(client client.Client, namespace string) Client {
+	return Client{Client: client, Namespace: namespace}
+}
+
+type Client struct {
+	client.Client
+	Namespace string
```
mfrancisc, line 14: I see the namespace ATM is only used by the `signup_service` . I was wondering if it will be used in other places , otherwise we could stick to the `client.Client` type only and pass the namespace to the signupService only. Just an idea.

mfrancisc (commented):
Looks Good 👍 

I have only few minor comments.

pkg/namespaced/client.go
MatousJobanek: Not only there,  it's used also via the `NamespacedName` method
https://github.com/codeready-toolchain/registration-service/blob/08d17055a7db2653f25fe72126a85be9a5b4a592/pkg/namespaced/client.go#L17-L22
in all cases when the code gets a resources from the namespace.

pkg/signup/service/signup_service.go
MatousJobanek: I took a quick glance and I didn't find anything. But TBH, this is not really needed - it can be done without checking the format of the provided value, it just requires second function/method which is hard with the current setup. Can be done later - let me create a story for it.

MatousJobanek: https://issues.redhat.com/browse/KUBESAW-200

pkg/signup/service/signup_service.go
```diff
@@ -589,33 +595,49 @@ func (s *ServiceImpl) DoGetUserSignupFromIdentifier(provider ResourceProvider, u
...
+	}
+	userSignups := &toolchainv1alpha1.UserSignupList{}
+	if err := s.List(gocontext.TODO(), userSignups, client.InNamespace(s.Namespace), labelSelector); err != nil {
 		return errors.NewInternalError(err, "failed listing userSignups")
 	}
-	for _, signup := range userSignups {
-		if signup.Spec.IdentityClaims.Sub != userID && signup.Spec.IdentityClaims.PreferredUsername != username && !states.Deactivated(signup) { // nolint:gosec
+
+	for _, signup := range userSignups.Items {
+		userSignup := signup
```
xcoulon, line 639: is it needed? IIRC, `signup` is not a pointer

xcoulon (approved):

pkg/signup/service/signup_service.go
MatousJobanek: no, but I create a pointer to the `userSignup` in the next line and since we still use go 1.20 we have to do this re-assignment

pkg/signup/service/signup_service.go
mfrancisc: Thanks for creating the Jira, and yeah it would be nicer if we could remove that regex matcher :)

pkg/namespaced/client.go
mfrancisc: Ok I see, I've missed that. then yes it makes sense to have a new struct holding also the namespace. thanks for the explanation.

mfrancisc (approved):

alexeykazakov (approved):

pkg/signup/service/signup_service.go
MatousJobanek: added a comment 8df7f52

xcoulon: great, thanks!

xcoulon (approved):
thanks for taking my comment into account, @MatousJobanek !

PR #475, opened 2024-10-03
Drop ResourceProvider abstraction

MatousJobanek:
Drops the ResourceProvider abstraction/interface and call the client directly instead

follow-up of: [KUBESAW-197](https://issues.redhat.com/browse/KUBESAW-197)
preparation for: [KUBESAW-198](https://issues.redhat.com/browse/KUBESAW-198)


pkg/signup/service/signup_service.go
```diff
@@ -654,34 +643,27 @@ func (s *ServiceImpl) PhoneNumberAlreadyInUse(userID, username, phoneNumberOrHas
...
-			log.Errorf(nil, err, "unable to create spacebindings selector for MUR %s", murName)
-			return "", ""
-		}
-
-		requirements := []labels.Requirement{*sbSelector}
-
-		sbs, err := provider.ListSpaceBindings(requirements...)
-		if err != nil {
+		sbs := &toolchainv1alpha1.SpaceBindingList{}
+		if err := cl.List(gocontext.TODO(), sbs, client.InNamespace(cl.Namespace),
```
alexeykazakov, line 649: Not for this PR but it would be nice to propagate the original context here.

alexeykazakov (approved):

pkg/signup/service/signup_service.go
MatousJobanek: Yes, I know. I decided to do it later as there are two other types of contexts used in the code - gin & echo - (plus the standard go context), and some of the code is called from both locations. Gin implements the standard go context interface, but echo doesn't. 
I have a story for this: [KUBESAW-199](https://issues.redhat.com/browse/KUBESAW-199)

MatousJobanek: I can actually change the places where I can easily use the gin context, for the other places I would need to update the method signature which would introduce many other changes, so let me do it as a separate PR later
usage of the gin context (where it's easy to do) done here: 8187796

pkg/application/service/services.go
```diff
@@ -11,8 +11,6 @@ import (
 type InformerService interface {
 	GetMasterUserRecord(name string) (*toolchainv1alpha1.MasterUserRecord, error)
 	GetSpace(name string) (*toolchainv1alpha1.Space, error)
-	GetToolchainStatus() (*toolchainv1alpha1.ToolchainStatus, error)
-	GetUserSignup(name string) (*toolchainv1alpha1.UserSignup, error)
 	ListSpaceBindings(reqs ...labels.Requirement) ([]toolchainv1alpha1.SpaceBinding, error)
```
mfrancisc, line 14: just for my understanding - why dropping only these two methods and not all the ones from the resource_provider.go ?

mfrancisc (approved):
Looks good 👍 

I have only one question/clarification.

pkg/application/service/services.go
MatousJobanek: This is the InformerService interface, not ResourceProvider. 
I dropped only those methods that are not used anywhere now, the rest is still used in the proxy logic

mfrancisc: I see, maybe I got lost/confused by the fact that resource provider and informerservice had overlapping methods 😕  :
```
type ResourceProvider interface {
	GetMasterUserRecord(name string) (*toolchainv1alpha1.MasterUserRecord, error)
	GetToolchainStatus() (*toolchainv1alpha1.ToolchainStatus, error)
	GetUserSignup(name string) (*toolchainv1alpha1.UserSignup, error)
	GetSpace(name string) (*toolchainv1alpha1.Space, error)
	ListSpaceBindings(reqs ...labels.Requirement) ([]toolchainv1alpha1.SpaceBinding, error)
}
```

MatousJobanek: Yes, they had - InformerService was one of the implementations of ResourceProvider, but the functions were not used only for the purpose of the ResourceProvider, but also in other cases (in proxy). But don't worry, the time when I'll be able to drop the InfomerService is coming closer :-)

mfrancisc: Yeah thanks, and sorry for the confusion I got lost with all the abstractions 😄

MatousJobanek: No worries, I get lost in all the abstraction every time when I'm working on it :smile:

PR #477, opened 2024-10-04
drop BaseService from MemberClusterService & use Client directly

MatousJobanek:
* drops BaseService from MemberClusterService
* uses the controller-runtime client directly for communication with k8s api
* the implementation of the SignupService is set directly (not via the BaseService) which simplifies the tests

this is a prerequisite for dropping the InformerService  [KUBESAW-198](https://issues.redhat.com/browse/KUBESAW-198)


alexeykazakov (approved):

mfrancisc (approved):
Nice 👍

PR #479, opened 2024-10-10
drop InformerService

MatousJobanek:
[KUBESAW-198](https://issues.redhat.com/browse/KUBESAW-198)

Drops completely the InformerService interface and its implementation. All usages are replaced by calling the client directly.


alexeykazakov (approved):
Nice!

PR #481, opened 2024-10-15
drop MockableApplication

MatousJobanek:
drop MockableApplication as it can be replaced with the "InClusterApplication" now

_no urgency with the review during the F2F preparation week - I had the changes locally from the last week, I wanted push it so I don't forget about it._


alexeykazakov (approved):

ranakan19 (approved):
Looks Good

test/util/application.go
```diff
@@ -11,6 +11,11 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+func PrepareInClusterApplication(t *testing.T, objects ...runtime.Object) application.Application {
```
mfrancisc, line 14: minor - It might seem a bit unclear at first to have functions named `PrepareInClusterApplication` and `PrepareInClusterApp` , but I see that you wanted to have one which returns only the `Application` type without the `fakeClient` , not sure how we could improve this.

mfrancisc (approved):
Great Job 👏 

Thanks for cleaning this up 🏅

test/util/application.go
MatousJobanek: Yes, I know that the names can be a bit confusing. I believe that I'll be able to improve them or get rid of one of them in the following PRs.

PR #483, opened 2024-11-11
drop MemberClusterService interface & rename the implementation

MatousJobanek:
* drop MemberClusterService interface 
* rename the implementation to `MemberClusters`
* move it to the `proxy` package
* provide the `GetMemberClustersFunc` as a parameter, so it uses the same as is used in `proxy.go`
    * this also slightly simplifies the tests
* drop "ServiceImpl options" as they are not needed anymore
* drop `ProxyFakeApp` as it is not needed anymore

### DIFF
--- a/pkg/proxy/members_test.go
+++ b/pkg/proxy/members_test.go
@@ -1,4 +1,4 @@
-package service_test
+package proxy_test
 
 import (
 	"context"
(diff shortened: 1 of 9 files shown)


alexeykazakov (approved):

pkg/proxy/members_test.go
```diff
@@ -166,7 +166,7 @@ func (s *TestClusterServiceSuite) TestGetClusterAccess() {
 						fakeClient.MockGet = nil
 					}()
 					nsClient := namespaced.NewClient(fakeClient, commontest.HostOperatorNs)
-					svc := service.NewMemberClusterService(nsClient, sc)
+					svc := proxy.NewMemberClusters(nsClient, sc, commoncluster.GetMemberClusters)
```
mfrancisc, line 169: very minor - I guess we could rename also these `svc` which stands for service if I'm not mistaken

mfrancisc (approved):
Really nice simplification 👍

pkg/proxy/members_test.go
MatousJobanek: good catch, I didn't pay much attention to the names in the test file.
renamed in 17a3ed7
Thanks

PR #486, opened 2024-11-20
metrics: measure roundtrips in seconds

xcoulon:
not in milliseconds, which results in histogram buckets under '1' being always empty


MatousJobanek (approved):

xcoulon:
thanks for the quick review @MatousJobanek ! 🙌

PR #492, opened 2025-01-09
use helper function for UserSignup creation in verification tests

MatousJobanek:
This PR replaces manual creation of UserSignups with helper functions from toolchain common.

**Why?** 
My ultimate goal is to clean up the leftovers that stayed there after the migration of the naming format of UserSignups that happened a few years ago.
https://github.com/codeready-toolchain/registration-service/blob/690bb9c2518438097c522edd9745932aac4badb9/pkg/signup/service/signup_service.go#L557-L575
We don't use the `sub` claim (userID) for UserSignups for a few years, but most of the unit tests still do that. This change is to minimize the final PR(s) that will drop the usage & support of the UserSignups with names equal to the `sub` claim (userID).

### DIFF
--- a/pkg/verification/service/verification_service_test.go
+++ b/pkg/verification/service/verification_service_test.go
@@ -339,8 +339,8 @@ func (s *TestVerificationServiceSuite) TestInitVerificationPassesWhenMaxCountRea
 	userSignup := testusersignup.NewUserSignup(
 		testusersignup.WithName("johny"),
 		testusersignup.WithLabel(toolchainv1alpha1.UserSignupUserPhoneHashLabelKey, "+1NUMBER"),
-		testusersignup.WithAnnotation(toolchainv1alpha1.UserSignupVerificationInitTimestampAnnotationKey, now.Format(verificationservice.TimestampLayout)),
-		testusersignup.WithAnnotation(toolchainv1alpha1.UserVerificationAttemptsAnnotationKey, "3"),
+		testusersignup.WithAnnotation(toolchainv1alpha1.UserSignupVerificationInitTimestampAnnotationKey, now.Add(-25*time.Hour).Format(verificationservice.TimestampLayout)),
+		testusersignup.WithAnnotation(toolchainv1alpha1.UserSignupVerificationCounterAnnotationKey, "3"),
 		testusersignup.WithAnnotation(toolchainv1alpha1.UserSignupVerificationCodeAnnotationKey, "123456"),
 		testusersignup.VerificationRequiredAgo(time.Second))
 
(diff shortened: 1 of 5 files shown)


pkg/verification/service/verification_service_test.go
```diff
@@ -134,43 +134,17 @@ func (s *TestVerificationServiceSuite) TestInitVerification() {
...
-		Spec: toolchainv1alpha1.UserSignupSpec{
-			IdentityClaims: toolchainv1alpha1.IdentityClaimsEmbedded{
-				PreferredUsername: "sbryzak@redhat.com",
-			},
-		},
-	}
+	userSignup := testusersignup.NewUserSignup(
+		testusersignup.WithName("johny"),
+		testusersignup.WithLabel(toolchainv1alpha1.UserSignupUserPhoneHashLabelKey, "+1NUMBER"),
+		testusersignup.VerificationRequired(0))
```
mfrancisc, line 140: does this have the same effect as :
```
states.SetVerificationRequired(userSignup, true)
```
?
It looks like the latter sets the `UserSignupStateVerificationRequired` in the `UserSignup.Spec.States` and it doesn't update the conditions .

mfrancisc: minor and not related to this PR - also I found it a bit confusing at first to see the `0` value as parameter `testusersignup.VerificationRequired(0)`

mfrancisc (approved):
Looks good! 
Thanks for cleaning it up.

I have only one minor comment/question.

pkg/verification/service/verification_service_test.go
MatousJobanek: Yeah, it's weird - I've created a PR to align the names a bit https://github.com/codeready-toolchain/toolchain-common/pull/445

>  It looks like the latter sets the UserSignupStateVerificationRequired in the UserSignup.Spec.States and it doesn't update the conditions

you are right, the `testusersignup.VerificationRequired` updates also the conditions which wasn't done before, technically speaking, the UserSignup should look like that, so it's closer to the real scenario now.

alexeykazakov (approved):

pkg/verification/service/verification_service_test.go
```diff
@@ -375,27 +336,13 @@ func (s *TestVerificationServiceSuite) TestInitVerificationPassesWhenMaxCountRea
...
-			IdentityClaims: toolchainv1alpha1.IdentityClaimsEmbedded{
-				PreferredUsername: "shane@redhat.com",
-			},
-		},
-	}
-	states.SetVerificationRequired(userSignup, true)
+	userSignup := testusersignup.NewUserSignup(
+		testusersignup.WithName("johny"),
+		testusersignup.WithLabel(toolchainv1alpha1.UserSignupUserPhoneHashLabelKey, "+1NUMBER"),
+		testusersignup.WithAnnotation(toolchainv1alpha1.UserSignupVerificationInitTimestampAnnotationKey, now.Format(verificationservice.TimestampLayout)),
```
metlos, line 342: The old code has `now.Add(-25 * time.Hour).Format(verificationservice.TimestampLayout)`. Is it ok to move the timestamp 25 hours forward? I don't understand the code at all, just observing the difference...

MatousJobanek: good catch, thanks. There was apparently also a bug in the actual prod code, fixed in https://github.com/codeready-toolchain/registration-service/commit/7891733c3928a90de88ac0feb0236057dc9c6f76

PR #494, opened 2025-01-16
use helper function for UserSignup creation in signup unit tests

MatousJobanek:
a follow-up of https://github.com/codeready-toolchain/registration-service/pull/492


pkg/controller/signup_test.go
```diff
@@ -644,29 +586,11 @@ func (s *TestSignupSuite) TestVerifyPhoneCodeHandler() {
...
-				IdentityClaims: crtapi.IdentityClaimsEmbedded{
-					PreferredUsername: "jsmith",
-					PropagatedClaims: crtapi.PropagatedClaims{
-						Sub: otherUserID,
-					},
-				},
-			},
-			Status: crtapi.UserSignupStatus{},
-		}
+		otherUserSignup := testusersignup.NewUserSignup(
```
mfrancisc, line 589: minor and not strictly related to your changes - it looks like now there is not much difference in how the "usersignup it's stored" compared to the other tests. I mean I'm not sure what the test case description refers to. I thought it was about how the `otherUserSignup` is configured ( with the `PropagatedClaims` filed being populated ) .

mfrancisc (approved):
Nice 👍 

Thanks for cleaning it up.

pkg/controller/signup_test.go
```diff
@@ -99,32 +95,24 @@ func (s *TestSignupSuite) TestSignupPostHandler() {
...
-			Spec: crtapi.UserSignupSpec{
-				IdentityClaims: crtapi.IdentityClaimsEmbedded{
-					PreferredUsername: "bill",
-				},
-			},
-			Status: crtapi.UserSignupStatus{
-				Conditions: []crtapi.Condition{
+		signup := testusersignup.NewUserSignup(
+			testusersignup.WithName("bill"),
+			func(userSignup *crtapi.UserSignup) {
```
metlos, line 102: `testusersignup.WithIncompleteSignup("test_reason", "test_message")` would make the test more readable IMHO, even if it this would be the single use of it.

pkg/controller/signup_test.go
```diff
@@ -99,32 +95,24 @@ func (s *TestSignupSuite) TestSignupPostHandler() {
...
 						Message: "test_message",
 					},
-				},
-			},
-		}
+				}
+			})
 
 		svc.MockSignup = func(ctx *gin.Context) (*crtapi.UserSignup, error) {
 			assert.Equal(s.T(), expectedUserID, ctx.GetString(context.SubKey))
```
metlos, line 114: Is this just testing `gin`?

pkg/controller/signup_test.go
```diff
@@ -300,11 +275,11 @@ func (s *TestSignupSuite) TestInitVerificationHandler() {
 			BodyString("")
 
 		data := []byte(fmt.Sprintf(`{"phone_number": "%s", "country_code": "1"}`, phoneNumber))
-		rr := initPhoneVerification(s.T(), handler, gin.Param{}, data, userID, "", http.MethodPut, "/api/v1/signup/verification")
+		rr := initPhoneVerification(s.T(), handler, gin.Param{}, data, userID, "johny", http.MethodPut, "/api/v1/signup/verification")
```
metlos, line 278: grammar nazi nitpick: "johnny"

pkg/controller/signup_test.go
MatousJobanek: Yes, you are completely right. It's a leftover that wasn't properly cleaned up after migration of UserSignups from using userIds to using usernames as the name of the resource. The end-goal is to clean this up, but this will require a lot of changes, so I'm doing it in small batches (for now some changes in unit tests that will help me to minimize the size of the changes of the final PRs). 
As for this test. It's not about how it's stored, but how it is retrieved - this test doesn't use the userID. Anyway, this sub-test will go away in the upcoming PR.

pkg/controller/signup_test.go
MatousJobanek: done in a103f78

pkg/controller/signup_test.go
MatousJobanek: 638f7ab

PR #500, opened 2025-01-24
move EncodeUserIdentifier & drop userID from tests

MatousJobanek:
depends on: https://github.com/codeready-toolchain/toolchain-common/pull/452

* moves the EncodeUserIdentifier to toolchain-common (so it can be used in the helper function)
* drops usage of userID from unit tests
* Improves the unit-tests so they use an encoded value as the name of the UserSignup CRs
* applies new helper functions introduced in the toolchain-common PR

[KUBESAW-254](https://issues.redhat.com/browse/KUBESAW-254)

This is hopefully the last PR before final dropping of the usage of userIDs as the UserSignup names

### DIFF
--- a/pkg/controller/signup_test.go
+++ b/pkg/controller/signup_test.go
@@ -28,7 +28,6 @@ import (
 	testconfig "github.com/codeready-toolchain/toolchain-common/pkg/test/config"
 	testsocialevent "github.com/codeready-toolchain/toolchain-common/pkg/test/socialevent"
 	testusersignup "github.com/codeready-toolchain/toolchain-common/pkg/test/usersignup"
-	apiv1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/gin-gonic/gin"
@@ -95,16 +94,16 @@ func (s *TestSignupSuite) TestSignupPostHandler() {
 		require.NoError(s.T(), err)
 		expectedUserID := ob.String()
 		ctx.Set(context.SubKey, expectedUserID)
-		ctx.Set(context.UsernameKey, "bill")
+		ctx.Set(context.UsernameKey, "bill@kubesaw")
 		ctx.Set(context.EmailKey, expectedUserID+"@test.com")
 		signup := testusersignup.NewUserSignup(
-			testusersignup.WithName("bill"),
-			IncompleteUserSignupCondition(),
+			testusersignup.WithEncodedName("bill@kubesaw"),
+			testusersignup.SignupIncomplete("test_reason", "test_message"),
 		)
 
 		svc.MockSignup = func(ctx *gin.Context) (*crtapi.UserSignup, error) {
 			assert.Equal(s.T(), expectedUserID, ctx.GetString(context.SubKey))
-			assert.Equal(s.T(), "bill", ctx.GetString(context.UsernameKey))
+			assert.Equal(s.T(), "bill@kubesaw", ctx.GetString(context.UsernameKey))
 			assert.Equal(s.T(), expectedUserID+"@test.com", ctx.GetString(context.EmailKey))
 			return signup, nil
 		}
@@ -147,19 +146,6 @@ func (s *TestSignupSuite) TestSignupPostHandler() {
 	})
 }
 
-func IncompleteUserSignupCondition() testusersignup.Modifier {
-	return func(userSignup *crtapi.UserSignup) {
-		userSignup.Status.Conditions = []crtapi.Condition{
-			{
-				Type:    crtapi.UserSignupComplete,
-				Status:  apiv1.ConditionFalse,
-				Reason:  "test_reason",
-				Message: "test_message",
-			},
-		}
-	}
-}
-
 func (s *TestSignupSuite) TestSignupGetHandler() {
 	// Create a request to pass to our handler. We don't have any query parameters for now, so we'll
 	// pass 'nil' as the third parameter.
@@ -175,7 +161,7 @@ func (s *TestSignupSuite) TestSignupGetHandler() {
 	// Create UserSignup
 	ob, err := uuid.NewV4()
 	require.NoError(s.T(), err)
-	userID := ob.String()
+	username := ob.String()
 
 	// Create Signup controller instance.
 	ctrl := controller.NewSignup(application)
@@ -186,7 +172,7 @@ func (s *TestSignupSuite) TestSignupGetHandler() {
 		rr := httptest.NewRecorder()
 		ctx, _ := gin.CreateTestContext(rr)
 		ctx.Request = req
-		ctx.Set(context.SubKey, userID)
+		ctx.Set(context.UsernameKey, username)
 
 		targetCluster, err := uuid.NewV4()
 		require.NoError(s.T(), err)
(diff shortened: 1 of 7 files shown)


mfrancisc (approved):
Looks good!

Thanks for cleaning it up.
I have only a couple of minor questions.

alexeykazakov (approved):

pkg/controller/signup_test.go
```diff
@@ -95,16 +94,16 @@ func (s *TestSignupSuite) TestSignupPostHandler() {
 		require.NoError(s.T(), err)
 		expectedUserID := ob.String()
 		ctx.Set(context.SubKey, expectedUserID)
-		ctx.Set(context.UsernameKey, "bill")
+		ctx.Set(context.UsernameKey, "bill@kubesaw")
 		ctx.Set(context.EmailKey, expectedUserID+"@test.com")
 		signup := testusersignup.NewUserSignup(
-			testusersignup.WithName("bill"),
-			IncompleteUserSignupCondition(),
+			testusersignup.WithEncodedName("bill@kubesaw"),
```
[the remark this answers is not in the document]
MatousJobanek: All UserSignups are stored with the encoded username. The reg-service gets SSO with the username and it needs to encode it when storing/retrieving the UserSignup in/from etcd.
This technically simulates the real scenario when the original format of the username is not DNS compliant and should be encoded. Applying this in all tests ensures that all code encodes the username before storing/retrieving it.

pkg/proxy/handlers/spacelister.go
```diff
@@ -41,10 +41,9 @@ func NewSpaceLister(client namespaced.Client, app application.Application, proxy
 }
 
 func (s *SpaceLister) GetProvisionedUserSignup(ctx echo.Context) (*signup.Signup, error) {
-	userID, _ := ctx.Get(context.SubKey).(string)
 	username, _ := ctx.Get(context.UsernameKey).(string)
 
-	userSignup, err := s.GetSignupFunc(nil, userID, username, false)
+	userSignup, err := s.GetSignupFunc(nil, "", username, false)
```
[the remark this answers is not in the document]
MatousJobanek: yup, in the next PR.

pkg/signup/service/signup_service_test.go
```diff
@@ -674,7 +577,7 @@ func (s *TestSignupServiceSuite) TestGetUserSignupFails() {
 
 	fakeClient, application := testutil.PrepareInClusterApp(s.T())
 	fakeClient.MockGet = func(ctx gocontext.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
-		if _, ok := obj.(*toolchainv1alpha1.UserSignup); ok && key.Name != username {
+		if _, ok := obj.(*toolchainv1alpha1.UserSignup); ok && key.Name == username {
```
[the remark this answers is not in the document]
MatousJobanek: The current code tries to get the UserSignup with the username first, if not present then with the userID. Since there is no UserSignup object stored in the fake client, then the logic tries both (even though the userID is empty).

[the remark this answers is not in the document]
MatousJobanek: TBH, the original code was just simply wrong, it worked only because there are two attempts of getting the UserSignup. This will be changed in the upcoming RP.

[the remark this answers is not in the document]
mfrancisc: yeah it wasn't obvious at first, thanks for the details!

pkg/controller/signup_test.go
```diff
@@ -95,16 +94,16 @@ func (s *TestSignupSuite) TestSignupPostHandler() {
...
+			testusersignup.SignupIncomplete("test_reason", "test_message"),
 		)
 
 		svc.MockSignup = func(ctx *gin.Context) (*crtapi.UserSignup, error) {
 			assert.Equal(s.T(), expectedUserID, ctx.GetString(context.SubKey))
-			assert.Equal(s.T(), "bill", ctx.GetString(context.UsernameKey))
```
metlos, line 107: What is the purpose of these asserts here? As far as I understand the flow, they're only checking that we set the appropriate values in the context a couple of lines above. Is that really needed?

What we're really should be testing there, IMHO, is that the signup service has been used. So what about instead of all those asserts on the context, having a "signupCalled" boolean that would be set to true in the `svc.MockSignup` and checked for being true after the call to handler?

Or maybe that's assumed by the correct status code of the response and we don't need any of that either?

pkg/controller/signup_test.go
```diff
@@ -481,7 +444,7 @@ func (s *TestSignupSuite) TestVerifyPhoneCodeHandler() {
 
 		require.Equal(s.T(), "Internal Server Error", bodyParams["status"])
 		require.InDelta(s.T(), float64(500), bodyParams["code"], 0.01)
-		require.Equal(s.T(), fmt.Sprintf("no user: error retrieving usersignup: %s", userID), bodyParams["message"])
+		require.Equal(s.T(), "no user: error retrieving usersignup: ", bodyParams["message"])
```
metlos, line 447: A message ending with a ":" looks kinda unfinished. Where do we store the rest of it?

pkg/controller/signup_test.go
```diff
@@ -507,7 +470,7 @@ func (s *TestSignupSuite) TestVerifyPhoneCodeHandler() {
 
 		require.Equal(s.T(), "Not Found", bodyParams["status"])
 		require.InDelta(s.T(), float64(404), bodyParams["code"], 0.01)
-		require.Equal(s.T(), "usersignups.toolchain.dev.openshift.com \"jsmith\" not found: user not found", bodyParams["message"])
+		require.Equal(s.T(), "usersignups.toolchain.dev.openshift.com \"fdebf2d6-jsmithkubesaw\" not found: user not found", bodyParams["message"])
```
metlos, line 473: where is the "fdebf2d6" part coming from? Maybe it'd be worth explaining it with a comment. I personally haven't yet figured it out.

pkg/controller/signup_test.go
MatousJobanek: good catch, just wait for the next PR :-) 
The thing is that I've modified the tests to rely on usernames only, but not the prod code yet. The prod code still tries to retrieve the UserSignup with username and userID, if there is some problem (like in this test case), the code returns an error with userID information only (which is empty in this case).
It's pretty hard to isolate some small incremental changes, but at the same time, the PR that would contain all related code would be simply huge.

pkg/controller/signup_test.go
MatousJobanek: TBH, (to keep my sanity intact) I didn't try to question the tests themself, I only modified the related changes (or dropped tests that didn't make sense anymore after my changes). 
There are many things weird, wrong, missing, or redundant in the unit tests, I hope that we will get to fixing them later as soon as the whole code is simplified/refactored. 
Anyway, I tried to improve the test here: ec1eefa

MatousJobanek: and as for you question:
> What we're really should be testing there

As I understand it, it verifies that the signup handler is called properly an returns corresponding response values

pkg/controller/signup_test.go
MatousJobanek: it's the result of encoding the username - see https://github.com/codeready-toolchain/toolchain-common/blob/8ef21eb833fa3176af6cd0cf3842a8b64dee7dfa/pkg/usersignup/usersignup.go#L82-L123
added a comment: 7e48ae0

fbm3307 (approved):

MatousJobanek:
@metlos I've address/answered all your review comments, let me know if it looks good to you now.

metlos (approved):

PR #505, opened 2025-02-10
drop verification-service option & use DefaultTransport instead

MatousJobanek:
when the `DefaultTransport` is used, then there is no need to mock the HTTP Client to make gock working, thus we can simplify the tests and the other related code.

### DIFF
--- a/pkg/application/service/factory/service_factory.go
+++ b/pkg/application/service/factory/service_factory.go
@@ -34,12 +34,9 @@ func (s *serviceContextImpl) Services() service.Services {
 }
 
 type ServiceFactory struct {
-	contextProducer            servicecontext.ServiceContextProducer
-	serviceContextOptions      []ServiceContextOption
-	verificationServiceFunc    func(opts ...verificationservice.VerificationServiceOption) service.VerificationService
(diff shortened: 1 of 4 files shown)


pkg/application/service/factory/service_factory.go
```diff
@@ -88,15 +81,6 @@ func NewServiceFactory(options ...Option) *ServiceFactory {
...
 
-	// default function to return an instance of Verification service
-	f.verificationServiceFunc = func(_ ...verificationservice.VerificationServiceOption) service.VerificationService {
-		return verificationservice.NewVerificationService(f.getContext(), f.verificationServiceOptions...)
-	}
-
-	f.signupServiceFunc = func(_ ...signupservice.SignupServiceOption) service.SignupService {
-		return signupservice.NewSignupService(f.getContext().Client(), f.signupServiceOptions...)
-	}
-
```
MatousJobanek, line 99: this was the only place where `verificationServiceFunc` and `signupServiceFunc` were used. Also making ready for additional cleanup

alexeykazakov (approved):

PR #523, opened 2025-03-27
chore(build): migrate to golangci-lint@v2 and golangci-lint-action@v7

xcoulon:
upgraded `.golangci.yml` with `golangci-lint migrate`

include fixes suggested by `golangci-lint run -c ./.golangci.yml`


MatousJobanek (approved):

pkg/middleware/promhttp_middleware_test.go
```diff
@@ -100,7 +100,7 @@ func assertMetricExists(t *testing.T, data []byte, name string, labels map[strin
 			return
 		}
 	}
-	assert.Fail(t, "unable to find metric '%s' with labels '%v'", name, labels)
+	assert.Fail(t, "unable to find metric with labels", "name", name, "labels", labels)
```
metlos, line 103: the last arg to the `assert.*` functions is called `msgAndArgs` and is actually printf-formatted (see https://github.com/stretchr/testify/blob/master/assert/assertions.go#L316)

```suggestion
	assert.Fail(t, "unable to find metric with labels", "name: %s, labels: %v", name, labels)
```

metlos (approved):

pkg/middleware/promhttp_middleware_test.go
xcoulon: ohhh, that's not the same logic as the controller logger! 🤦‍♂️

rajivnathan (approved):

PR #524, opened 2025-03-31
update labstack to avoid the vulnerability in jwt

metlos:
This fixes the following issue reported by `govulncheck`:

```
Vulnerability #1: GO-2025-3553
    Excessive memory allocation during header parsing in
    github.com/golang-jwt/jwt
  More info: https://pkg.go.dev/vuln/GO-2025-3553
  Module: github.com/golang-jwt/jwt
    Found in: github.com/golang-jwt/jwt@v3.2.2+incompatible
    Fixed in: N/A
    Example traces found:
      #1: pkg/proxy/proxy.go:35:2: proxy.init calls middleware.init, which calls jwt.init
```


alexeykazakov (approved):

pkg/proxy/metrics_server_test.go
```diff
@@ -54,14 +54,13 @@ func TestProxyMetricsServer(t *testing.T) {
 	require.NoError(t, err)
 	require.NotNil(t, resp)
 	require.Equal(t, http.StatusOK, resp.StatusCode)
-	require.Equal(t, "text/plain; version=0.0.4; charset=utf-8", resp.Header.Get("Content-Type"))
```
MatousJobanek, line 57: was this change intentional? it apparently causes unit test failures

rajivnathan (approved):

pkg/proxy/metrics_server_test.go
metlos: Yeah, it was intentional because I was getting the test errors without it locally.. 😕

PR #532, opened 2025-05-14
fix: make ui config unsecured route

mfrancisc:


mfrancisc:
Closing as not required. 

The secured routes only require SSO token and the user already has that when it opens Dev Sandbox from consoledot

PR #534, opened 2025-06-10
Update verification-code entrypoint

mfrancisc:
Modify the `/api/v1/signup/verification/activation-code` endpoint so that if usersignup does not exists it creates it.

Jira: https://issues.redhat.com/browse/SANDBOX-1209

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/1158

### DIFF
--- a/pkg/controller/signup_test.go
+++ b/pkg/controller/signup_test.go
@@ -28,6 +28,7 @@ import (
 	testsocialevent "github.com/codeready-toolchain/toolchain-common/pkg/test/socialevent"
 	testusersignup "github.com/codeready-toolchain/toolchain-common/pkg/test/usersignup"
 	"github.com/codeready-toolchain/toolchain-common/pkg/usersignup"
+	signupcommon "github.com/codeready-toolchain/toolchain-common/pkg/usersignup"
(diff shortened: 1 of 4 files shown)


alexeykazakov (commented):
Thanks!

Can we also add e2e tests for this too?
- There is no usersignup
- There is a usersignup pending verification (I guess we already have this test)
- There is a usersingup which is deactivated

mfrancisc:
> There is a usersignup pending verification (I guess we already have this test)

@alexeykazakov  it looks like the parallel e2e test suite runs without automatic approval and we approve "manually". See: https://github.com/codeready-toolchain/toolchain-e2e/blob/c63fa0f08e841d589abaef3e3268736766a466ea/test/e2e/parallel/registration_service_test.go#L719-L732

Thus unless we move the test elsewhere I'm not sure how we can test this in e2e.

alexeykazakov:
> Thus unless we move the test elsewhere I'm not sure how we can test this in e2e.

Yeah... since the approval strategy is a global configuration we can't turn it on and off just for specific tests (because we run tests in parallel). So let's skip this e2e test case then. It's not an easy task.

PR #536, opened 2025-06-16
fix: set user unready when no default ns

mfrancisc:
Fix the GET /signup api so that it returns `signup.status.ready=false` when defaultNamespace for the user is unset.

see: https://redhat-internal.slack.com/archives/CHK0J6HT6/p1749804989753299

related PR's:
- https://github.com/codeready-toolchain/api/pull/481

Jira: https://issues.redhat.com/browse/SANDBOX-1214

### DIFF
--- a/pkg/signup/service/signup_service.go
+++ b/pkg/signup/service/signup_service.go
@@ -442,6 +442,16 @@ func (s *ServiceImpl) DoGetSignup(ctx *gin.Context, cl namespaced.Client, userna
 		signupResponse.DefaultUserNamespace = defaultNamespace
 	}
 
+	if defaultNamespace == "" {
+		// default namespace is not set, so we need to set the status to not ready
+		signupResponse.Status = signup.Status{
+			Ready:                false,
+			Reason:               toolchainv1alpha1.UserSignupNoDefaultNamespaceReason,
+			Message:              "No default namespace found",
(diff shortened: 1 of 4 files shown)


pkg/signup/service/signup_service.go
```diff
@@ -442,6 +442,16 @@ func (s *ServiceImpl) DoGetSignup(ctx *gin.Context, cl namespaced.Client, userna
 		signupResponse.DefaultUserNamespace = defaultNamespace
 	}
 
+	if defaultNamespace == "" {
+		// default namespace is not set, so we need to set the status to not ready
+		signupResponse.Status = signup.Status{
+			Ready:                false,
+			Reason:               toolchainv1alpha1.UserSignupNoDefaultNamespaceReason,
+			Message:              "No default namespace found",
```
alexeykazakov, line 450: WDYT about this message instead?
```suggestion
			Message:              "Default namespace not available (yet)",
```

alexeykazakov (approved):
Still not sure how this can happen but it won't hurt to have this check anyway.

mfrancisc:
Let's close this for now as discussed here https://redhat-internal.slack.com/archives/CHK0J6HT6/p1749804989753299 it will be addressed differently as part of another effort.

PR #543, opened 2025-08-20
feat: add route for getting sandbox segment key

mfrancisc:
see: https://redhat-internal.slack.com/archives/CHK0J6HT6/p1755618057521369

This is needed in order to send events from our new UI at https://sandbox.redhat.com/ to Amplitude.

Jira: https://issues.redhat.com/browse/SANDBOX-1394

e2e PR:  https://github.com/codeready-toolchain/toolchain-e2e/pull/1183

 

## Summary by CodeRabbit

* **New Features**
  * Added an unsecured API endpoint GET /api/v1/analytics/segment-write-key to retrieve the sandbox analytics Segment write key (mirrors the existing devspaces key endpoint).

* **Tests**
  * Added tests verifying the new endpoint returns HTTP 200 and the configured sandbox key.

### DIFF
--- a/pkg/server/routes.go
+++ b/pkg/server/routes.go
@@ -69,8 +69,9 @@ func (srv *RegistrationServer) SetupRoutes(proxyPort string, reg *prometheus.Reg
 			middleware.InstrumentRoundTripperDuration(histVec))
 		unsecuredV1.GET("/health", healthCheckCtrl.GetHandler) // TODO: move to root (`/`)?
 		unsecuredV1.GET("/authconfig", authConfigCtrl.GetHandler)
-		unsecuredV1.GET("/segment-write-key", analyticsCtrl.GetDevSpacesSegmentWriteKey) //expose the devspaces segment key
-		unsecuredV1.GET("/analytics/segment-write-key", analyticsCtrl.GetSandboxSegmentWriteKey)
+		// segment keys endpoints
+		unsecuredV1.GET("/segment-write-key", analyticsCtrl.GetDevSpacesSegmentWriteKey)         // expose the devspaces segment key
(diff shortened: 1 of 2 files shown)


pkg/server/routes.go
```diff
@@ -70,6 +70,7 @@ func (srv *RegistrationServer) SetupRoutes(proxyPort string, reg *prometheus.Reg
 		unsecuredV1.GET("/health", healthCheckCtrl.GetHandler) // TODO: move to root (`/`)?
 		unsecuredV1.GET("/authconfig", authConfigCtrl.GetHandler)
 		unsecuredV1.GET("/segment-write-key", analyticsCtrl.GetDevSpacesSegmentWriteKey) //expose the devspaces segment key
+		unsecuredV1.GET("/analytics/segment-write-key", analyticsCtrl.GetSandboxSegmentWriteKey)
```
metlos, line 73: While the change itself is fine, I wonder how difficult would it be to change the devspaces segment key endpoint to something like `/devspaces/segment-write-key`? Because the placing of something specific to devspaces at the root while something relating to the sandbox itself at a lower level seems a bit "surprising". Not a blocker though.

metlos (approved):
The change itself looks fine, just please maybe explain in a comment the logic behind the URL hierarchy for the 2 segment keys we're now exposing.

pkg/server/routes.go
alexeykazakov: It's not difficult to change on our side. But it will break devspaces. We would need to collaborate with them on it and do a migration. Let's leave it out of scope for this change.

alexeykazakov (approved):

mfrancisc:
/retest 

potential flaky test is being discussed here https://redhat-internal.slack.com/archives/C09430LK7QE/p1756203780905619

PR #550, opened 2025-09-10
SANDBOX-1392: Drop dependency on Che instance

fbm3307:
This  PR is to have the  che-url  hardcoded similar to  AI dasboard  url 

the paired PR
- toolchain-e2e - https://github.com/codeready-toolchain/toolchain-e2e/pull/1198

the related PR
- Member-Operator -   https://github.com/codeready-toolchain/member-operator/pull/701
- Toolchain-e2e - https://github.com/codeready-toolchain/toolchain-e2e/pull/1197

Assisted by: Cursor, claude code 

## Summary by CodeRabbit

* **New Features**
  * Signup status now provides a DevSpaces-based Che Dashboard URL derived from the Console URL, giving a consistent HTTPS link across clusters and populated whenever available.
* **Bug Fixes**
  * Prevents missing or cluster-specific Che Dashboard URLs by deriving a canonical DevSpaces URL.
* **Tests**
  * Updated expectations to use DevSpaces URLs and adjusted related error message wording.


pkg/signup/service/signup_service.go
```diff
@@ -450,6 +449,9 @@ func (s *ServiceImpl) DoGetSignup(ctx *gin.Context, cl namespaced.Client, userna
 		// set RHODS member URL
 		signupResponse.RHODSMemberURL = getRHODSMemberURL(*signupResponse)
 
+		// set Che Dashboard URL
+		signupResponse.CheDashboardURL = getAppsURL("devspaces", *signupResponse)
```
metlos, line 453: Just out of my curiosity - this set of PRs is about removing Che from our member clusters but here we seem to be replacing it with a devspaces link, also within the cluster.

Is this about removing Che from the data model and just adding the links solely in the signup response?

fbm3307: this is about removing/using member status to fetch the `che url ` and hardcode it similar to AI url so that the dependency on the running instance of `che` is removed

mfrancisc (approved):
Looks good 👍

pkg/signup/service/signup_service.go
MatousJobanek: just to be clear, we are not removing Che (DevSpaces) from the member clusters. We are only removing the check of the Che route, and the configuration from the ToolchainConfig/MemberConfig CRs so the operators don't depend on theirs existence anyhow. It complicated a bit the configuration part of the new member clusters.

However, we still need to provide the link to the DevSpaces dashboard in the UI - that's why there is this code that hard-codes the DevSpaces link. Theoretically, we could have this configurable (see https://issues.redhat.com/browse/SANDBOX-1393), but it's not really a priority for now - hardcoding is completely sufficient

rsoaresd (approved):

PR #594, opened 2026-05-21
feat: signup service acts based on the response from account verifier

mfrancisc:
When AccountVerifierEnabled is true, the signup flow now checks the account-verifier response and acts accordingly. The result and reasons are stored as annotations on the UserSignup. 

Jira: https://redhat.atlassian.net/browse/SANDBOX-1848

## Summary by CodeRabbit

* **New Features**
  * Account verification on signup records verifier results and reasons; outcomes can approve, reject (blocks signup), or require phone verification.
  * Configurable verifier mode exposed with default "log".

* **Bug Fixes**
  * Verifier errors are treated permissively (fails‑open) to avoid blocking signups.
  * Rejected signups (including reactivations) are consistently blocked from re-signup.

* **Tests**
  * Added comprehensive tests for verifier outcomes, modes, reactivation, and error handling.

### DIFF
--- a/pkg/signup/service/signup_service.go
+++ b/pkg/signup/service/signup_service.go
@@ -154,6 +154,9 @@ func (s *ServiceImpl) newUserSignup(ctx *gin.Context, accountVerifierResp *accou
 		if accountVerifierResp.Result == accountVerifierResultPhoneVerification {
 			states.SetVerificationRequired(userSignup, true)
 		}
+		if accountVerifierResp.Result == accountVerifierResultReject {
+			states.SetRejected(userSignup, true)
+		}
 	}
 
 	// set the skip-auto-create-space annotation to true if the no-space query parameter was set to true
@@ -354,26 +357,42 @@ func (s *ServiceImpl) Signup(ctx *gin.Context) (*toolchainv1alpha1.UserSignup, e
 	userSignup := &toolchainv1alpha1.UserSignup{}
 	if err := s.Get(ctx, s.NamespacedName(encodedUsername), userSignup); err != nil {
 		if apierrors.IsNotFound(err) {
-			// New Signup
+			// New Signup — always create the UserSignup so that a rejected user is recorded in K8s
+			// (enables metrics and prevents repeated verifier calls on retries).
 			log.WithValues(map[string]interface{}{"encoded_username": encodedUsername}).Info(ctx, "user not found, creating a new one")
 			accountVerifierResp := s.verifyAccount(ctx)
+			userSignup, err := s.createUserSignup(ctx, accountVerifierResp)
+			if err != nil {
+				return nil, err
+			}
 			if isAccountVerifierRejected(accountVerifierResp) {
 				return nil, ForbiddenBannedError
 			}
-			return s.createUserSignup(ctx, accountVerifierResp)
+			return userSignup, nil
 		}
 		return nil, err
+	}
+
+	// If the existing UserSignup was rejected by the account verifier, block the user immediately
+	// without calling the verifier again.
+	if states.Rejected(userSignup) {
+		return nil, ForbiddenBannedError
 	}
 
 	// Check UserSignup status to determine whether user signup is deactivated
 	signupCondition, found := condition.FindConditionByType(userSignup.Status.Conditions, toolchainv1alpha1.UserSignupComplete)
 	if found && signupCondition.Status == apiv1.ConditionTrue && signupCondition.Reason == toolchainv1alpha1.UserSignupUserDeactivatedReason {
-		// Signup is deactivated. We need to reactivate it
+		// Signup is deactivated. We need to reactivate it — always update the UserSignup so
+		// that a rejected reactivation is also recorded.
 		accountVerifierResp := s.verifyAccount(ctx)
+		userSignup, err := s.reactivateUserSignup(ctx, userSignup, accountVerifierResp)
+		if err != nil {
+			return nil, err
+		}
 		if isAccountVerifierRejected(accountVerifierResp) {
 			return nil, ForbiddenBannedError
 		}
-		return s.reactivateUserSignup(ctx, userSignup, accountVerifierResp)
+		return userSignup, nil
 	}
 
 	return nil, apierrors.NewConflict(schema.GroupResource{}, "", fmt.Errorf(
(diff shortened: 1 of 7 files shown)


alexeykazakov (approved):

pkg/signup/service/signup_service.go
```diff
@@ -303,9 +318,23 @@ func (s *ServiceImpl) verifyAccount(ctx *gin.Context) {
 	if email == "" {
 		return
 	}
-	if err := callAccountVerifier(ctx, verifierURL, email); err != nil {
+	resp, err := callAccountVerifier(ctx, verifierURL, email)
+	if err != nil {
 		log.Error(ctx, err, "account verifier call failed")
+		return
 	}
+	if cfg.AccountVerifierEnabled() && resp != nil {
```
metlos: Shouldn't we just bail out if account verifier is not enabled at the very beginning of this method?

mfrancisc: that is also an option. Or we could add yet another flag to decide what todo :D 

In case there's an issue with the account verifier I thought that we could disable the acting but still keep calling it so it's useful for debugging ( in case of issues ) or collecting data from new checks before enabling it again.

metlos: what if we did want to switch the account verifier off because of the problems?

pkg/signup/service/signup_service.go
```diff
@@ -321,6 +350,9 @@ func (s *ServiceImpl) Signup(ctx *gin.Context) (*toolchainv1alpha1.UserSignup, e
 			// New Signup
 			log.WithValues(map[string]interface{}{"encoded_username": encodedUsername}).Info(ctx, "user not found, creating a new one")
 			s.verifyAccount(ctx)
+			if isAccountVerifierRejected(ctx) {
+				return nil, ForbiddenBannedError
+			}
 			return s.createUserSignup(ctx)
```
MatousJobanek: I thought that we would still create UserSignup, so we can:
* take a look at the annotation to see why it was rejected
* manually approve the UserSignup in case we would need to bypass the rejection for the specific account

all these rejected UserSignups would get cleaned up after 7 days anyway, and we can have another metric in the dashboard for these abandoned UserSignups with the rejected state.

pkg/signup/service/signup_service.go
```diff
@@ -141,6 +141,23 @@ func (s *ServiceImpl) newUserSignup(ctx *gin.Context) (*toolchainv1alpha1.UserSi
 
 	states.SetVerificationRequired(userSignup, verificationRequired)
 
+	if val, exists := ctx.Get(context.AccountVerifierResult); exists {
```
MatousJobanek: I'm wondering why not calling the `verifyAccount` directly from the `newUserSignup` method and let the `verifyAccount` return the `accountVerifierResponse` - we wouldn't need to deal with setting, getting, and casting the context info, but I'm maybe missing some small detail there

MatousJobanek (approved):
Looks good, but I would still create the UserSignup even when the user is rejected. But I'm fine with keeping it as it is for now.

pkg/signup/service/signup_service.go
mfrancisc: just set `AccountVerifierEnabled` to false in the toolchainconfig , or remove the `AccountVerifierURL` from config so it's not even being called anymore.

pkg/signup/service/signup_service.go
mfrancisc: TBH I don't see the value of creating additional CRs if the users are rejected, and the rejected metric is already stored in the verification_logs schema in the account verifier DB. But if we think there's any usecase for that we can create them.

pkg/signup/service/signup_service.go
mfrancisc: makes sense! pls see https://github.com/codeready-toolchain/registration-service/pull/594/changes/3815145c25c8318d1a8081613d7b54a6055ade73

pkg/signup/service/signup_service.go
metlos: sorry, but this is a hill i'm will to die on :) So I have a setting called 'accountVerifiedEnabled' that, when set to false, will cause the system to still send out requests to the account verifier,
but will just not return any response to the user. What kind of debugging in production do you have in mind where this is needed? how is this following 'the principle of the least astonishment'?
can you provide an example where such multi-level switch (enabled makes the functionality (un)usable by the caller but still invoked, and url which is the 'true' on/off switch) is used?
this seems to me like such a niche usecase that it either should be called something completely descriptive like 'dontSendAccountVerifierResponsesToCallers' or just not be there at all.

pkg/signup/service/signup_service.go
MatousJobanek: as mentioned above, another usecase is to manually approve a specific user that was rejected. I know that we could achieve the same thing with domain override.

as for the metrics, the difference is that abandoned UserSignup gives us granularity at the level of UserSignups - how many user signups were rejected. Account-verifier metrics gives us how many requests were rejected. If the UserSignup is not created, then user can try to sign up multiple times, since there is no record that the user was rejected before, the reg-service would always call account-verifier again.

metlos (approved):

pkg/signup/service/signup_service.go
mfrancisc: Thanks for the details, it does make sense, if the additional CRs in the cluster are not a concern .

I've implemented the `rejected` usersignup state, see:
- https://github.com/codeready-toolchain/registration-service/pull/594/commits/327987320e9db4ad5e87b8367d57afd3de9df07e

and related PRs:
- https://github.com/codeready-toolchain/api/pull/508
- https://github.com/codeready-toolchain/host-operator/pull/1267
- https://github.com/codeready-toolchain/toolchain-common/pull/531

pkg/signup/service/signup_service.go
```diff
@@ -318,29 +357,51 @@ func (s *ServiceImpl) Signup(ctx *gin.Context) (*toolchainv1alpha1.UserSignup, e
...
 			log.WithValues(map[string]interface{}{"encoded_username": encodedUsername}).Info(ctx, "user not found, creating a new one")
-			s.verifyAccount(ctx)
-			return s.createUserSignup(ctx)
+			accountVerifierResp := s.verifyAccount(ctx)
+			userSignup, err := s.createUserSignup(ctx, accountVerifierResp)
+			if err != nil {
+				return nil, err
+			}
+			if isAccountVerifierRejected(accountVerifierResp) {
+				return nil, ForbiddenBannedError
```
rajivnathan: The ForbiddenBannedError will have the text "Access to the Developer Sandbox has been suspended due to suspicious activity or detected abuse." but I guess a different error would be more appropriate here because the account is being rejected before any activity?

rajivnathan (approved):
Looks good! 👍  Not sure if it was mentioned somewhere already but should we have new e2e tests for the following cases:
1. UserSignup rejection case
2. Approving a rejected UserSignup case when a signup is mistakenly rejected?

pkg/signup/service/signup_service.go
mfrancisc: makes sense. Changed in https://github.com/codeready-toolchain/registration-service/pull/594/changes/c81f843ef917ba0f0cc49d318222b20c59e06c2c

PR #595, opened 2026-06-04
feat: add Twilio Lookup v2 API

fbm3307:
## Summary
- Add Twilio Lookup v2 phone  check integration
- Introduce tri-state `PhoneLookupMode` config (disabled/log/enabled) for gradual rollout
- Store lookup results as UserSignup annotations

Depends on:
- https://github.com/codeready-toolchain/api/pull/487
- https://github.com/codeready-toolchain/toolchain-common/pull/597
- https://github.com/codeready-toolchain/toolchain-e2e/pull/1285 
- https://github.com/codeready-toolchain/host-operator/pull/1268

Assisted by: Cursor

## Summary by CodeRabbit

* **New Features**
  * Twilio phone lookup added to assess number risk (carrier, line type, country, block status, risk score).
  * New configuration accessor for phone-lookup mode (default: log).

* **Behavior Changes**
  * Modes: enabled (block), log (warn/fail-open), disabled.
  * Lookups skipped for US numbers and certain retry/unchanged-number cases; results recorded on signups; high-risk numbers can be rejected when enabled.

* **Tests**
  * Expanded coverage for parsing, edge cases, error handling, modes, retries, and annotation updates.

### DIFF
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -331,3 +331,7 @@ func (r VerificationConfig) PhoneLookupMode() toolchainv1alpha1.PhoneLookupMode 
 	}
 	return toolchainv1alpha1.PhoneLookupModeLog
 }
+
+func (r VerificationConfig) PhoneLookupExcludedCountries() []string {
+	return r.c.PhoneLookupExcludedCountries
+}
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -324,3 +324,7 @@ func (r VerificationConfig) CaptchaServiceAccountFileContents() string {
 	content := r.registrationServiceSecret(key)
 	return string(content)
 }
+
+func (r VerificationConfig) PhoneLookupMode() string {
+	return commonconfig.GetString(r.c.PhoneLookupMode, "log")
+}
--- a/pkg/configuration/configuration.go
+++ b/pkg/configuration/configuration.go
@@ -325,6 +325,9 @@ func (r VerificationConfig) CaptchaServiceAccountFileContents() string {
 	return string(content)
 }
 
-func (r VerificationConfig) PhoneLookupMode() string {
-	return commonconfig.GetString(r.c.PhoneLookupMode, "log")
+func (r VerificationConfig) PhoneLookupMode() toolchainv1alpha1.PhoneLookupMode {
+	if r.c.PhoneLookupMode != nil {
+		return *r.c.PhoneLookupMode
+	}
+	return toolchainv1alpha1.PhoneLookupModeLog
 }
--- a/pkg/configuration/configuration_test.go
+++ b/pkg/configuration/configuration_test.go
@@ -3,7 +3,7 @@ package configuration_test
 import (
 	"testing"
 
-	"github.com/codeready-toolchain/api/api/v1alpha1"
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 	"github.com/codeready-toolchain/registration-service/pkg/configuration"
 	"github.com/codeready-toolchain/registration-service/test"
 	commonconfig "github.com/codeready-toolchain/toolchain-common/pkg/configuration"
@@ -69,7 +69,7 @@ func TestRegistrationService(t *testing.T) {
 		assert.InDelta(t, float32(0), regServiceCfg.Verification().CaptchaRequiredScore(), 0.01)
 		assert.True(t, regServiceCfg.Verification().CaptchaAllowLowScoreReactivation())
 		assert.Empty(t, regServiceCfg.Verification().CaptchaServiceAccountFileContents())
-		assert.Equal(t, "log", regServiceCfg.Verification().PhoneLookupMode())
+		assert.Equal(t, toolchainv1alpha1.PhoneLookupModeLog, regServiceCfg.Verification().PhoneLookupMode())
 		assert.False(t, regServiceCfg.PublicViewerEnabled())
 		assert.Empty(t, regServiceCfg.AccountVerifierURL())
 	})
@@ -101,7 +101,7 @@ func TestRegistrationService(t *testing.T) {
 			Verification().CaptchaScoreThreshold("0.7").
 			Verification().CaptchaRequiredScore("0.5").
 			Verification().CaptchaAllowLowScoreReactivation(false).
-			Verification().PhoneLookupMode("enabled").
+			Verification().PhoneLookupMode(toolchainv1alpha1.PhoneLookupModeEnabled).
 			Verification().Secret().Ref("verification-secrets").
 			TwilioAccountSID("twilio.sid").
 			TwilioAuthToken("twilio.token").
@@ -159,7 +159,7 @@ func TestRegistrationService(t *testing.T) {
 		assert.InDelta(t, float32(0.5), regServiceCfg.Verification().CaptchaRequiredScore(), 0.01)
 		assert.False(t, regServiceCfg.Verification().CaptchaAllowLowScoreReactivation())
 		assert.Equal(t, "example-content", regServiceCfg.Verification().CaptchaServiceAccountFileContents())
-		assert.Equal(t, "enabled", regServiceCfg.Verification().PhoneLookupMode())
+		assert.Equal(t, toolchainv1alpha1.PhoneLookupModeEnabled, regServiceCfg.Verification().PhoneLookupMode())
 		assert.False(t, regServiceCfg.PublicViewerEnabled())
 		assert.Equal(t, "https://verifier.example.com", regServiceCfg.AccountVerifierURL())
 	})
@@ -168,15 +168,15 @@ func TestRegistrationService(t *testing.T) {
 func TestPhoneLookupMode(t *testing.T) {
 	t.Run("returns configured value", func(t *testing.T) {
 		cfg := commonconfig.NewToolchainConfigObjWithReset(t, testconfig.RegistrationService().
-			Verification().PhoneLookupMode("disabled"))
+			Verification().PhoneLookupMode(toolchainv1alpha1.PhoneLookupModeDisabled))
 		regServiceCfg := configuration.NewRegistrationServiceConfig(cfg, map[string]map[string]string{})
-		assert.Equal(t, "disabled", regServiceCfg.Verification().PhoneLookupMode())
+		assert.Equal(t, toolchainv1alpha1.PhoneLookupModeDisabled, regServiceCfg.Verification().PhoneLookupMode())
 	})
 
 	t.Run("defaults to log when unset", func(t *testing.T) {
 		cfg := commonconfig.NewToolchainConfigObjWithReset(t)
 		regServiceCfg := configuration.NewRegistrationServiceConfig(cfg, map[string]map[string]string{})
-		assert.Equal(t, "log", regServiceCfg.Verification().PhoneLookupMode())
+		assert.Equal(t, toolchainv1alpha1.PhoneLookupModeLog, regServiceCfg.Verification().PhoneLookupMode())
 	})
 }
 
@@ -184,15 +184,15 @@ func TestPublicViewerConfiguration(t *testing.T) {
 	tt := map[string]struct {
 		name               string
 		expectedValue      bool
-		publicViewerConfig *v1alpha1.PublicViewerConfiguration
+		publicViewerConfig *toolchainv1alpha1.PublicViewerConfiguration
 	}{
 		"public-viewer is explicitly enabled": {
 			expectedValue:      true,
-			publicViewerConfig: &v1alpha1.PublicViewerConfiguration{Enabled: true},
+			publicViewerConfig: &toolchainv1alpha1.PublicViewerConfiguration{Enabled: true},
 		},
 		"public-viewer is explicitly disabled": {
 			expectedValue:      false,
-			publicViewerConfig: &v1alpha1.PublicViewerConfiguration{Enabled: false},
+			publicViewerConfig: &toolchainv1alpha1.PublicViewerConfiguration{Enabled: false},
 		},
 		"public-viewer config not set, assume disabled": {
 			expectedValue:      false,
--- a/pkg/configuration/configuration_test.go
+++ b/pkg/configuration/configuration_test.go
@@ -70,6 +70,7 @@ func TestRegistrationService(t *testing.T) {
 		assert.True(t, regServiceCfg.Verification().CaptchaAllowLowScoreReactivation())
 		assert.Empty(t, regServiceCfg.Verification().CaptchaServiceAccountFileContents())
 		assert.Equal(t, toolchainv1alpha1.PhoneLookupModeLog, regServiceCfg.Verification().PhoneLookupMode())
+		assert.Empty(t, regServiceCfg.Verification().PhoneLookupExcludedCountries())
 		assert.False(t, regServiceCfg.PublicViewerEnabled())
 		assert.Empty(t, regServiceCfg.AccountVerifierURL())
 		assert.Equal(t, "log", regServiceCfg.AccountVerifierMode())
@@ -105,6 +106,7 @@ func TestRegistrationService(t *testing.T) {
 			Verification().CaptchaRequiredScore("0.5").
 			Verification().CaptchaAllowLowScoreReactivation(false).
 			Verification().PhoneLookupMode(toolchainv1alpha1.PhoneLookupModeEnabled).
+			Verification().PhoneLookupExcludedCountries([]string{"US", "CA"}).
 			Verification().Secret().Ref("verification-secrets").
 			TwilioAccountSID("twilio.sid").
 			TwilioAuthToken("twilio.token").
@@ -160,26 +162,13 @@ func TestRegistrationService(t *testing.T) {
 		assert.False(t, regServiceCfg.Verification().CaptchaAllowLowScoreReactivation())
 		assert.Equal(t, "example-content", regServiceCfg.Verification().CaptchaServiceAccountFileContents())
 		assert.Equal(t, toolchainv1alpha1.PhoneLookupModeEnabled, regServiceCfg.Verification().PhoneLookupMode())
+		assert.Equal(t, []string{"US", "CA"}, regServiceCfg.Verification().PhoneLookupExcludedCountries())
 		assert.False(t, regServiceCfg.PublicViewerEnabled())
 		assert.Equal(t, "https://verifier.example.com", regServiceCfg.AccountVerifierURL())
 		assert.Equal(t, "enabled", regServiceCfg.AccountVerifierMode())
 	})
 }
 
-func TestPhoneLookupMode(t *testing.T) {
-	t.Run("returns configured value", func(t *testing.T) {
-		cfg := commonconfig.NewToolchainConfigObjWithReset(t, testconfig.RegistrationService().
-			Verification().PhoneLookupMode(toolchainv1alpha1.PhoneLookupModeDisabled))
-		regServiceCfg := configuration.NewRegistrationServiceConfig(cfg, map[string]map[string]string{})
-		assert.Equal(t, toolchainv1alpha1.PhoneLookupModeDisabled, regServiceCfg.Verification().PhoneLookupMode())
-	})
-
-	t.Run("defaults to log when unset", func(t *testing.T) {
-		cfg := commonconfig.NewToolchainConfigObjWithReset(t)
-		regServiceCfg := configuration.NewRegistrationServiceConfig(cfg, map[string]map[string]string{})
-		assert.Equal(t, toolchainv1alpha1.PhoneLookupModeLog, regServiceCfg.Verification().PhoneLookupMode())
-	})
-}
 
 func TestPublicViewerConfiguration(t *testing.T) {
 	tt := map[string]struct {
--- a/pkg/configuration/configuration_test.go
+++ b/pkg/configuration/configuration_test.go
@@ -69,6 +69,7 @@ func TestRegistrationService(t *testing.T) {
 		assert.InDelta(t, float32(0), regServiceCfg.Verification().CaptchaRequiredScore(), 0.01)
 		assert.True(t, regServiceCfg.Verification().CaptchaAllowLowScoreReactivation())
 		assert.Empty(t, regServiceCfg.Verification().CaptchaServiceAccountFileContents())
+		assert.Equal(t, "log", regServiceCfg.Verification().PhoneLookupMode())
 		assert.False(t, regServiceCfg.PublicViewerEnabled())
 		assert.Empty(t, regServiceCfg.AccountVerifierURL())
 	})
@@ -100,6 +101,7 @@ func TestRegistrationService(t *testing.T) {
 			Verification().CaptchaScoreThreshold("0.7").
 			Verification().CaptchaRequiredScore("0.5").
 			Verification().CaptchaAllowLowScoreReactivation(false).
+			Verification().PhoneLookupMode("enabled").
 			Verification().Secret().Ref("verification-secrets").
 			TwilioAccountSID("twilio.sid").
 			TwilioAuthToken("twilio.token").
@@ -157,8 +159,24 @@ func TestRegistrationService(t *testing.T) {
 		assert.InDelta(t, float32(0.5), regServiceCfg.Verification().CaptchaRequiredScore(), 0.01)
 		assert.False(t, regServiceCfg.Verification().CaptchaAllowLowScoreReactivation())
 		assert.Equal(t, "example-content", regServiceCfg.Verification().CaptchaServiceAccountFileContents())
+		assert.Equal(t, "enabled", regServiceCfg.Verification().PhoneLookupMode())
 		assert.False(t, regServiceCfg.PublicViewerEnabled())
 		assert.Equal(t, "https://verifier.example.com", regServiceCfg.AccountVerifierURL())
+	})
+}
+
+func TestPhoneLookupMode(t *testing.T) {
+	t.Run("returns configured value", func(t *testing.T) {
+		cfg := commonconfig.NewToolchainConfigObjWithReset(t, testconfig.RegistrationService().
+			Verification().PhoneLookupMode("disabled"))
+		regServiceCfg := configuration.NewRegistrationServiceConfig(cfg, map[string]map[string]string{})
+		assert.Equal(t, "disabled", regServiceCfg.Verification().PhoneLookupMode())
+	})
+
+	t.Run("defaults to log when unset", func(t *testing.T) {
+		cfg := commonconfig.NewToolchainConfigObjWithReset(t)
+		regServiceCfg := configuration.NewRegistrationServiceConfig(cfg, map[string]map[string]string{})
+		assert.Equal(t, "log", regServiceCfg.Verification().PhoneLookupMode())
 	})
 }
 
--- a/pkg/configuration/configuration_test.go
+++ b/pkg/configuration/configuration_test.go
@@ -169,7 +169,6 @@ func TestRegistrationService(t *testing.T) {
 	})
 }
 
-
 func TestPublicViewerConfiguration(t *testing.T) {
 	tt := map[string]struct {
 		name               string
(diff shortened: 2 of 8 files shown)


pkg/verification/sender/phone_lookup.go
```diff
@@ -0,0 +1,72 @@
...
+		result.CountryCode = *resp.CountryCode
+	}
+	result.CarrierRiskCategory = resp.SmsPumpingRisk.CarrierRiskCategory
+	result.NumberBlocked = resp.SmsPumpingRisk.NumberBlocked
+	result.RiskScore = resp.SmsPumpingRisk.SmsPumpingRiskScore
+	result.CarrierName = resp.LineTypeIntelligence.CarrierName
+	result.LineType = resp.LineTypeIntelligence.Type
+
+	return result, nil
+}
```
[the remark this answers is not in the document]
metlos: I may be incorrect, but SmsPumpingRisk is not nillable: https://github.com/twilio/twilio-go/blob/main/rest/lookups/v2/model_lookup_response.go#L38. LineTypeIntelligence is likewise non-nillable.

pkg/verification/service/verification_service.go
```diff
@@ -142,27 +155,62 @@ func (s *ServiceImpl) InitVerification(ctx *gin.Context, username, e164PhoneNumb
 		log.Error(ctx, err, fmt.Sprintf("%d attempts made. the daily limit of %d has been exceeded", counter, dailyLimit))
 		initError = crterrors.NewForbiddenError("daily limit exceeded", "cannot generate new verification code")
 	} else {
-		// generate verification code
-		verificationCode, err := generateVerificationCode()
-		if err != nil {
-			return crterrors.NewInternalError(err, "error while generating verification code")
+		mode := cfg.Verification().PhoneLookupMode()
+		if mode != toolchainv1alpha1.PhoneLookupModeDisabled && countryCode != "1" {
```
metlos: For those of us not fully aware of the intricacies of phone country codes, it may be useful to note here that US and CAN share the country code of '1'.

pkg/verification/service/verification_service.go
```diff
@@ -203,6 +251,10 @@ func (s *ServiceImpl) InitVerification(ctx *gin.Context, username, e164PhoneNumb
 	return initError
 }
 
+func isHighRiskPhone(result *sender.PhoneLookupResult) bool {
+	return result.CarrierRiskCategory == "high" || result.NumberBlocked
```
metlos, line 276: Another nitpick: maybe comment here that 'high' is the highest category out of the available 'low', 'mild', 'moderate' and 'high'.

pkg/verification/service/verification_service_test.go
```diff
@@ -872,6 +872,253 @@ func (s *TestVerificationServiceSuite) testVerifyActivationCode(targetCluster st
...
+		JSON(body)
+}
+
+func mockTwilioSMS() {
+	gock.New("https://api.twilio.com").
+		Reply(http.StatusNoContent).
+		BodyString("")
+}
+
+func (s *TestVerificationServiceSuite) TestInitVerificationPhoneLookup() {
```
metlos, line 906: There seems to be a lot of repetition in these tests. Maybe worth simplifying them?

metlos (approved):
I have only a couple of minor comments that don't block merging, if you don't want to address them now. Great job! :rocket:

pkg/verification/service/verification_service.go
fbm3307: ah yes, updated it with comment

pkg/verification/service/verification_service.go
fbm3307: updated it with explnation

pkg/verification/service/verification_service_test.go
fbm3307: simplified with table driven

MatousJobanek: I'm sorry, but the table-driven test is over-complicated approach that is even harder to read.
let's get back to the original version and think about the duplicated code - if it could be moved up in the parent test or extracted into a separate function without too many of parameters, nested structures, if-else staments, etc...

MatousJobanek: And also, keep using `// given // when // then` comments/sections

fbm3307: done

pkg/configuration/configuration.go
```diff
@@ -324,3 +324,14 @@ func (r VerificationConfig) CaptchaServiceAccountFileContents() string {
 	content := r.registrationServiceSecret(key)
 	return string(content)
 }
+
+func (r VerificationConfig) PhoneLookupMode() toolchainv1alpha1.PhoneLookupMode {
+	if r.c.PhoneLookupMode != nil {
+		return *r.c.PhoneLookupMode
+	}
+	return toolchainv1alpha1.PhoneLookupModeLog
```
rajivnathan, line 336: Maybe the default should be disabled since there's an associated cost with enabling it?

pkg/verification/service/verification_service.go
```diff
@@ -203,6 +261,45 @@ func (s *ServiceImpl) InitVerification(ctx *gin.Context, username, e164PhoneNumb
...
+	for _, c := range excluded {
+		if c == regionCode {
+			return true
+		}
+	}
+	return false
+}
+
+// isLookupRejected checks whether the phone lookup details annotation indicates rejection.
+func isLookupRejected(signup *toolchainv1alpha1.UserSignup) bool {
```
rajivnathan: The name has 'lookup' which is a bit generic considering it's specifically checking the phone lookup annotation value so maybe a function name like 'isPhoneNumRejected' would be more clear?

rajivnathan (approved):

pkg/configuration/configuration.go
fbm3307: i did think about that, but currently i thought we are working in a time crunch for sorting out UK number bots, its better to start with log, else if we default to disable, we would then deploy it with log mod and then with enabled, so i just removed the extra step for now, and also data collection starts in log mod .
 if in future we want it disabled its just a configuration change

pkg/verification/service/verification_service.go
```diff
@@ -83,6 +94,10 @@ func (s *ServiceImpl) InitVerification(ctx *gin.Context, username, e164PhoneNumb
 		return crterrors.NewBadRequest("forbidden request", "verification code will not be sent")
 	}
 
+	if isLookupRejected(signup) {
+		return crterrors.NewForbiddenError("phone verification rejected", "cannot proceed with verification")
+	}
```
MatousJobanek: if the UserSignup was already rejected before, then it won't reach this line anymore - see the if-statement above.
Also, we should not differentiate between rejected based on account-verifier response and Lookup api response

pkg/verification/sender/phone_lookup.go
```diff
@@ -0,0 +1,72 @@
...
+		Username: accountSID,
+		Password: authToken,
+		Client:   baseClient,
+	})
+	return &TwilioPhoneLookup{client: client}
+}
+
+// LookupPhone fetches sms_pumping_risk and line_type_intelligence for the given E.164 number.
+func (t *TwilioPhoneLookup) LookupPhone(ctx context.Context, phoneNumber string) (*PhoneLookupResult, error) {
+	_ = ctx
```
MatousJobanek: Can you please clarify why the `LookupPhone` requires the context to be provided as a parameter when it's immediately discarded and not used at all?

pkg/verification/sender/phone_lookup_test.go
```diff
@@ -0,0 +1,139 @@
+package sender_test
```
MatousJobanek, line 1: as already mentioned in the other test, can you please include `// given // when // then` comments?

pkg/verification/service/verification_service.go
```diff
@@ -203,6 +261,45 @@ func (s *ServiceImpl) InitVerification(ctx *gin.Context, username, e164PhoneNumb
...
+	return result.CarrierRiskCategory == "high" || result.NumberBlocked
+}
+
+// regionCodeFromE164 parses an E.164 phone number and returns its ISO 3166-1 alpha-2 region code.
+// Returns empty string if parsing fails.
+func regionCodeFromE164(number string) string {
+	parsed, err := phonenumbers.Parse(number, "")
+	if err != nil {
+		return ""
+	}
```
MatousJobanek: let's log the error here and not silently ignore it

pkg/verification/service/verification_service.go
```diff
@@ -203,6 +261,45 @@ func (s *ServiceImpl) InitVerification(ctx *gin.Context, username, e164PhoneNumb
 	return initError
 }
 
+// isHighRiskPhone returns true when Twilio Lookup reports the highest risk category ("high";
+// available categories are low, mild, moderate, high) or when the number is blocked.
+func isHighRiskPhone(result *sender.PhoneLookupResult) bool {
+	return result.CarrierRiskCategory == "high" || result.NumberBlocked
+}
+
+// regionCodeFromE164 parses an E.164 phone number and returns its ISO 3166-1 alpha-2 region code.
```
MatousJobanek: only a naming nitpick - is ISO 3166-1 alpha-2 a "region code", or "country code"?  Just to be aligned as in API https://github.com/codeready-toolchain/api/pull/509/changes#diff-d7ecdfa67046592cbc7578f4e1af8a7a08016463e6e08192d548e0f5e46c4f60R407-R411

pkg/verification/service/verification_service.go
```diff
@@ -203,6 +261,45 @@ func (s *ServiceImpl) InitVerification(ctx *gin.Context, username, e164PhoneNumb
...
+}
+
+// isCountryExcluded returns true if the given region code is in the excluded list.
+func isCountryExcluded(regionCode string, excluded []string) bool {
+	for _, c := range excluded {
+		if c == regionCode {
+			return true
+		}
+	}
+	return false
```
MatousJobanek:
```suggestion
	return slices.Contains(excluded, regionCode)
```

pkg/verification/service/verification_service.go
```diff
@@ -142,27 +157,70 @@ func (s *ServiceImpl) InitVerification(ctx *gin.Context, username, e164PhoneNumb
...
+					annotationValues[toolchainv1alpha1.UserSignupPhoneLookupNumberBlockedAnnotationKey] = strconv.FormatBool(result.NumberBlocked)
+					details := map[string]string{
+						"risk_score":   strconv.Itoa(result.RiskScore),
+						"carrier_name": result.CarrierName,
+						"line_type":    result.LineType,
+					}
+					if isHighRiskPhone(result) {
+						details["result"] = "rejected"
+						if mode == toolchainv1alpha1.PhoneLookupModeEnabled {
+							initError = crterrors.NewForbiddenError("phone verification rejected", "cannot proceed with verification")
```
MatousJobanek: The UserSignup should be set to rejected, we don't need to store the result in the JSON and then parse it. Or am I missing some small detail?

pkg/verification/service/verification_service.go
```diff
@@ -142,27 +157,70 @@ func (s *ServiceImpl) InitVerification(ctx *gin.Context, username, e164PhoneNumb
...
+				if lookupErr != nil {
+					log.Error(ctx, lookupErr, "phone lookup failed, proceeding (fail-open)")
+				} else {
+					annotationValues[toolchainv1alpha1.UserSignupPhoneLookupCarrierRiskAnnotationKey] = result.CarrierRiskCategory
+					annotationValues[toolchainv1alpha1.UserSignupPhoneLookupNumberBlockedAnnotationKey] = strconv.FormatBool(result.NumberBlocked)
+					details := map[string]string{
+						"risk_score":   strconv.Itoa(result.RiskScore),
+						"carrier_name": result.CarrierName,
+						"line_type":    result.LineType,
+					}
```
MatousJobanek: well, you don't need to map it into a map - you can define it at the level of the type:
```go
type PhoneLookupResult struct {
	PhoneLookupResultDetails
	CarrierRiskCategory string
	NumberBlocked       bool
	CountryCode         string
}

type PhoneLookupResultDetails struct {
	RiskScore   int    `json:"risk_score"`
	CarrierName string `json:"carrier_name"`
	LineType    string `json:"line_type"`
}
```
and then
```go 
detailsJSON, err := json.Marshal(result.PhoneLookupResultDetails)
```

pkg/verification/service/verification_service.go
fbm3307: done

pkg/verification/sender/phone_lookup.go
fbm3307: ctx is accepted for interface consistency with the PhoneLooker interface, so callers can pass request-scoped context. The Twilio Go SDK's FetchPhoneNumber doesn't support per-call context yet, so the parameter is named _ and a comment explains why. When the SDK adds context support, we can wire it through without changing the interface.

pkg/verification/sender/phone_lookup_test.go
fbm3307: yes done

pkg/verification/service/verification_service.go
fbm3307: okay

pkg/verification/service/verification_service.go
fbm3307: done — renamed 
iso country is 2 letter "US"/"CA"

pkg/verification/service/verification_service.go
fbm3307: Done — removed

pkg/verification/service/verification_service.go
fbm3307: Done — introduced PhoneLookupResultDetails struct with json tags, embedded in PhoneLookupResult. Details annotation is now json.Marshal(result.PhoneLookupResultDetails).

pkg/verification/service/verification_service.go
fbm3307: This function (isLookupRejected) has been removed entirely. We now use states.Rejected(signup) from toolchain-common/pkg/states, which checks the rejected state directly on the UserSignup spec — consistent with how the account-verifier already handles rejection.

pkg/verification/sender/phone_lookup.go
MatousJobanek: I'm sorry but I don't follow, the PhoneLooker interface is a type that you added, the param can be removed there too. Apart from that `TwilioPhoneLookup` is the only implementation of the interface - there is no need for having it there at all. 
Anyway, can you explain why do we need to have the context when it's not used and not needed?

fbm3307: you are right we do not need it , i missed the above, i have removed it now

pkg/configuration/configuration_test.go
```diff
@@ -69,6 +69,7 @@ func TestRegistrationService(t *testing.T) {
 		assert.InDelta(t, float32(0), regServiceCfg.Verification().CaptchaRequiredScore(), 0.01)
 		assert.True(t, regServiceCfg.Verification().CaptchaAllowLowScoreReactivation())
 		assert.Empty(t, regServiceCfg.Verification().CaptchaServiceAccountFileContents())
+		assert.Equal(t, toolchainv1alpha1.PhoneLookupModeLog, regServiceCfg.Verification().PhoneLookupMode())
```
MatousJobanek, line 72: I'm missing updates of the test for `PhoneLookupExcludedCountries`

pkg/configuration/configuration_test.go
```diff
@@ -157,25 +159,41 @@ func TestRegistrationService(t *testing.T) {
...
+		assert.Equal(t, toolchainv1alpha1.PhoneLookupModeDisabled, regServiceCfg.Verification().PhoneLookupMode())
+	})
+
+	t.Run("defaults to log when unset", func(t *testing.T) {
+		cfg := commonconfig.NewToolchainConfigObjWithReset(t)
+		regServiceCfg := configuration.NewRegistrationServiceConfig(cfg, map[string]map[string]string{})
+		assert.Equal(t, toolchainv1alpha1.PhoneLookupModeLog, regServiceCfg.Verification().PhoneLookupMode())
+	})
+}
+
```
MatousJobanek: this looks like a duplication of the above, or did I miss something different there?

pkg/verification/sender/phone_lookup_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+import (
+	"net/http"
+	"testing"
+
+	sender2 "github.com/codeready-toolchain/registration-service/pkg/verification/sender"
```
MatousJobanek: nitpick 
```suggestion
	"github.com/codeready-toolchain/registration-service/pkg/verification/sender"
```

pkg/verification/sender/phone_lookup_test.go
```diff
@@ -0,0 +1,163 @@
...
+	const (
+		accountSID = "ACaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+		authToken  = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+		phone      = "+447700900000"
+	)
+
+	setupLookup := func(t *testing.T, status int, body interface{}) sender2.PhoneLooker {
+		t.Helper()
+		httpClient := &http.Client{Transport: &http.Transport{}}
+		gock.InterceptClient(httpClient)
```
MatousJobanek: FYI, this is actually not needed - the gock automatically intercepts the `http.DefaultClient`

pkg/verification/service/verification_service.go
```diff
@@ -203,6 +228,64 @@ func (s *ServiceImpl) InitVerification(ctx *gin.Context, username, e164PhoneNumb
...
+		} else {
+			log.Info(ctx, fmt.Sprintf("high risk phone detected (carrier_risk=%s, blocked=%t), proceeding (phone lookup mode=%s)",
+				result.CarrierRiskCategory, result.NumberBlocked, mode))
+		}
+	}
+	detailsJSON, err := json.Marshal(result.PhoneLookupResultDetails)
+	if err != nil {
+		log.Error(ctx, err, "failed to marshal phone lookup details")
+	}
+	annotationValues[toolchainv1alpha1.UserSignupPhoneLookupDetailsAnnotationKey] = string(detailsJSON)
```
MatousJobanek: nitpick, we don't expect error here, but the annotation should be set only when there was no error

pkg/verification/service/verification_service.go
```diff
@@ -203,6 +228,64 @@ func (s *ServiceImpl) InitVerification(ctx *gin.Context, username, e164PhoneNumb
...
+	return rejectSignup, initError
```
MatousJobanek: a version a bit easier to read:
```suggestion
	detailsJSON, err := json.Marshal(result.PhoneLookupResultDetails)
	if err != nil {
		log.Error(ctx, err, "failed to marshal phone lookup details")
	} else {
		annotationValues[toolchainv1alpha1.UserSignupPhoneLookupDetailsAnnotationKey] = string(detailsJSON)
	}

	if isHighRiskPhone(result) {
		if mode == toolchainv1alpha1.PhoneLookupModeEnabled {
			return true, crterrors.NewForbiddenError("phone verification rejected", "cannot proceed with verification")
		}
		log.Info(ctx, fmt.Sprintf("high risk phone detected (carrier_risk=%s, blocked=%t), proceeding (phone lookup mode=%s)",
			result.CarrierRiskCategory, result.NumberBlocked, mode))
	}

	return false, nil
```

pkg/verification/service/verification_service.go
```diff
@@ -203,6 +228,64 @@ func (s *ServiceImpl) InitVerification(ctx *gin.Context, username, e164PhoneNumb
...
+	annotationValues[toolchainv1alpha1.UserSignupPhoneLookupCarrierRiskAnnotationKey] = result.CarrierRiskCategory
+	annotationValues[toolchainv1alpha1.UserSignupPhoneLookupNumberBlockedAnnotationKey] = strconv.FormatBool(result.NumberBlocked)
+	if isHighRiskPhone(result) {
+		if mode == toolchainv1alpha1.PhoneLookupModeEnabled {
+			initError = crterrors.NewForbiddenError("phone verification rejected", "cannot proceed with verification")
+			rejectSignup = true
+		} else {
+			log.Info(ctx, fmt.Sprintf("high risk phone detected (carrier_risk=%s, blocked=%t), proceeding (phone lookup mode=%s)",
+				result.CarrierRiskCategory, result.NumberBlocked, mode))
+		}
```
MatousJobanek: btw, we want to log it also when the mode is `enabled` see my proposed version in the other comment

pkg/verification/service/verification_service_test.go
```diff
@@ -872,6 +873,288 @@ func (s *TestVerificationServiceSuite) testVerifyActivationCode(targetCluster st
...
+		ctx, _ := gin.CreateTestContext(httptest.NewRecorder())
+		err := application.VerificationService().InitVerification(ctx, "lookup-err@kubesaw", lookupUKPhone, "44")
+
+		// then
+		require.NoError(s.T(), err)
+
+		updated := &toolchainv1alpha1.UserSignup{}
+		require.NoError(s.T(), fakeClient.Get(gocontext.TODO(), client.ObjectKeyFromObject(userSignup), updated))
+		assert.Empty(s.T(), updated.Annotations[toolchainv1alpha1.UserSignupPhoneLookupDetailsAnnotationKey])
+		assert.NotEmpty(s.T(), updated.Annotations[toolchainv1alpha1.UserSignupVerificationCodeAnnotationKey])
```
MatousJobanek, line 1030:
```suggestion
		assert.NotEmpty(s.T(), updated.Annotations[toolchainv1alpha1.UserSignupVerificationCodeAnnotationKey])
		assert.False(s.T(), states.Rejected(updated))
```

pkg/verification/service/verification_service_test.go
```diff
@@ -872,6 +873,288 @@ func (s *TestVerificationServiceSuite) testVerifyActivationCode(targetCluster st
...
+		err := application.VerificationService().InitVerification(ctx, "lookup-us@kubesaw", "+12025551234", "1")
+
+		// then
+		require.NoError(s.T(), err)
+
+		updated := &toolchainv1alpha1.UserSignup{}
+		require.NoError(s.T(), fakeClient.Get(gocontext.TODO(), client.ObjectKeyFromObject(userSignup), updated))
+		assert.Empty(s.T(), updated.Annotations[toolchainv1alpha1.UserSignupPhoneLookupDetailsAnnotationKey])
+		assert.NotEmpty(s.T(), updated.Annotations[toolchainv1alpha1.UserSignupVerificationCodeAnnotationKey])
+		assert.True(s.T(), gock.IsDone())
```
MatousJobanek, line 1057:
```suggestion
		assert.True(s.T(), gock.IsDone())
		assert.False(s.T(), states.Rejected(updated))
```

pkg/verification/service/verification_service_test.go
```diff
@@ -872,6 +873,288 @@ func (s *TestVerificationServiceSuite) testVerifyActivationCode(targetCluster st
...
+		err := application.VerificationService().InitVerification(ctx, "lookup-off@kubesaw", lookupUKPhone, "44")
+
+		// then
+		require.NoError(s.T(), err)
+
+		updated := &toolchainv1alpha1.UserSignup{}
+		require.NoError(s.T(), fakeClient.Get(gocontext.TODO(), client.ObjectKeyFromObject(userSignup), updated))
+		assert.Empty(s.T(), updated.Annotations[toolchainv1alpha1.UserSignupPhoneLookupDetailsAnnotationKey])
+		assert.NotEmpty(s.T(), updated.Annotations[toolchainv1alpha1.UserSignupVerificationCodeAnnotationKey])
+		assert.True(s.T(), gock.IsDone())
```
MatousJobanek, line 1083:
```suggestion
		assert.True(s.T(), gock.IsDone())
		assert.False(s.T(), states.Rejected(updated))
```

pkg/verification/service/verification_service_test.go
```diff
@@ -872,6 +873,288 @@ func (s *TestVerificationServiceSuite) testVerifyActivationCode(targetCluster st
...
+		// when
+		ctx, _ := gin.CreateTestContext(httptest.NewRecorder())
+		err := application.VerificationService().InitVerification(ctx, "lookup-retry@kubesaw", lookupUKPhone, "44")
+
+		// then
+		require.Error(s.T(), err)
+		var e *crterrors.Error
+		require.ErrorAs(s.T(), err, &e)
+		assert.Equal(s.T(), http.StatusForbidden, e.Code)
+		assert.True(s.T(), gock.IsDone())
```
MatousJobanek: gock should not be done, right? the code didn't call the mock in this particular case

pkg/verification/service/verification_service_test.go
```diff
@@ -872,6 +873,288 @@ func (s *TestVerificationServiceSuite) testVerifyActivationCode(targetCluster st
...
+		ctx, _ := gin.CreateTestContext(httptest.NewRecorder())
+		err := application.VerificationService().InitVerification(ctx, "lookup-same@kubesaw", lookupUKPhone, "44")
+
+		// then
+		require.NoError(s.T(), err)
+
+		updated := &toolchainv1alpha1.UserSignup{}
+		require.NoError(s.T(), fakeClient.Get(gocontext.TODO(), client.ObjectKeyFromObject(userSignup), updated))
+		assert.NotEmpty(s.T(), updated.Annotations[toolchainv1alpha1.UserSignupVerificationCodeAnnotationKey])
+		assert.True(s.T(), gock.IsDone())
```
MatousJobanek: the same

MatousJobanek (approved):
Approving the PR since there is no blocker anymore, but please take a look at my comments above

pkg/configuration/configuration_test.go
fbm3307: added that

pkg/configuration/configuration_test.go
fbm3307: no it was duplicate, removed it

pkg/verification/sender/phone_lookup_test.go
fbm3307: yes removed

pkg/verification/service/verification_service.go
fbm3307: okay done

pkg/verification/service/verification_service_test.go
fbm3307: right

pkg/verification/service/verification_service.go
fbm3307: yes, updated it to log in enabled

pkg/verification/service/verification_service_test.go
MatousJobanek: I'm wondering why it was passing, do you know?

fbm3307: my understanding is that gock.IsDone() was passing because it returns true when no pending mocks remain — since no lookup mock was registered in that test, it was always true. it didnt tell us if we skipped it or not 

When a mock is matched by an HTTP request, it gets removed from the mocks slice. If no mocks were registered in the first place, the slice is empty → IsDone() = true

MatousJobanek: but the gock mock was setup, so there should be pending mocks :thinking:

fbm3307: I looked back through the commits and no gock mock was ever registered in this particular subtest (unlike the adjacent ones like "high risk phone..." which call mockTwilioLookup). That's why gock.IsDone() was trivially true. I've removed it in the latest push since it wasn't validating anything. 
or are you refering to something else which i might have missed to understand?

MatousJobanek: sorry, not this one, but the other ones `"same phone retry skips lookup"` and `"excluded country US skips lookup"` has `mockTwilioSMS()` and at the same time it had `assert.True(s.T(), gock.IsDone())`

Actually, `"excluded country US skips lookup"` still has both - maybe worth removing it from there too.

Anyway, this is not a blocker, I' just wanted to understand why it was passing before to better understand what `gock.IsDone` actually does

PR #606, opened 2026-07-09
SANDBOX-1892: add Prometheus metrics for Twilio phone lookup

fbm3307:
Emit counters for successful lookups (allowed/blocked by risk category) and API errors so Grafana can track bot blocking and Lookup spend.

Related PR
Sandbox-Sre - https://github.com/codeready-toolchain/sandbox-sre/pull/3248

Assisted by: Cursor

## Summary by CodeRabbit

* **New Features**
  * Added Prometheus metrics for phone lookup outcomes, including allowed/blocked states and risk categories.
  * Added Prometheus metrics for phone lookup failures (fail-open) with error-type labeling.
  * Metrics are now registered during service startup for immediate monitoring.

* **Bug Fixes**
  * Phone-lookup verification now consistently records metrics for both success paths and error/fail-open scenarios.

* **Tests**
  * Added/extended unit tests to verify metric registration and correct counter increments by label values.

### DIFF
--- a/pkg/verification/service/verification_service.go
+++ b/pkg/verification/service/verification_service.go
@@ -262,15 +262,18 @@ func (s *ServiceImpl) performPhoneLookup(ctx *gin.Context, cfg configuration.Reg
 		annotationValues[toolchainv1alpha1.UserSignupPhoneLookupDetailsAnnotationKey] = string(detailsJSON)
 	}
 	if isHighRiskPhone(result) {
-		PhoneLookupTotal.WithLabelValues("blocked", result.CarrierRiskCategory).Inc()
 		log.Info(ctx, fmt.Sprintf("high risk phone detected (carrier_risk=%s, blocked=%t, phone_lookup_mode=%s)",
 			result.CarrierRiskCategory, result.NumberBlocked, mode))
 		if mode == toolchainv1alpha1.PhoneLookupModeEnabled {
+			// Count as blocked only when the signup is actually rejected.
+			PhoneLookupTotal.WithLabelValues("blocked", result.CarrierRiskCategory).Inc()
 			return true, crterrors.NewForbiddenError("phone verification rejected", "cannot proceed with verification")
 		}
-	} else {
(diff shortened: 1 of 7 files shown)


pkg/verification/service/verification_service.go
```diff
@@ -264,9 +265,15 @@ func (s *ServiceImpl) performPhoneLookup(ctx *gin.Context, cfg configuration.Reg
 		log.Info(ctx, fmt.Sprintf("high risk phone detected (carrier_risk=%s, blocked=%t, phone_lookup_mode=%s)",
 			result.CarrierRiskCategory, result.NumberBlocked, mode))
 		if mode == toolchainv1alpha1.PhoneLookupModeEnabled {
+			// Count as blocked only when the signup is actually rejected.
+			PhoneLookupTotal.WithLabelValues("blocked", result.CarrierRiskCategory).Inc()
```
alexeykazakov: I guess "blocked" and "allowed" could be consts in pkg/verification/service/phone_lookup_metrics.go but it's cosmetic.

alexeykazakov (approved):

pkg/verification/service/verification_service.go
```diff
@@ -249,6 +249,7 @@ func (s *ServiceImpl) performPhoneLookup(ctx *gin.Context, cfg configuration.Reg
 	}
 	result, lookupErr := s.PhoneLookupService.LookupPhone(e164PhoneNumber)
 	if lookupErr != nil {
+		PhoneLookupErrorsTotal.WithLabelValues("api_error").Inc()
```
metlos: This is the only use of the error metric, so we only ever get "error_type=api_error". can we for example extract the http status code from the lookup error? that way we'd at least distinguish roughly why the lookups are failing and we'd gain visibility into the availability of the lookup api (i.e. we'd be able to distinguish between 500 and 503 for example).

metlos (approved):
One minor suggestion but otherwise looks great :+1:.

